PR #72, opened 2015-01-02
Fix compile errors.

carlthronson:
Signed-off-by: Carl Thronson <cthronson@actuate.com>


carlthronson:
I'm not sure if this is the best solution.  But, this should at least fix the compile errors for now.

PR #100, opened 2015-01-09
BIRT bug fixes

carlthronson:
After all this set of commits that I did today, I wanted to make sure they compile.
They compile in my fork.  But, I don't know if they will compile after they are pulled.


carlthronson:
Never mind this one.  I think I made a mistake.  Checking again.

PR #125, opened 2015-01-29
provide license information

rkcpi:


yulinwang:
Could you please describe why you contribute this information?

rkcpi:
Because it's missing.

yulinwang:
Sorry, I can't commit this change, because:
1. We already have license information from eclipse.
2. there's no sign-off, but it's required to check in.

Thanks

PR #225, opened 2015-11-13
Bug 481003 - Report onPageEnd and onPageStart scripts broken at report render time with ClassCastException

FreemanZhang:
1. Bug was introduced in fixing Bug 431702 (Performance bottleneck in engine executor because of using arrayList.contains() method a lot.) The committer used a TreeSet to improve performance but does not implement Comparable interface, which result in ClassCastException
2. A linkedHashedSet will be enough for improving performance of contains method and simultaneously maintains the insertion order implied by arrayList.add() method


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=481003

PR #228, opened 2015-11-17
fixed error occurred during compile using maven

dominicx:
Caused by: java.lang.RuntimeException: No solution found because the problem is unsatisfiable.: [Unable to satisfy depen
dency from org.eclipse.birt.example.feature.group 4.5.0.qualifier to org.eclipse.birt.example 0.0.0.;


greatyan:
The version 4.6.0 is correct.

PR #234, opened 2015-11-18
BIRT 1388: Incorrect wording in context menu to create a new master page

FreemanZhang:
Signed-off-by: Shijie Zhang <szhang@opentext.com>


FreemanZhang:
1. Right click on masterpage in context menu and in the dropdown menu there will be an option called "Insert Element". Right click on other items in context menu will also display an option called "Insert Element". However, after clicking on "Insert Element", the dialog popped out is different, thus misleading.
2. Defined another text iteral and use a conditional assignment to differentiate between two case.

PR #291, opened 2016-04-06
- Summary: fix unit test for engine

sguan-actuate:
One fix for ReportDesignWriter to avoid showing jtidy when it is true on attribute dump
And other minor changes on test


engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/parser/ReportDesignWriter.java
```diff
@@ -93,6 +93,8 @@ public void write( OutputStream out, Report report ) throws Exception
 
 		Document document;
 		Element element;
+		
+		final static String JTIDY_ATTRIBUTE = "is-jtidy";
```
greatyan: need reuse the constant definition.

PR #324, opened 2016-06-20
Solved problem about Too Long java.io.IOException: File name too long

cwichoski:
See this forum post about the problem:
https://www.eclipse.org/forums/index.php/m/1735454/#msg_1735454

### DIFF
--- a/viewer/org.eclipse.birt.report.viewer/birt/WEB-INF/classes/org/eclipse/birt/report/session/ViewingCache.java
+++ b/viewer/org.eclipse.birt.report.viewer/birt/WEB-INF/classes/org/eclipse/birt/report/session/ViewingCache.java
@@ -120,10 +120,10 @@ public class ViewingCache implements Serializable
 		String folder = ""; //$NON-NLS-1$
 		if ( sessionId != null )
 		{
-			folder = ( prefix + sessionId ) + File.separator;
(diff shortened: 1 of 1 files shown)


yulinwang:
Sorry I don't understand your fix and problem.
Could you help file a bug in bugzilla which includes description or reports to help reproduce?

cwichoski:
Hi, 

To reproduce the problem we need to run in an app server where the Session ID has more than 255 characters, as is the case for WSO2 AS, the solution I proposed in this pull request was instead to use the full content of sessionId String, we use the hashcode of String, as Java consider that as unique too, I think we dont has any side effects, as those paths are hardcoded inside BIRT just to create temporary store for generated documents by session. In such manner we has same feature, but with a more reduced path as hashcode() method just return ints, with this solution we will only get this type of problem again when some day the session ID passes the mark of 4294967296 (2^32) characters.

What do you think?

yulinwang:
It looks reasonable. Can you sign CLA and add sign-off? Otherwise it cannot be merged to eclipse source code repository. Thank you!

Detail is:
The following users do not have valid CLA. Please see this page for details, or to sign a CLA.
clovis@wichoski.eti.br

The following users have invalid Signed-off-by footers. Please use the Signed-Off-By (git commit -s) mechanism to confirm the origin of the submission. The identity in the Signed-off-By must match the Author or the Committer.
clovis@wichoski.eti.br

PR #342, opened 2016-08-26
Set root parent version to 0.0.1

serguei-actuate:
Set root parent version to 0.0.1 since it's the new single parent
version throughout the POM hierarchy now.

Signed-off-by: Serguei Krivtsov <skrivtsov@actuate.com>


serguei-actuate:
It looks like this change is going to be rolled back. I'm closing the pull request now.

PR #347, opened 2016-10-06
fix NPE when ODADriverClass not set

flybyray:
works now if reports set ODAJndiName only

if driverClass is null because ODADriverClass is not set the call to:

org.eclipse.birt.report.data.oda.jdbc.utils.ResourceLocator.resolveConnectionProperties

will lead to an NPE in

org.eclipse.birt.report.data.oda.jdbc.utils.JDBCDriverInfoManager.getDriversInfo

Line 57: if( **driverClassName.**equals( info.getDriverClassName( )) )

### DIFF
--- a/data/org.eclipse.birt.report.data.oda.jdbc/src/org/eclipse/birt/report/data/oda/jdbc/Connection.java
+++ b/data/org.eclipse.birt.report.data.oda.jdbc/src/org/eclipse/birt/report/data/oda/jdbc/Connection.java
@@ -28,7 +28,6 @@ import org.eclipse.birt.report.data.bidi.utils.core.BidiTransform;
 import org.eclipse.birt.report.data.oda.i18n.ResourceConstants;
 import org.eclipse.birt.report.data.oda.jdbc.bidi.BidiCallStatement;
 import org.eclipse.birt.report.data.oda.jdbc.bidi.BidiStatement;
-import org.eclipse.birt.report.data.oda.jdbc.utils.ResourceLocator;
 import org.eclipse.datatools.connectivity.oda.IConnection;
 import org.eclipse.datatools.connectivity.oda.IDataSetMetaData;
 import org.eclipse.datatools.connectivity.oda.IQuery;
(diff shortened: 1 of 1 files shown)


yulinwang:
Could you provide a test case or report design to verify?
You can provide this in bugzilla: https://bugs.eclipse.org/bugs/.
Thx

flybyray:
I try to get a test report.
But maybe you check the history of the commited file. It is very easy to see that a bug was introduced by a new feature or rewrite of existing code. the new behavior was added without tests.

we Fond the Bug by upgrading frommen 4.4.2 to 4.6.0 release.
But I do not understand why the commit from 30 december 2014 
https://github.com/eclipse/birt/commit/f0343d26f1350caa280c9e647ab1646d6afdaa71#diff-17a56216dd8b89b3a47608e8cfa80206https://github.com/eclipse/birt/commit/f0343d26f1350caa280c9e647ab1646d6afdaa71#diff-17a56216dd8b89b3a47608e8cfa80206

is in the Release 4.4.2 from 17 Februar 2015

https://github.com/eclipse/birt/blob/BIRT_4_4_2_Release_201502171805/data/org.eclipse.birt.report.data.oda.jdbc/src/org/eclipse/birt/report/data/oda/jdbc/Connection.java

yulinwang:
Fixing NPE is fine, but you cannot simply remove the code which was usually added on purpose.
Can you change your fix by adding null pointer check instead?

flybyray:
"code which was usually added on purpose" - without adding testing code ??? 
some mentors would remove the whole code if there are no tests for it !!!
i will try to fix more, but the whole project is to complicated to check out and build. where is some documentation about all this.
for now i can only manually set project references in eclipse and hope that somewhere the eclipse classes get build.

yulinwang:
We do have full testing for each release. Some tests are not public because it has integrated with commercial product.
However your change doesn't have any tests and simply remove the logic. This is unacceptable.

Building BIRT could be very simple. Follow the readme. It's one step without any additional configuration.

PR #353, opened 2016-10-19
this will add simple-jndi support

flybyray:
it is very usefull for testing and deployment
see example usage video:
http://www.screencast.com/t/vABWQl9BVrrx

perhaps it shows up some advanced usage scenario for birt by adding simple-jndi

penthao is also using simple-jndi


yulinwang:
We cannot simply uploading jar to git. Usually 3p jar must be downloaded via maven from external site.
Do you know where this jar is from?

flybyray:
https://mvnrepository.com/artifact/simple-jndi/simple-jndi/0.11.4.1

flybyray:
@yulinwang At all you do not need to package the simple JNDI jar . 

> It is entirely library based, so no server instances are started, and it sits upon Java .properties files, XML files or Windows-style .ini files, so it is easy to use and simple to understand.

BIRT already looks for jndi.properties file. So the simple jndi jar should only reside in classpath.
But here you could provide help where the BIRT osgi runtime searches for "dropin" jars.

The only change of this PR would be in JDBCDriverManager bezaubernde there are anoying assumptions  https://github.com/eclipse/birt/pull/353/files#diff-a03c5190900210df82b67760b11311dcR200

PR #357, opened 2016-11-14
MongoDb 3.2 Plugin 

rgoteti:


rgoteti:
MongoDB 3.2 plugin with kerberos authentication

PR #361, opened 2016-11-15
Alias are not recognize and create new column in metadata producing error in stream

sguan-actuate:
This will fix that issue.


sguan-actuate:
added comments

PR #369, opened 2017-02-09
Overflow hidden property does not exist on table cell in DOCX

sguan-actuate:
This enhancement is to crop the words that exceed the width of the cell.
If the property is not set, it will wrap around cell and expand the
height of the row and table changing the overall layout


engine/org.eclipse.birt.report.engine.emitter.wpml/src/org/eclipse/birt/report/engine/emitter/wpml/writer/AbstractWordXmlWriter.java
```diff
@@ -1137,6 +1145,12 @@ public void writeTextInRun( int type, String txt, IStyle style,
 		}
 		else
 		{
+			if ( CSSConstants.CSS_OVERFLOW_HIDDEN_VALUE
```
yulinwang: add comments to explain

PR #386, opened 2017-04-07
Make EngineTask accessible from task's app context

garyxue:
This change adds an entry to every task's appContext to point to the task itself.


garyxue:
There are line ending issues with executioncontext. Will fix that and resubmit.

PR #419, opened 2017-09-12
Reducing presence of deprecated Any type for aggregation functions

Charling-Huang:
Add case switch statement to specify the output data type of an aggregation binding from the aggregation function when possible.


yulinwang:
API should be very simple to set the value only.
It's client's responsibility to set correct data type.

PR #427, opened 2017-12-12
Fixed emitter output to more than one page when hitting cache limit

jonfang:
Description of Issue: emitter with "multiple page" option disabled does not export content into single page due to hitting table cache limit

Description of Resolution: Add an emitter trigger function at the API layer.  Cache limit check is skipped when the emitter disables multiple sheet


engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/emitter/IContentEmitter.java
```diff
@@ -123,4 +123,6 @@
 	void startListGroup( IListGroupContent group ) throws BirtException;
 
 	void endListGroup( IListGroupContent group ) throws BirtException;
+	
+	boolean isEmitterEnableMultipleSheet( );
```
yulinwang: Can we change name to "isMultipleSheetsEnabled" and add comments for this method?

jonfang:
Revised api name; added comment

PR #435, opened 2018-01-26
fix NPE for width percentage unit conversion

jonfang:


jonfang:
Chart percentage width within a grid cell is calculated using the formula:
Chart width = Chart % width * cell width 

In case cell width is also percentage, program will continue looking for the next parent that has an actual measurement:
Chart width = Chart % width * cell % width * table width ….. * master page width

The root cause is missed handling of case where cell width is null:
Chart width = Chart%width * null * table width ….. * master page width = null

Thus throwing NPE. Code is added to properly handle the cell width.

PR #443, opened 2018-03-30
update crosstabUtil to properly handle special characters

jonfang:
crosstabUtil does not properly handle special characters for column names. This leads to corner case issue when the column name is like "[asasas]][", "?12", etc
Properly update the utility when column name contains special characters

### DIFF
--- a/xtab/org.eclipse.birt.report.item.crosstab.core/src/org/eclipse/birt/report/item/crosstab/core/util/CrosstabUtil.java
+++ b/xtab/org.eclipse.birt.report.item.crosstab.core/src/org/eclipse/birt/report/item/crosstab/core/util/CrosstabUtil.java
@@ -1116,6 +1116,10 @@ public final class CrosstabUtil implements ICrosstabConstants
 		{
 			columnName = "[^\\]]+";
 		}
+		else
+		{
+			columnName = Pattern.quote( columnName );
+		}
 		
 		ExpressionHandle expr = column
 				.getExpressionProperty( ComputedColumn.EXPRESSION_MEMBER );		


xtab/org.eclipse.birt.report.item.crosstab.core/src/org/eclipse/birt/report/item/crosstab/core/util/CrosstabUtil.java
```diff
@@ -1116,6 +1116,10 @@ public static boolean validateBinding( ComputedColumnHandle column, String colum
 		{
 			columnName = "[^\\]]+";
 		}
+		else
+		{
+			columnName = processSpecialCharacters( columnName );
```
yulinwang: Validation should not change column name, and it should just check invalid or not.

jonfang: fix code to just treat column name as string literal and remove added logic

xtab/org.eclipse.birt.report.item.crosstab.core/src/org/eclipse/birt/report/item/crosstab/core/util/CrosstabUtil.java
```diff
@@ -1116,6 +1116,10 @@ public static boolean validateBinding( ComputedColumnHandle column, String colum
 		{
 			columnName = "[^\\]]+";
 		}
+		else
+		{
```
yulinwang: This use case should be invalid and return false. In fact, such expression should be blocked in syntax check in model, and should not hit engine runtime, so better fix is to enhance model syntax check to throw meaningful error.

PR #454, opened 2018-06-29
fix regression bug on data set preview fetch row limit

jonfang:
Root cause of the issue:
Regression caused by commits 0b411e2d, 6314f9b7, and 92ceae23
- There are 2 row fetch limit values for data set preview in BDPro. One is the data set preview display limit(default is 500), the other is data set source fetch limit(default is unlimited)
- Original intent of commits is to push down preview display limit(500) to ODA prepared statement when ODA data set fetch limit is default/ unlimited. This is to address accessing Big Data type of data source.
- This causes a regression where the 500 value is hardcoded and data set is always fetching at most 500 rows from data source. 
- The regression causes 2 issues:
1.Data set filters are always applied only on the fetched 500 rows of data rather than the entire data set
2.Customly setting the data source fetch size for size>500 no longer works as preview always fetch 500 rows


jonfang:
Updated the fix where the default max row is still 500. But manually setting the max fetch limit in designer now works properly, whereas before setting any limit > 500 will always restrict the max to fetch to 500.

PR #459, opened 2018-08-03
Modify date parsing code to be able to return the date format pattern…

Charling-Huang:
… string

### DIFF
--- a/core/org.eclipse.birt.core/src/org/eclipse/birt/core/data/DataTypeUtil.java
+++ b/core/org.eclipse.birt.core/src/org/eclipse/birt/core/data/DataTypeUtil.java
@@ -823,6 +823,26 @@ public final class DataTypeUtil
 	public static Date toDate( String source, ULocale locale, TimeZone timeZone )
 			throws BirtException
 	{
+		DateFormat dateFormat = getDateFormat( source, locale, timeZone );
+		Date resultDate = null;
+		try
+		{
+			resultDate = dateFormat.parse( source );
+			return resultDate;
+		}
+		catch ( ParseException e )
+		{
+			throw new CoreException(
+					ResourceConstants.CONVERT_FAILS,
+					new Object[]{
+							source.toString( ), "Date"
+					} );
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


core/org.eclipse.birt.core/src/org/eclipse/birt/core/data/DataTypeUtil.java
```diff
@@ -822,6 +822,26 @@ public static Date toDate( String source, ULocale locale )
...
+		{
+			throw new CoreException(
+					ResourceConstants.CONVERT_FAILS,
+					new Object[]{
+							source.toString( ), "Date"
+					} );
+		}
+	}
+
+	public static DateFormat getDateFormat( String source, ULocale locale, TimeZone timeZone )
```
yulinwang: Please add comments and tag @since 4.8

core/org.eclipse.birt.core/src/org/eclipse/birt/core/data/DateFormatISO8601.java
```diff
@@ -42,17 +43,39 @@
...
+			return resultDate;
+		}
+		catch ( ParseException e )
+		{
+			throw new CoreException( ResourceConstants.CONVERT_FAILS,
+					new Object[]{source.toString( ), "Date"} );
+		}
+	}
+
+	public static SimpleDateFormat getDateFormat( String source, TimeZone timeZone ) throws BirtException,
```
yulinwang: Add comments and since.
If SimpleDataFormat is returned, please consider changing name to getSimpleDateFormat.

core/org.eclipse.birt.core/src/org/eclipse/birt/core/data/DateFormatISO8601.java
```diff
@@ -36,23 +37,57 @@
 	
 	/**
 	 * Parse a date/time string.
+	 * @since 4.8
```
yulinwang: Only add since tag for new method. No need to add for existing method.

PR #466, opened 2018-09-03
Fix for "Needs Cache for Data-Engine" not being respected for Scripted DataSets

rvinjamu:
If multiple report items on a report have the same Dataset as the source and if "Needs cache for data-engine" property is set to true, then DataSet generation should happen only once. Currently this optimization is failing for Scripted Data Sets. This change fixes the issue.


rvinjamu:
This logic is controlled by a specific flag on the dataset - 'Needs Cache for Data Engine' - the default value for which is true. In case the customer has two report items with the same scripted data set as source and he needs the data to be dynamic, he can set this flag to false for the scripted dataset in question.

rvinjamu:
based on last comment - reopening it for review

PR #469, opened 2018-09-07
Resolve images issue inside generated Excel report 

gorchenok:
This issue related to Excel report that contains more than 256 columns.
All Images after 256 column are shrink to one column width


yulinwang:
Could you file a bug in bugzilla https://bugs.eclipse.org/bugs/index.cgi?
Please attach a report design so that it's easier to reproduce and verify your fix.
Thanks for your contribution.

PR #470, opened 2018-09-14
add value check for N for N and Percent aggregation accumulator

jonfang:
For TopN/BottomN aggregation BIRT has a NAccumulator that contains a BasicCachedArray which stores all   running N elements. Aggregation is always assuming there is at least 1 element. So when all elements are filtered and aggregation is evaluated on empty input, the default N = -1 falls through to initialized size of BasicCachedArray. 
Add fix to properly evaluate the N value before passing it to initialize BasicCachedArray. The fix is similar for Top percent N aggregation.


yulinwang:
Top N should support negative values as well. It doesn't make sense to convert them to ZERO. Please test negative value case.

jonfang:
Added unittests to cover both top N and top N percent aggregation against empty data set. Now the resulted computed column will return null instead of throwing errors

PR #473, opened 2018-09-28
Fix for Security Issue : https://bugs.eclipse.org/bugs/show_bug.cgi?id=538142

rvinjamu:
…d=538142

Through Birt Viewer, one can perform operations such as generating a report as a frameset, output the report document in some format, generate a report document and so on. One parameter these operations take is "__document" that represents the report document file to be generated. Using executable extension such as .jsp for the __document parameter and using a specially crafted report, one can access sensitive information about the system or do something nasty on the server by accessing the generated report document. The fix introduces a configurable whitelist and blacklist for the report document extension.


grundsch:
Hi,
Is there a CVE for this vulnerability? If not, do you plan to request one?
Thanks for feedback,
Stéphane

PR #475, opened 2018-10-24
Fix to handle the chart view for the x tab

vnijhawa:


chart/org.eclipse.birt.chart.reportitem/src/org/eclipse/birt/chart/reportitem/ChartReportItemUtil.java
```diff
@@ -251,8 +253,12 @@ public static boolean isEmpty( IBaseResultSet set ) throws BirtException
 		if ( set instanceof IQueryResultSet )
 		{
 			return ( (IQueryResultSet) set ).isEmpty( );
+		}  else if ( set instanceof ICubeResultSet )
+		{
+			  CubeCursor cursor = (( (ICubeResultSet) set ).getCubeCursor());
+			  return cursor.getOrdinateEdge().size() != 0;
```
yulinwang: Please call isEmpty() instead and check null for cursor

PR #490, opened 2018-12-27
Housekeeping

Flugtiger:
I just downloaded BIRT and setup a development environment for it. These are some changes I did during this process. Maybe some of them can be merged back.


yulinwang:
This change list is not valid since the CI build failed

Flugtiger:
Build is fixed now.

The main change is the removal of the .project and .classpath files from the repo since they get generated by m2e when importing the maven projects into eclipse.

yulinwang:
I'm sorry we cannot accept removing .project and .classpath files, because these files are critical metadata for eclipse project even without maven.

Flugtiger:
Well, I'd argue that those files are generated by the m2e eclipse plugin anyway (from the meta data in the pom.xml, plugin.xml, MANIFEST.MF, etc.) so there is no need to include them in the repository. There are even some .project files missing in the repo that are generated by eclipse.

Nevertheless, I deleted the commit that has removed the .project files from this pull request. Instead I added a commit to add all those .project files that were missing in the repo.

Flugtiger:
Now, I see the root of the problem. The .project files that are included in the repo are missing the maven nature. I consider this a bad thing since it allows the maven build and the internal build of eclipse to diverge. We should fix this.

I added another commit that adds all of the missing maven natures to the .project files.

In this process, files under the .settings folder have also been modified by m2e. Can we agree that including these files in the repository is discouraged? Can I remove them or shall I commit the modifications done by m2e?

UI/org.eclipse.birt.report.designer.ui.editors/build.properties
```diff
@@ -1,5 +1,5 @@
 source.. = src/
-output.. = bin/
```
yulinwang: I'm not sure if this change may cause eclipse PDE's problem. Did you test PDE environment in eclipse IDE?
BIRT source code not only support maven, but also support traditional plugin PDE. Using different output folder can make sure PDE and maven work simultaneously.

UI/org.eclipse.birt.report.designer.ui.editors/pom.xml
```diff
@@ -7,7 +7,6 @@
 		<version>4.8.0-SNAPSHOT</version>
 		<relativePath>../../</relativePath>
 	</parent>
-	<groupId>org.eclipse.birt</groupId>
```
yulinwang: This change seems unnecessary, because it can inherit from parent.

UI/.project
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
yulinwang: Root folder and module folder like UI, Chart, Engine... are not project because no jar will be generated from this folder. We
What is the reason to add .project file here?

UI/org.eclipse.birt.report.designer.samplereports/.project
```diff
@@ -5,6 +5,11 @@
 	<projects>
 	</projects>
 	<buildSpec>
+		<buildCommand>
```
yulinwang: This project is not java project.

UI/org.eclipse.birt.report.designer.ui.editors/build.properties
Flugtiger: This change is motivated by Eclipse PDE and fixes a warning inside Eclipse.

To get a buildable PDE workspace, I imported the BIRT projects using maven import. In this process the m2e connector updates the Eclipse projects using the meta data in the pom.xml. This makes "target/classes" the output folder of the projects. This in turn leads to a warning from PDE along the lines of "'bin/' is not an output folder of the project".

UI/org.eclipse.birt.report.designer.ui.editors/pom.xml
Flugtiger: It's the other way around. -> Without this change Maven outputs a warning *because* the groupId can be inherited from the parent.

UI/.project
Flugtiger: The .project files are generated by Eclipse when importing the maven projects.

Flugtiger:
You are right, this pull request is getting too big. I will split it up into several new pull requests.

yulinwang:
Since it has been split to multiple small pull requests, can you close this one?

Flugtiger:
I'm still not quite sure how to use the maven nature in the eclipse projects of BIRT. I will close this pr and submit a new one if I have a good solution. Is it possible to discuss this on the developer mailing list?

yulinwang:
Sure

PR #492, opened 2019-01-06
Add PDE target definition for Photon

Flugtiger:
I created a PDE target definition for the current target platform of the master branch. It resolves all dependencies currently needed by BIRT.

It is almost in sync with the tycho target platform. Only difference is the addition of a recent Apache Jasper version from Gemini as it is required by the Jetty Version currently used by BIRT.

Signed-off-by: Alexander Lehmann <a.lehm@gmx.de>

### DIFF
--- a/BIRT_photon.target
+++ b/BIRT_photon.target
@@ -1,0 +1,67 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?><?pde version="3.8"?><target name="BIRT Photon target platform">
+	<locations>
+		<location includeAllPlatforms="false" includeConfigurePhase="true" includeMode="planner" includeSource="true" type="InstallableUnit">
+			<repository location="http://download.eclipse.org/tools/orbit/downloads/drops/R20180330011457/repository/"/>
+			<unit id="org.eclipse.orbit.mongodb" version="3.2.2.v20170222-2110"/>
+			<unit id="com.lowagie.text" version="2.1.7.v201004222200"/>
+			<unit id="org.apache.batik.bridge" version="1.7.0.v201011041433"/>
+			<unit id="org.apache.batik.css" version="1.7.0.v201011041433"/>
+			<unit id="org.apache.batik.dom" version="1.7.1.v201505191845"/>
+			<unit id="org.apache.batik.dom.svg" version="1.7.0.v201011041433"/>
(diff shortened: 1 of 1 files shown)


yulinwang:
Could you tell me the benefit to have this file?
BIRT uses OSGi bundle to resolve dependencies, and maven tycho is using that too. I don't understand why you need to extra file to specify dependencies.

Flugtiger:
This file is needed to be able to build, run & debug BIRT using Eclipse PDE. It tells PDE which plugins have to be present in the target platform besides those inside the Eclipse workspace.

If not via Eclipse PDE, how do you make changes to BIRT? Do you always run a maven build whenever you want to test something?

yulinwang:
Yes, we use maven to run CI/release build, and use PDE to compile/run/debug locally.
The best practice to develop BIRT is you can use BIRT as IDE as well, so that you only need to pick some projects to debug source code while other plugins can load from IDE directly.

yulinwang:
Using BIRT as IDE is easier for you to develop but not mandatory. It can help you reduce work.
You can use any IDE and take advantage of plugin dependency or feature definition to find which plugin are required to compile and run.  
I just don't see any reason to introduce duplicate configuration with more work to maintain. All of these definitions can be found in plugin manifest.

Flugtiger:
It's not more work to maintain, if we let tycho use the target definition file as well. It would give us fine-grained control over the version and sources of plugins used by BIRT. And it would allow us to easily swap the target platform with a newer or older one.

Flugtiger:
Two answers ;-) :

1. Yes, much like the separate maven profiles we have defined for each platform in the parent-pom
2. No, these version do not need to be updated manually. The target platform editor of PDE does most of the work for you. You only need to update the URLs of the p2-repos and choose which features shall be included from each repo. I admit: choosing the specific features of each repo is some extra work compared to the current solution (which only consists of the p2-repo URLs in the parent-pom). But this extra bit of work is rewarded with the certainty, that all required features are still there. Additionally, if we let tycho also use the target platform definition of PDE, it should be able to resolve the target platform much faster, as stated by [the tycho docs:](https://wiki.eclipse.org/Tycho/Target_Platform#Which_approach_shall_I_use_for_the_target_platform_of_my_project.3F)

"Using a target file instead of p2 repositories in the POM may also speed up the build because fewer bundles in the target platform will often lead to a faster dependency resolution."

PR #496, opened 2019-01-08
Add Automatic-Module-Name Header for Java 9

Flugtiger:
This fixes a warning


yulinwang:
Conflict must be resolved at first.

PR #498, opened 2019-01-08
Fixed plugin.xml warnings (missing unbounded in extension point schema)

Flugtiger:
Some of the extensions specified by BIRT itself did not match the extension point schema -> updated the schemas -> resolved warnings

### DIFF
--- a/chart/org.eclipse.birt.chart.engine/schema/aggregatefunctions.exsd
+++ b/chart/org.eclipse.birt.chart.engine/schema/aggregatefunctions.exsd
@@ -1,6 +1,6 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <!-- Schema file written by PDE -->
-<schema targetNamespace="org.eclipse.birt.chart.engine">
+<schema targetNamespace="org.eclipse.birt.chart.engine" xmlns="http://www.w3.org/2001/XMLSchema">
 <annotation>
       <appInfo>
          <meta.schema plugin="org.eclipse.birt.chart.engine" id="aggregatefunctions" name="BIRT Chart Library Aggregate Function Extension"/>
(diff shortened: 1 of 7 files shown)


data/org.eclipse.birt.report.data.bidi.utils/build.properties
```diff
@@ -2,6 +2,5 @@ source.. = src/
 output.. = bin/
 bin.includes = META-INF/,\
                .,\
-               plugin.xml,\
```
yulinwang: I don't see why you remove plugin.xml. It's still required.

chart/org.eclipse.birt.chart.engine/schema/aggregatefunctions.exsd
```diff
@@ -11,8 +11,13 @@
...
 
    <element name="extension">
+      <annotation>
+         <appInfo>
+            <meta.element />
+         </appInfo>
+      </annotation>
       <complexType>
-         <sequence>
+         <sequence minOccurs="1" maxOccurs="unbounded">
```
yulinwang: This change is incorrect. It should be [1,1], and not shown by default.

Flugtiger: Project "org.eclipes.birt.chart.engine.extension" defines multiple aggregatefunctions in this exact way (i.e. as a sequence of multiple aggregateFunction elements). This contradicts [1,1] and is resolved by setting [1,∞). The concerning file is:
chart\org.eclipse.birt.chart.engine.extension\plugin.xml

yulinwang: That looks weird. Even it's defined as [1,1], it's still working properly. But it should be as [1,∞)

PR #500, opened 2019-01-12
Fixed outdated/invalid classpath entries in MANIFESTs

Flugtiger:


engine/org.eclipse.birt.report.engine.emitter.config/META-INF/MANIFEST.MF
```diff
@@ -3,8 +3,7 @@ Bundle-ManifestVersion: 2
 Bundle-Name: BIRT Emitter Configuration Plug-in
 Bundle-SymbolicName: org.eclipse.birt.report.engine.emitter.config;singleton:=true
 Bundle-Version: 4.8.0.qualifier
-Bundle-ClassPath: library.jar,
```
yulinwang: I do find library.jar in this plugin, so cannot simply remove this from classpath.

yulinwang:
I find these internal jars are used by fragments instead of host plugin itself.
I need further verification to ensure it won't impact class loading.

Flugtiger:
As far as I could tell, none of those jars I removed are even existent anymore.

yulinwang:
We find a similar regression which is related to change https://github.com/eclipse/birt/commit/66d91e3f64fddccb7361635939a293de4e19f4db.
The library.jar in classpath cannot be removed now, because it's used to be overridden by fragment.
I am still thinking how to remove this unnecessary reference in host plugin while keep fragment working.

Flugtiger:
ok, i'll close this pull request until there is a solution

PR #505, opened 2019-01-17
retrieve latest version of datatools that supports Oracle 12 and SQL server 2016

sguan-actuate:


sguan-actuate:
the checksum of repository is wrong

PR #510, opened 2019-02-02
Fix core tests

Flugtiger:
This pull request fixes the unit tests of the birt.core plugin. Many tests have failed on my machine because they were locale or time zone dependent.


Flugtiger:
Superseded by #516

PR #520, opened 2019-03-02
Make tycho use a pde target platform definition file

Flugtiger:
This PR adds a new project 'build/org.eclipse.birt.target' containing the current target platform definition file needed to build BIRT. Then tycho is reconfigured to use this file for the dependency resolution in the build process.

The target definition file can be set as target platform in Eclipse via Settings -> Plugin Development -> Target Platform. It should resolve all dependencies of BIRT.

By using the target platform definition file, the tycho build is shortened by about 3 minutes.

I'm not sure how we could supply different target platform definitions for the different Eclipse Platform versions. The status quo uses maven profiles but there are also different git branches for the different Eclipse versions. How does this go together? If we use different branches for different Eclipse versions, we could have a single target definition file that has different content on each branch. In this case there would be no need for multiple maven profiles. We could also create multiple target platform projects for the different Eclipse versions and switch between them by maven profile.
What is the prefered solution here?


yulinwang:
Currently, BIRT allows integrating with different eclipse versions like master(photon) or oxygen.
Could you add support for oxygen target as well?

PR #521, opened 2019-03-02
Fix warning: 'no required exec env set'

Flugtiger:


yulinwang:
Could you update these JDK to 1.8?
The jdk version has been set in manfest.mf. They are supposed to be 1.8, and these classpathentry should be consistent with them.

Flugtiger:
all of them?

Btw: Should 1.8 maybe be set as minimum for the whole BIRT project?

yulinwang:
I just find I didn't publish these comments.
Sorry for late reply.

PR #523, opened 2019-03-02
Fix missing/unused serialVersionUID

Flugtiger:


data/org.eclipse.birt.core.script.function/src/org/eclipse/birt/core/script/function/bre/BirtComp.java
```diff
@@ -36,11 +36,6 @@
 {
 
 
-	/**
-	 * 
-	 */
-	private static final long serialVersionUID = 29055295217417372L;
```
yulinwang: why do you remove this serial version UID? I think it was intendedly added.

Flugtiger: Because the class BirtComp does not implement Serializable (although maybe it did sometime in the past).

yulinwang: These functions like BirtComp need to be serializable in some cases.
I would prefer to add the Serializable interface rather than remove the UID.

Flugtiger: I just added a commit that makes IScriptFunctionExecutors Serializable and adds the serialVersionUID accordingly

PR #531, opened 2019-04-02
commit before closing the connection

bkiselka:
This leaves no open transactions using the postgres driver the default way. As the postgres driver uses the policy DBConfig.SET_COMMIT_TO_FALSE configured in config.xml, which sets autocommit to false by default.

The open transaction was introduced by this commit: https://github.com/eclipse/birt/commit/3936b041de5c98c041a1b0a664387d10291ccdb5

To show the "idle in transaction" open database connection (which is reused, though!) run 
SELECT * FROM pg_stat_activity
on the database after generating a report using BIRT.


bkiselka:
tried a new with #536

PR #532, opened 2019-04-04
In preview of dataset Filter is not applied on complete results but filtered on maxrows only.

rrimmana:
If we have filter and when we do preview results of DataSet filter,
filtering should be done on complete results
not just on the limited results obtained by limited max rows/ rowfetchlimit size.

Signed-off-by: Ravi Kiran Rimmanapudi <rrimmana@opentext.com>


yulinwang:
It looks like your change may have potential problems.
1.	New change will not add max row limit if there’s no filter. Is that correct?
2.	If I understand the problem correctly, the root cause of this problem is when we should apply the max row limit. Is it on final result set after query processing, or on the raw data before query processing?

yulinwang:
Since push down filter doesn't have filter events, it still applies max row here. Non push down filter will be affected but there's another max row being applied after this operation. It should be safe to change like this.

PR #536, opened 2019-04-25
commit before closing the connection

bkiselka:
This leaves no open transactions using the postgres driver the default way. As the postgres driver uses the policy DBConfig.SET_COMMIT_TO_FALSE configured in config.xml, which sets autocommit to false by default.

The open transaction was introduced by this commit: eclipse@3936b04

To show the "idle in transaction" open database connection (which is reused, though!) run
SELECT * FROM pg_stat_activity
on the database after generating a report using BIRT with PostgreSQL JNDI.

Signed-off-by: Bernhard Kiselka <bk@wigeogis.com>

### DIFF
--- a/data/org.eclipse.birt.report.data.oda.jdbc/src/org/eclipse/birt/report/data/oda/jdbc/Connection.java
+++ b/data/org.eclipse.birt.report.data.oda.jdbc/src/org/eclipse/birt/report/data/oda/jdbc/Connection.java
@@ -476,6 +476,15 @@ public class Connection implements IConnection
 
 			if ( jdbcConn.isClosed( ) == false )
 			{
+				// if the policy DBConfig.SET_COMMIT_TO_FALSE is used, which sets autocommit to false by default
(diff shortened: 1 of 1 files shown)


yulinwang:
Hi bkiselka,
Could you file a bug in eclipse bugzilla and attach your reproducible report design or steps, so that our QA can verify the fix?

Thanks,
Yulin

bkiselka:
Bug containing also the report design file: 
https://bugs.eclipse.org/bugs/show_bug.cgi?id=546041
In fact, it's very simple. Please state, if you also need a database dump / backup script to verify!

yulinwang:
Thank you, Bernhard.
If you can provide anything like screenshot or script, to help our QA to verify, it would be the best.
Basically, we want to understand why this change is necessary and what's the problem without the fix.

yulinwang:
From my understanding, if you close connection, any still running transactions will be rolled back and terminated automatically. Why do you need to commit again?

yulinwang:
It looks like not all JDBC drivers will commit/rollback transaction when the connection is closed.  I agree it's a valid fix.

PR #549, opened 2019-07-24
Fix for XSS vulnerability Bug 546816

sguan-actuate:
Reflected XSS vulnerability in the __format URL parameter
It should also take care of the other parameters.
It make sure all parameter are escape to avoid attack.
tested with associate unit test.

Signed-off-by: shiheng guan <guans@opentext.com>

### DIFF
--- a/viewer/org.eclipse.birt.report.viewer/birt/WEB-INF/classes/org/eclipse/birt/report/utility/ParameterAccessor.java
+++ b/viewer/org.eclipse.birt.report.viewer/birt/WEB-INF/classes/org/eclipse/birt/report/utility/ParameterAccessor.java
@@ -2104,7 +2104,7 @@ public class ParameterAccessor
 			{
 			}
 		}
-		return request.getParameter( parameterName );
+		return htmlEncode( request.getParameter( parameterName ) );
 	}
 
 	/**
@@ -2258,7 +2258,7 @@ public class ParameterAccessor
 		{
 			return null;
 		}
-
+		filePath = htmlDecode( filePath );
 		if ( isEncodedPaths( request ) )
 		{
 			return decodeBase64( filePath );


Caroline-Jean:
Hi, please does it solve the problem of injection SQL attack too ? or do you know if someone else worked on that and solved it ? I didn't find anything about that. Thanks

hvbtup:
It is not BIRT's task to prevent SQL injection.

If you write your SQL statements correctly, SQL injection is not an issue.

Correctly means:

1) **Don't use string concatenation to create SQL statements!**
2) **If you have to break rule 1: Don't use user input in this string concetenation.**
3) **If you really have to break rule 2: Use correct escaping (and test it!!!!!)**

For 99% of all use cases, just obey to rule 1.
Use bind variables in your statements. That's those `?` in JDBC syntax.
By using bind variables, you avoid any potential security issues with SQL injection for any decent database.

To explain rule 2:

In very rare cases, you find yourself in a situation where you need to dynamically modify your SQL statement in the `beforeOpen` event.

This does usually *not* involve user input, though!
Always use bind variables for referring to user input.

Typical cases might be:
- You want to dynamically adjust the ORDER BY clause.
- You want to dynamically adjust he table to be selected FROM.

For changing the ORDER BY clause, the best option is to allow only a fixed set of column lists and to a construction like this (untested):

    var orderBy = { "A": "1, 2, 3", "B": "1 DESC, 2 DESC, 3 DESC" } [this.getInputParameterValue("param_XXX")];
    if (orderBy == undefined) {
        orderBy = "1, 2, 3";
    }
    this.queryText = this.queryText + " ORDER BY " + orderBy;

The next best (more flexible) option would be to allow the column sequence to be specified directly with a report parameter.
In this case, it is absolutely important to check that the parameter matches a strict regular expression, e.g. it MUST be comma separated sequence of digits and optionally followed by the word "DESC". But you see, writing such a regex is not trivial for most developers. A too relaxed check (or not reacting accordingly if the check fails) might allow SQL injection attacks.
 
For dynamically swapping the table in the FROM clause, a similar approach with the same caveats can be used.

**The general idea is to make sure that the SQL statement is only concatenated from known, constant values and the users can only choose which combination of these values should be used, but they cannot directly add text to the SQL statement.**

To explain rule 3:

Only in very, very, very rare cases (always talk with someone else before, maybe there is a different way) - I cannot think of any real-life use-case, actually - you may need to use untrusted user input as string constants in your SQL (e.g. because the query optimizer isn't doing his job with bind variables.
In this case, it is absolutely important to double-check the user input.
- Make sure that you understand what you're doing. This might sound logical, but there's so many programmers just copy&pasting code without understanding it...
- Use a white list of allowed characters (e.g. only letters and numbers).
- Just to be sure, always escape the (already white-listed) input. Make sure you use a correct, well-known, function for this escaping.
- Explain with comments why you need to modify the SQL text.
- Use 4-eyes principle before commiting your code. 

A fourth rule:

4) Make sure that all BIRT reports deployed to your server actually obey rules 1-3.

In other words: 
Do not allow untrusted sources to place programs (here: *.rptdesign, *.rptlibrary, *.js, *.jar files) onto your server.

PR #551, opened 2019-08-03
Replaced static dependencies to Apache POI with plugin dependency

Flugtiger:
Linking the Apache POI library statically has multiple disadvantages:
* The build log is polluted with 'Detected duplicate entry'-Messages yielded by the FrameworkPacker
* The maven artifact of BIRT does also link statically to Apache POI which prevents mavens conflict resolution mechanism to work

To resolve these issues, all static references to Apache POI are replaced with plugin dependencies to the POI plugin coming from orbit.

Signed-off-by: Alexander Lehmann <a.lehm@gmx.de>


yulinwang:
Your change looks good.
However, the travis CI fails because of missing JDK 8 toolchain.
I need some time to resolve it, or if you can resolve the travis CI issue, that can expedite the whole progress.

yulinwang:
Now travis CI can pass, but your changelist still fails because for some reason it always tries to compile with jdk 11. Could you make it pass so that I can merge it?

Flugtiger:
I rebased this PR onto master, it now uses the correct JDK...

PR #553, opened 2019-08-06
Fix Travis build

Flugtiger:
Travis has updated the default linux distribution from trusty to xenial. On xenial, OracleJDK 8 is not supported anymore. My solution: use OpenJDK 8 on xenial. Alternative solutions: use OracleJDK 8 on trusty or update to Oracle/OpenJDK 11 on xenial.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: java
 jdk:
-  - openjdk11
+  - openjdk8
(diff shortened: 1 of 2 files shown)


Flugtiger:
Travis doesn't queue the build, closing and reopening this pr to bump it a little.

PR #567, opened 2020-02-14
Added support for Prepending and Appending PDF documents to PDF output

TheLaserNinja:
Adding support to append and prepend PDF documents to PDF output stream.
This was previously committed, but held up due to not being a comitter
previously.

Have the ability to append boiler plate legal documentation or financial
statements to a report, the ability to append PDF files based on a
comma-seperated list or a ArrayList defined in the Report Context has
been added.

This has been tested in actual customer environments and works. To set:
-Create a BIRT Report
-Create a User Property called AppendList.
-Set AppendList under Advanced to a Comma-Seperated list of paths to PDF
files to append

-Alternatively, create an ArrayList of String in the report context set
in global variable called appendPDF.

Signed-off-by: TheLaserNinja <36558386+TheLaserNinja@users.noreply.github.com>

### DIFF
--- a/engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
+++ b/engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
@@ -13,13 +13,21 @@ package org.eclipse.birt.report.engine.emitter.pdf;
 
 import java.awt.Color;
 import java.io.BufferedOutputStream;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.InputStream;
 import java.io.OutputStream;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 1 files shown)


bkiselka:
Based on my local tests the "path to the PDF file" on windows needed to be absolute, which is not very practical for multiple different environments.
Thus my idea for an improvement to add an else-part for every of the 4 f.exists() checks: 
```java
if (f.exists()) {
	FileInputStream fis = new FileInputStream( f );
	pdfs.add(fis);
} else {
	// get the file using context.getResource() for relative or universal paths
	URL url = context.getResource(fileName);
	InputStream is = new BufferedInputStream(url.openStream());
	pdfs.add(is);
}
```
That way it is possible to define e.g. `append.pdf` (without full path)  as value of `appendList` and/or `prepend/introduction.pdf` as value of `prependList` relative to BIRT's workingFolder (where also the *.rptdesign reside).

Additionally I personally also prefer the PDFs to be concatenated by `concatPDFs(pdfs, false);` instead of `concatPDFs(pdfs, true);` causing no pagination added to the bottom of the appended page, which will be printed as "1 of 1" on an appended PDF with only one single page. I think a pagination does not make sense anyway, because the number of pages of the PDFs concatenated are not considered at all.

PR #577, opened 2020-04-15
Fix Bug 543133   missing plugin in birt 4.8 runtime

asc666:
build.xml for Runtime modified: exlusion of  *enablement* jars removed


asc666:
I am Sorry this pullreqest was erroneously added

Issue #585, opened 2021-03-04
Switch from Travis CI to GitHub Actions


ruspl-afed:
GitHub Actions seems to be the way forward to use for GitHub-hosted projects. Let's switch to GitHub Actions:
 * add GitHub Actions configs
 * update README
 * remove Travis CI configs

ruspl-afed:
We are done here

PR #587, opened 2021-03-04
Switch from Travis CI to GitHub Actions #585

ruspl-afed:
Update README
Remove Travis manifests

Signed-off-by: Alexander Fedorov <alexander.fedorov@arsysop.ru>


Flugtiger:
LGTM, but I'm still waiting for my committer permissions ;-)

ruspl-afed:
@Flugtiger it seems that cache magic works! The first pipeline was 30+mins downloding everything, now it is just 10+mins and starts from "resolving". Thank you for your suggestion for the previous PR.
I think we have even more room to optimize the build time after sorting out all the p2 repos and products that we are producing during the build.

wimjongman:
Thanks, Alexanders, for fixing this. Your committer requests are pending. In the meantime, ping me, and I will merge.

PR #590, opened 2021-03-05
Fix core tests

Flugtiger:
These commits try to fix the tests of the core plugin. One of the problems was the tests being dependent on the systems timezone / locale. My solution was to make the tests independent. Another solution would be to alter the system tz / locale in the maven build to the one the tests assume to be active (unclear what exactly it needs to be).

### DIFF
--- a/core/org.eclipse.birt.core.tests/test/org/eclipse/birt/core/archive/AllArchiveTests.java
+++ b/core/org.eclipse.birt.core.tests/test/org/eclipse/birt/core/archive/AllArchiveTests.java
@@ -57,7 +57,7 @@ public class AllArchiveTests
 		test.addTestSuite( FolderToArchiveTest.class );
 		test.addTestSuite( InputStreamRefreshTest.class );
 		test.addTestSuite( SpecialCharacterTest.class );
-		test.addTestSuite( TestBuffer.class );
+		// test.addTestSuite( TestBuffer.class );
 		// add all test classes here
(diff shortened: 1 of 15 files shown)


core/org.eclipse.birt.core.tests/test/org/eclipse/birt/core/archive/AllArchiveTests.java
```diff
@@ -57,7 +57,7 @@ public static Test suite( )
 		test.addTestSuite( FolderToArchiveTest.class );
 		test.addTestSuite( InputStreamRefreshTest.class );
 		test.addTestSuite( SpecialCharacterTest.class );
-		test.addTestSuite( TestBuffer.class );
+		// test.addTestSuite( TestBuffer.class );
```
ruspl-afed, line 60: We shouldn't have a code that just commented out. Let's either remove it or better convert it to the TestCase

ruspl-afed (changes requested):
Let's try to enable tests for this bundle to check the result of the fix

Flugtiger:
Although tests have passed in this CI run, there's an occasional dead lock in org.eclipse.birt.core.archive.FolderArchiveTest which causes the build to hang indefinitely sometimes.

ruspl-afed:
@Flugtiger I suggest to merge this one and fight with `FolderArchiveTest` separately

PR #591, opened 2021-03-05
Fix model tests

Flugtiger:
These commits try to fix the tests of the model plugin and a bug with TableHandle#setWidthToFitColumns. The biggest problem is the tests being dependent on the system locale. My solution was to make the tests independent. Another possible solution would be to set the locale used in the maven build to the one that the tests need (unclear what that would be).


ruspl-afed:
It would be great to see that tests are passing the pipeline as a result of the fix.
I suggest to enable the tests for this PR and ensure that the fix is working

Flugtiger:
I will try to enable tests only for the bundle I fixed. This is what I did in my travis build. Just have to migrate it to github actions.

ruspl-afed:
> I will try to enable tests only for the bundle I fixed. This is what I did in my travis build. 

+1

> Just have to migrate it to github actions.

That should be easy, it is in the `master` already.

Flugtiger:
The tests for the module are now enabled and two tests were failing. I will look into it tomorrow.

wimjongman:
Can this one go as well?

Flugtiger:
Nah, still need to fix those tests.

wimjongman:
> go searching my faith in Java compilers.

😂 Let me know when you found it.

Flugtiger:
Puhh, I could now reproduce the test failure locally. I had to set the target platform to locale 'en' (my default locale is 'de'). The test then failed but the Eclipse debugger kept telling me lies: it said 'messages' was null but the != null check returned true 😒. The solution was now to add the missing property to the property file and additionally remove the built jar-file from git so the new property could actually be effective (why is there a jar file checked into git??? I had to debug down into java.util.ResourceBundle to find out why the new property was not found either 😠).

ruspl-afed:
> and additionally remove the built jar-file from git so the new property could actually be effective (why is there a jar file checked into git??? I had to debug down into java.util.ResourceBundle to find out why the new property was not found as well 😠).

Awesome

wimjongman:
For some reason, I pushed the java 1.8 conversion. I never accepted the pull request. I guess I must have pushed to master directly. It compiles fine but there is a blocking test.

Flugtiger:
@wimjongman this can now be merged from my side

wimjongman (approved):
Wonderful!

PR #595, opened 2021-03-05
Make .settings folders ignored by git

### LINKED ISSUE #594: All .settings folders are listed by git as "untracked"
...they should be added to .gitignore

Flugtiger:
Fixes #594

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 bin/
 /.reviewboardrc
+.settings


.gitignore
```diff
@@ -1,2 +1,3 @@
 bin/
 /.reviewboardrc
+.settings
```
ruspl-afed, line 3: I'm not so sure regarding this because .settings are widely used to provide valuable information, it allows to:
* enforce compiler settings (used by tycho during build)
* enabled DS for OSGi components
* enforce UTF-8 for resources
* enforce formatting
* and more

Flugtiger: In my experience, the poms as single source of truth are well sufficient, since m2e syncs all important eclipse settings with the poms. For the rest (e.g. formatter profile) an Oomph setup can be used. The downside of checking .settings into git is, that keeping them in sync with the poms can be quite a pita, since m2e tends to reorder xml elements occasionally.

ruspl-afed: That was my position as well some time ago, but then I switched to pomless and .setting-centric approach.

To be consistent one should set `useProjectSettings` to `false` to fully rely on maven compilation settings
https://www.eclipse.org/tycho/sitedocs/tycho-compiler-plugin/compile-mojo.html#useProjectSettings

> Oomph setup can be used

True! And I love Oomph very much. But a lot of people avoid it and just clone the repo.

Flugtiger: Pomless + .settings sounds like a valid alternative. As long as there is a single source of truth I am happy 😉. If you have experience with pomless and you think we should switch towards it, let's do it.
In the meantime: do you think we should put .settings in to git or should we set useProjectSettings to false and make git ignore them?

wimjongman: +1 for putting .settings in git (do not ignore)

I agree with Alexander F. .settings contain crucial Eclipse information. For example the formatting rules.

Flugtiger:
closed in favor of #597

PR #601, opened 2021-03-07
Code cleanup #600

wimjongman:
Code cleanup proposal


Flugtiger:
I'm afraid, but changing the required Java runtime to Java 11 will take quite some effort apart from simply changing the JRE in the eclipse settings. You can take a look at the necessary steps here: https://www.optimaw.ch/change-to-jdk11-lts/

ruspl-afed:
> I'm afraid, but changing the required Java runtime to Java 11 will take quite some effort apart from simply changing the JRE in the eclipse settings. You can take a look at the necessary steps here: https://www.optimaw.ch/change-to-jdk11-lts/

Changing to Java 11 looks like a breaking change for downstream. I suggest to keep Java 8 for the nearest release and then plan the switch to Java 11 at mailing list.

wimjongman:
> Changing to Java 11 looks like a breaking change for downstream. I suggest to keep Java 8 for the nearest release and then plan the switch to Java 11 at mailing list.

Yes, it is important that we pick the right version. But how do you define downstream?

Eclipse requires Java 11 since 4.17. BIRT 4.8 still runs on Eclipse 4.16?

wimjongman:
.. and what are the thoughts about the code clean-up suggestion?

ruspl-afed:
> Yes, it is important that we pick the right version. But how do you define downstream?

I define the `downstream` is a set of component both inside and outside the SimRel that consumes Eclipse BIRT as a part of its target. Switching to Java 11 will be a breaking change for these component and should be discussed separately.

> Eclipse requires Java 11 since 4.17.  BIRT 4.8 still runs on Eclipse 4.16?

Current Eclipse BIRT has Eclipse 4.11 as a target.

This is why I asked about target platform for the nearest release in the mailing list, let's discuss there.

ruspl-afed:
> .. and what are the thoughts about the code clean-up suggestion?

I like new format, however I haven't considered every option in detail. I suppose it is close to the Platform settings, right?

May be we can have these set of "sample settings" somethere outside of the real bundle to be discussed separately (before applied to the bundles).
In CDT we have a script that applies "sample settings" to every patch set and fails if the tree is dirty after that.

Issue #603, opened 2021-03-11
Set versions to 4.9


wimjongman:
All versions must be set to 4.9.0

PR #604, opened 2021-03-11
Set versions to 4.9 #603

wimjongman:
Mass replace versions from 4.11 (never released) to 4.9


wimjongman (commented):
A lengthy patch but it looks good to me!

PR #608, opened 2021-03-12
Update ci.yml

wimjongman:
Use verify instead of install since there is no local maven to use anyway. Might save a couple of milliseconds

@ruspl-afed @Flugtiger


Flugtiger:
The artifacts that get installed into the local maven repo are used by the subsequent test stages, so mvn install is unfortunately necessary.

wimjongman:
Got it! Thanks.

Issue #609, opened 2021-03-12
Prepare for Java > 8


wimjongman:
In order to compile with recent java compilers, the dependencies need to be a little more explicit in some cases.

wimjongman:
Fixed!

PR #610, opened 2021-03-12
Prepare for Java > 8 #609

wimjongman:
Added javax.xml in order to not use w3c.dom classes from the
jdk.


wimjongman:
I have prepared the build so that we can run it on a JDK > 8.

In addition, I fixed some additional test suites: Chart and Engine

ruspl-afed (approved):
I think this one should be merged. A lot of manifest were upgraded and we have more tests enabled. Great work @wimjongman

Issue #612, opened 2021-03-12
Pull request builds 2 times


wimjongman:
See below. why does it build twice?

![image](https://user-images.githubusercontent.com/180969/110995141-241c1100-837a-11eb-95c8-386aba026f95.png)

Flugtiger:
I guess it builds the first time, because it's a push to a branch in this repo and the second time, because there is a pull request on this branch. To make a pull request one should use a branch in ones fork and not in the official repo.

wimjongman:
Can you fix it so that it only builds when there is a push to master?

wimjongman:
Fixed!

PR #614, opened 2021-03-15
Pull request builds 2 times #612

wimjongman:
Fixed to only build on push when pushed to master

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,8 +12,11 @@
 ###############################################################################
 
 name: CI
-
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
 
 jobs:
   build:


ruspl-afed (approved):
Thanks @wimjongman, looks reasonable. 
Let's see how it will work for subsequent pushes to the source branch of PR

wimjongman:
Maybe we don't have to build when pushed to master because we should never push directly. It is too late then anyway because the change has already been merged. The main build will pick up the build and deploy it.

(the main build is green by the way 💪)

ruspl-afed:
> Maybe we don't have to build when pushed to master

I would keep it for a while. I have an impression that it is triggered after PR merged to master.

wimjongman:
We can keep it but it is not very useful. Deployment build is here:

https://ci.eclipse.org/birt/

ruspl-afed:
> but it is not very useful

For me it is a kind of proof that "build from sources" works in `master`

> Deployment build is here:

Thank you for the link!

PR #624, opened 2021-03-24
Space is encoded in dataset parameter #623

SteveSchafer-Innovent:
Redo fix to bug 456816 (CVE-2019-11776) to html-encode format parameter in the jsp rather than encoding all parameters which was causing problems in prepared queries.

### DIFF
--- a/viewer/org.eclipse.birt.report.viewer/birt/WEB-INF/classes/org/eclipse/birt/report/utility/ParameterAccessor.java
+++ b/viewer/org.eclipse.birt.report.viewer/birt/WEB-INF/classes/org/eclipse/birt/report/utility/ParameterAccessor.java
@@ -2104,7 +2104,7 @@ public class ParameterAccessor
 			{
 			}
 		}
-		return htmlEncode( request.getParameter( parameterName ) );
+		return request.getParameter( parameterName );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


waynebeaton:
AFAICT, Steve is good-to-go as far as ECA/Signed-off-by requirements are concerned. It appears that Steve has two accounts with the same email address and that's confusing the ECA checker. I'll get the webmaster to sort it out.

In the meantime, committers can ignore the failed "eclipsefdn/eca" check.

SteveSchafer-Innovent:
Thanks Wayne.  The CI is failing unit tests with a bunch of errors I don't understand, but I think unit tests aren't working right now so hopefully this isn't a problem.

wimjongman (approved):

wimjongman:
Looks good to me. Thanks, @SteveSchafer-Innovent

ruspl-afed:
> The CI is failing unit tests with a bunch of errors I don't understand, but I think unit tests aren't working right now so hopefully this isn't a problem.

We are trying to restore tests part by part. 
Currently we have only two test bundles active, you can have a look [here](https://github.com/eclipse/birt/blob/master/.github/workflows/ci.yml#L54)
And these tests started to fail after this change.
It would be great to understand why.

PR #627, opened 2021-03-30
Support for Word 2016

hvbargen:
Newer releases of MS Word (starting from Word 2016) in certain cases calculate the indent differently when opening documents made for older version of MS Word (like those created by BIRT).

This causes text to be misaligned by Word for documents created with the old word emitter or the DOCX emitter.
Furthermore, Word shows a hint that the document is opened in "compatibility mode".

Now the default is to create Word 2016 output with the DOCX emitter, which solves these problems.


ruspl-afed:
The test failure seems to be not related to this change.
But eclipsefdn/eca check is important. 
Do you have signed Eclipse Contributor Agreement? Did you point to your GitHub account in your Eclipse profile?

wimjongman:
> Henning Von Bargen (h.vonbargen@****.com) did not include the "Signed-off-by footer" 
> which is required for all commits made by a contributor.

Don't worry about that, we are going to drop that requirement this week. We can accept the patch if the build is green.

wimjongman:
> So I'm a bit lost here. How can I create a PR which eca will accept?

@hvbargen ^^ Don't worry about it, Eclipse has lifted this requirement literally today:

https://www.eclipse.org/lists/eclipse.org-committers/msg01291.html

hvbargen:
BTW in https://bugs.eclipse.org/bugs/show_bug.cgi?id=565629 I have attached a sample report which demonstrates the issue (Run->View Report->As DOCX).
Is there anything more I need to do?
I (re-)signed the Eclipse commiter licence agreement a few days ago with the Eclipse account which is linked to my hvbtup GIT account.

wimjongman:
Thanks, Henning!

Issue #629, opened 2021-04-01
Dragging column widths with the mouse working "opposite direction"


hvbargen:
This is quite hard to explain with words.
Say you have created a new 2-column table item.
If you try to drag the column widths with the mouse to make e.g. the first column ca. 3 cm wide, this results in the first column having a width of (available-width - 3 cm).

This did not happen in BIRT 4.2.1.
OTOH, in BIRT 4.2.1, dragging the column width of *any* column caused *every* column to have a fixed width - and the table as a whole, too.
Now, in the 2-column example, the second column has no specified width - which is good.
When coaching, I always tell users to *not* use the mouse to set column widths, because I think it's usually much better if the table and at least one column do not have a specified width. With newer BIRT releases, this recommendation is no longer as important as before.

I guess this issue and the changed behavior in comparison to 4.2.1 are closely related.

I'll try to fix this myself after I'll have created PRs for some other bugs.

hvbargen:
I reckon the issue is caused by a change in https://github.com/eclipse/birt/commit/7dd84fd56c08a5751b956cc52c207c42696ce3f3 in file ColumnDragTracker.java.
I'll  try to fix this.

hvbargen:
The fix for this bug is included in https://github.com/eclipse/birt/pull/630 - I forgot to create a new branch and a new PR.

wimjongman:
Please create separate PR's in the future.

PR #630, opened 2021-04-01
Fix BIRT bug 562873 PDF Rendering error with dynamic text item after …

hvbargen:
…page break.
For more info, see https://bugs.eclipse.org/bugs/show_bug.cgi?id=562873


ruspl-afed:
I can only say that code looks better after your fix

hvbargen:
I cannot imagine how the failing test DocArchiveLockManagerTest.testThread could be caused by this PR.
And the second issue #629 (which landed in this PR by accident) is immediately visible to a BIRT-Designer user, so I think it is important to get this into the next release.

PR #634, opened 2021-04-06
Update .settings files.

SteveSchafer-Innovent:
Update .settings files. #634

"org.eclipse.core.runtime.prefs"
"org.eclipse.jdt.core.prefs"
"org.eclipse.jdt.ui.prefs"
"org.eclipse.pde.api.tools.prefs"
"org.eclipse.pde.prefs"
were copied from eclipse.platform.ui/bundles/org.eclipse.ui/.settings
to each .settings folder where the parent folder contained a .project file
and the .project file contained a javanature.
.settings folders were created where necessary.
All existing properties were retained.
Additional properties were set in order to relax error settings.
This was accomplished with the bash script copy-settings which I included for reference at the root level.
I have relaxed the error settings to the point where I only have 16 errors which is the lowest I've been able to get it.
I'm still unsure if I'm importing the code correctly (see issue 628).


SteveSchafer-Innovent:
The additional settings to relax error levels are documented at the bottom of the copy-settings bash script.

wimjongman:
Can this go?

ruspl-afed:
+1 from my side
Please note that build log increased significantly in size since we have much more issues reported

Issue #641, opened 2021-04-16
XLSX tab naming can cause IllegalArgumentException


SteveSchafer-Innovent:
The spudsoft emitter uses the table of contents entries for tab naming.  POI and/or Excel has some restrictions on what characters can comprise a tab name, and this can lead errors that can be data-dependent.  Additionally there is a size limit on the name but the emitter code doesn't take that into account when searching for duplicate names, which can also lead to errors.  The latter issue is definitely a bug.  For the first issue I think it would be acceptable for the emitter to edit the names before setting them, although this might be a matter of opinion.   Please leave comments as to whether you agree.

Related bugzilla: https://bugs.eclipse.org/bugs/show_bug.cgi?id=441382

I can implement these fixes but I don't know of yaytay (spudsoft author) is still involved.

hvbargen:
Steve, in my fork at https://github.com/hvbtup/birt I updated the XLSX emitter to the latest version of Jim's (yaytay) code when it was still available. The repository was at bitbucket.org but is no longer available. 
https://github.com/Yaytay is very likely the same person - I don't know if he's still interested in BIRT.
I don't know if this might help with this issue, I just wanted to let you know.

I stumbled  over this issue myself once, and I think it's OK if the emitter replaces invalid characters with a replacement characters (for example, "_") or removes them, and if the emitter truncates the name.
Anyway, if the emitter has to change the name, this should be logged.

This is much better than a lengthy error stack trace.

SteveSchafer-Innovent:
PR merged and closed so closing this issue

PR #650, opened 2021-05-03
XLSX tab naming can cause IllegalArgumentException #641

SteveSchafer-Innovent:
The spudsoft emitter uses the table of contents entries for tab naming. POI and/or Excel has some restrictions on what characters can comprise a tab name, and this can lead errors that can be data-dependent. Additionally there is a size limit on the name but the emitter code doesn't take that into account when searching for duplicate names, which can also lead to errors.

### DIFF
--- a/engine/uk.co.spudsoft.birt.emitters.excel/src/uk/co/spudsoft/birt/emitters/excel/HandlerState.java
+++ b/engine/uk.co.spudsoft.birt.emitters.excel/src/uk/co/spudsoft/birt/emitters/excel/HandlerState.java
@@ -20,6 +20,7 @@ import java.util.Map;
 
 import org.apache.poi.ss.usermodel.Sheet;
 import org.apache.poi.ss.usermodel.Workbook;
+import org.eclipse.birt.core.exception.BirtException;
 import org.eclipse.birt.report.engine.api.IRenderOption;
 import org.eclipse.birt.report.engine.api.ReportEngine;
 import org.eclipse.birt.report.engine.emitter.IContentEmitter;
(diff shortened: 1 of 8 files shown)


hvbargen (commented):
Looks good to me.

But can you explain why the name is truncated to 26 characters when 31 are allowed?
Your comment says "max length is 31 but leave room for digits"
Where are the digits added and why max. 5 digits?

This might break existing reports which are using eg. 30 characters to create unique sheet names.

hvbargen:
> I can revisit the idea of only shortening the names if necessary if you 
> think that would be better.

What about an approach like this:

Allow two algorithms A and B. Which one to use can be set as an emitter 
option, defaulting to A.

B works like you already coded it and described it here.

A shortens the name only if necessary.

I don't remember the details of the emitter, but IIRC when a page break 
occurs, the emitter infers an intended sheet name N from the data.
This name N cannot always be used as-is, because sheet names must be 
unique, max 31 characters long etc.

Both algorithms A and B would first purify N (removing illegal characters, 
then cut after 31 characters), resulting in a possible sheet name P.

Algorithm B (but not A) would cut P to 26 characters here.

Next test if P already has already been used (and thus would result in 
duplicate sheet names).

If P is unique, fine. Use it and add it to the set of used names.

Otherwise we need numbering to make the name unique.

With algorithm A, cut P to 26 now.

In a loop from n=1 to 9999, try the name P + space + n (formatted with 
leading zeroes eg 0001). Break and use that name if it is free.

If the loop ends without a break, raise an exception.

This could be made more effective by using a Map name -> n instead of a Set 
with the names.

So, the difference between the algorithms is just the time when we cut the 
length to 26.

SteveSchafer-Innovent:
Thanks for the suggestion.  I think in the interest of simplicity there should be only one algorithm and in the interest of not changing things any more than necessary,  it would be better to not pad the numbers with leading zeroes.  So I'm thinking of only shortening the names when necessary.  The prepareSheetName method already uses a map.  If the original name collides with an existing one, and I need to shorten it, I'll be introducing a new name which may or may not collide with a name already in the map.  If it doesn't collide (and I've shortened it) then I'll start it out with index 1, otherwise I'll increment the index.  I'll only shorten the name enough for the space and however many digits are necessary.

When I shorten the name, I could add an ellipsis (the period is not an illegal character).   That would waste three characters but it would make it more clear that the name has been shortened.  However it introduces a situation where the tab name would be different on an already working report, so I think I won't do that.

If someone has designed a report to take the 31 character limit into account and the report runs successfully then anything I do to shorten the names will have no affect on that report, because the names are already short enough.

hvbargen:
Sounds good to me.

SteveSchafer-Innovent:
@krishnakittu I used strings so I could replace with nothing.  Possibly there's another way that can be done.

Build is failing.  I'm not clear how that's related to what I changed but will investigate.

wimjongman:
The build is generating web.xml. Do we need to push it?

wimjongman:
> Build is failing. I'm not clear how that's related to what I changed but will investigate.

It looks accidental. I have restarted the build.

wimjongman:
The build is failing in subsequent passes. This means that tests are not repeatable and they rely on some intermittent results. The test `"ElementDefnTest.testGetGroupNames:84 expected:<6> but was:<0>"` looks related to your patch.

 ### Failures: 
org.eclipse.birt.report.model.command.ContentCommandTest.testCopyAndPaste(org.eclipse.birt.report.model.command.ContentCommandTest)
  Run 1: ContentCommandTest.testCopyAndPaste:1617
  Run 2: PASS
  Run 3: PASS

org.eclipse.birt.report.model.metadata.ElementDefnTest.testGetGroupNames(org.eclipse.birt.report.model.metadata.ElementDefnTest)
  Run 1: PASS
  Run 2: PASS
  Run 3: ElementDefnTest.testGetGroupNames:84 expected:<6> but was:<0>

hvbargen:
I don't understand the testGetGroupNames at all. Why is the magic value 6 expected?
Regarding web.xml: Maybe it is just shuffled, I mean the content is logically the same, only in a different (non-)order?

engine/uk.co.spudsoft.birt.emitters.excel/src/uk/co/spudsoft/birt/emitters/excel/HandlerState.java
```diff
@@ -289,20 +290,83 @@ public int findRowsSpanned(int rowX, int colX) {
...
+			if (correction <= 0) {
+				this.sheetNames.put(sheetName, nameCount);
+				return preparedName;
+			}
+			if (correction > sheetName.length()) {
+				throw new BirtException(EmitterServices.getPluginName(),
+						"Unable to fit sheet name into the maximum allowed length", //$NON-NLS-1$
+						null);
+			}
+			sheetName = sheetName.substring(0, sheetName.length() - correction);
```
hvbargen, line 367: My gut feeling says there must be a bug somewhere in this code.
The attribute sheetName is modified here, but nameCount is unchanged. Shouldn't nameCount be re-initialized then?
Not sure if I grep the code, my brains seems to go into stand-by after a hard working day...

SteveSchafer-Innovent: I thought about that.  At first I reset the name count to zero but then I thought that it might be valuable to have some indication that the tab is an increment of the previous tab even though the name is shorter.  I had considered adding ellipses to make that more clear but realized that might break existing reports, so I settled on just letting the index increment.  Make sense?

SteveSchafer-Innovent: Regarding web.xml, it looks like that file is generated.  I pushed it in hopes it had something to do with the failures but no luck.

SteveSchafer-Innovent:
I'm trying to run the test locally with the same mvn command as in the CI in order to be able to use the eclipse debugger but the test is failing for different reasons (even on the master branch, freshly pulled).  I'm using OpenJDK 11.

`Running org.eclipse.birt.report.model.metadata.ColorPropertyTypeTest
Tests run: 15, Failures: 0, Errors: 15, Skipped: 0, Time elapsed: 0.023 s <<< FAILURE! - in org.eclipse.birt.report.model.metadata.ColorPropertyTypeTest
testValidateXml(org.eclipse.birt.report.model.metadata.ColorPropertyTypeTest)  Time elapsed: 0.004 s  <<< ERROR!
org.eclipse.birt.report.model.api.DesignFileException: Syntax error found, and see error details.
	at org.eclipse.birt.report.model.metadata.ColorPropertyTypeTest.setUp(ColorPropertyTypeTest.java:34)
`
15 more errors all on the same line of code (openDesign):

`	protected void setUp() throws Exception {
		super.setUp();
		openDesign("ColorPropertyTypeTest.xml", TEST_LOCALE); //$NON-NLS-1$
	}
`

wimjongman:
Steve, I believe local test failures are caused by using different locales. Also, not all tests are active on the server ATM:

https://github.com/eclipse/birt/blob/master/.github/workflows/ci.yml

SteveSchafer-Innovent:
@wimjongman does that mean my maven operation is using a different locale than the CI?  How do I tell what locale the CI is using?  I don't see any mention of locale in ci.yml.

[innovent@kirk birt]$ mvn --version
Apache Maven 3.6.3 (Red Hat 3.6.3-8)
Maven home: /usr/share/maven
Java version: 11.0.11, vendor: Red Hat, Inc., runtime: /usr/lib/jvm/java-11-openjdk-11.0.11.0.9-0.fc34.x86_64
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "5.11.17-300.fc34.x86_64", arch: "amd64", family: "unix"

I'm running the same test that fails in CI, the first one listed in ci.yml:

mvn -B -pl :org.eclipse.birt.report.model.tests integration-test

SteveSchafer-Innovent:
I reverted the branch to its original state and added two static variables and two functions to HandlerState and it's still failing a unit test.  I think I need to start over from scratch.

SteveSchafer-Innovent:
I have verified that this branch is now identical to the BIRT master branch (the only differing files are some jar files and some eclipse prefs files that exist in master but not here) and yet the CI is still failing.

SteveSchafer-Innovent:
Since this PR was used to debug the CI, I'm closing it and will open a new one for issue 641.

Issue #651, opened 2021-05-12
Batik vulnerability


SteveSchafer-Innovent:
CVE-2018-8013 appears to be addressed by JIRA issue BATIK-1222 which changed AbstractDocument.java which is part of the batik.dom jar.  It was fixed in batik 1.10 but the version of that jar used by BIRT is 1.9.  Unfortunately neither SVN comments nor JIRA issues specifically reference CVE numbers, but I'm pretty sure this was the fix.  Recommend upgrading to batik 1.13.  I'm already in the process of doing this in 4.8 and will apply the change to 4.9 when I'm done.

SteveSchafer-Innovent:
PR merged and closed so closing this issue.

PR #654, opened 2021-06-18
Batik vulnerabilities #651

SteveSchafer-Innovent:
Update batik plugins to version 1.14.
    
### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ https://eclipse.org/birt
 [![Build Status](https://github.com/eclipse/birt/workflows/CI/badge.svg)](https://github.com/eclipse/birt/actions)
 
 ## Building BIRT
-BIRT is built with [Apache Maven](http://maven.apache.org).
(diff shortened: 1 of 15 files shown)


TheLaserNinja:
Good news is the build completes. The bad news is it fails during tests. :/

wimjongman:
The tests complete normally in other builds. Is the test fail related to this change?

ruspl-afed:
> The tests complete normally in other builds. Is the test fail related to this change?

It could be since we have something related to the images in the failure trace. I also can see another suspect entry:
```
WARNING: org/apache/batik/w3c/dom/ElementTraversal
java.lang.NoClassDefFoundError: org/apache/batik/w3c/dom/ElementTraversal
```

SteveSchafer-Innovent:
@ruspl-afed, good catch.  It passes now that I added the jar that contains the definition of ElementTraversal.

PR #656, opened 2021-06-27
Fixed a bug with IndexOutOfBoundsException when PDF wordbreak option …

hvbargen:
…is set and whitespace:nowrap property ist set.

A simple test report with a label containing text "Blablabla BlablaBla BlaBlaBla BlaBlaBla Blablabla" (Arial, 10pt) on a 50mm wide page causes BIRT to crash when

        IPDFRenderOption options = new PDFRenderOption();
        options.setOption("pdfRenderOption.wordBreak", true);

is used and the label has 

            <property name="whiteSpace">nowrap</property>

This PR fixes that bug.


hvonbargen:
The failing CI tests are totally unrelated to the changes here.
I confirmed the bug and the bug fix locally, furthermore this code is in production since 1 week without problems.

wimjongman:
Thanks!!

PR #439, opened 2018-02-17
Added support for PDF appending. Due to a number of customer requests to

TheLaserNinja:
have the ability to append boiler plate legal documentation or financial
statements to a report, the ability to append PDF files based on a
comma-seperated list or a ArrayList defined in the Report Context has
been added.

This has been tested in actual customer environments and works. To set:
-Create a BIRT Report
-Create a User Property called AppendList.
-Set AppendList under Advanced to a Comma-Seperated list of paths to PDF files to append

-Alternatively, create an ArrayList of String in the report context set in global variable called appendPDF.

PR #661, opened 2021-07-19
Enable relative paths for PDF append and prepend

bkiselka:
enable relative or universal paths by using context.getResource()

and do not print a "1 of 1" at the bottom of an added page by setting paginate = false for concatPDFs()
Reasoning: the prepended PDF is usually a cover page without pagination and the appended an annex also without a pagination

This relates to #439 and #567


TheLaserNinja:
Let me commit @SteveSchafer-Innovent fixes for CI/CD. Once confirmed, I will merge.

bkiselka:
> Let me commit @SteveSchafer-Innovent fixes for CI/CD. Once confirmed, I will merge.

Thanks you very much!

Issue #662, opened 2021-07-23
Axis vulnerabilities


SteveSchafer-Innovent:
I'm aware of the following vulnerabilities related to axis 1.4:

CVE-2019-0227 | Server Side Request Forgery (SSRF)
CVE-2018-8032 | Cross-site scripting (XSS)
CVE-2014-3596 | SSL spoofing
CVE-2012-5784 | SSL spoofing
CVE-2007-2353 | Request non-existent WSDL file reveals AXIS installation path

Axis 1.4 is used for SOAP communication in the viewer.  The Orbit plugin org.apache.axis_1.4.0.v201411182030.jar embeds axis.jar which hasn't changed since 2006.  A github repository for axis 1.4 has been kept up to date however: https://github.com/apache/axis-axis1-java which contains fixes for all of these bugs except possibly CVE-2007-2353.  The jar file can be rebuilt from this repository and I have done this successfully for one of our clients.

I'm not sure what the right approach is for this.
- Build the jar from source and update the Orbit plugin (how to do that?)
- Build the jar from source and embed it in the BIRT repo without using the Orbit plugin (how to do that?)
- Convert to axis 2.  This is may be a rabbit hole.

I welcome comments.

SteveSchafer-Innovent:
I think I'm going to try creating a new axis plugin in the common directory, similar to org.w3c.sac.  I'll build axis.jar from the github repo and put it in a lib folder and export everything necessary in an OSGI manifest.  So it will be pretty much the same as the current Orbit bundle except with a newer jar.  If this works, an Orbit bundle could be created from it.  It requires a committer to do that.

wimjongman:
Steve, I am working on creating a new build (#640) from a target platform. The build already works. The test currently fails.

With this new build, we can provide bundles from maven central. [1]

[1] https://mvnrepository.com/search?q=axis

SteveSchafer-Innovent:
Good to know, but the axis on maven central is from 2006.  It's necessary to build from the source on github to get the vulnerability fixes.

SteveSchafer-Innovent:
Closing because pull request has been merged.

tsiddiq:
Hi Steve,
[...]

PR #664, opened 2021-08-06
Axis Vulnerabilities #662

SteveSchafer-Innovent:
Use axis.jar built from the github axis-1 repo.


SteveSchafer-Innovent:
What I did here was build axis-1.4.1-SNAPSHOT.jar from the github https://github.com/apache/axis-axis1-java repository and create a plugin in the common area that contains that jar in a lib folder and a manifest that exports all the packages.  I also changed the ant build script in org.eclipse.birt.api to copy the jar file into the target/lib folder so it gets included in the engine build.

Issue #660, opened 2021-07-19
java.lang.IllegalArgumentException: Invalid version number: Version number may be negative or greater than 255


Adi-raj-2021:
Hi Team,
 
I have a query related to this issue
“java.lang.IllegalArgumentException: Invalid version number: Version number may be negative or greater than 255"
 
Let me brief you few analysis part
As currently we are using "com.ibm.icu_3.6.1.v20070906.jar" jar in our project.
This jar is getting used in reporting tool(Birt report) of our product.
 
In recent months we upgraded our product java to 8.0.6.15 version but our reporting tool throws this error
*************
Caused by: java.lang.IllegalArgumentException: Invalid version number: Version number may be negative or greater than 255
at com.ibm.icu.util.VersionInfo.getInstance(VersionInfo.java:194)
at com.ibm.icu.impl.ICUDebug.getInstanceLenient(ICUDebug.java:65)
*************
 
At previous , I approached icu-support with short analysis over this issue.
we got response at that time i.e "this issue fixed in icu4j.jar(of version 68).So you can use this version of jar".
 
Later we analysed more and test , found we are using "com.ibm.icu_3.6.1.v20070906.jar" in the reporting tool not icu4j.jar.
Further we upgarded our java version to 8.0.6.20,8.0.6.25 & 8.0.6.30. Finally we are using java 8.0.6.30 now. And still our birt repoting tool throwing the same error which we got on 8.0.6.15 version of java.
 
Although the error was in Birt module so I checked with Eclipse support who managed this Birt project. But they directed me that this is an ICU4J issue.
 
As we analysed deeply the issue and found the source of exception is from versionInfo.class. We know there is a check placed in versionInfo.class to check the minor/macro version of java and it hardcoded as 255.
But java version start from 8.0.6.15(261) exceeding the limit of 255.It is the main source of issue.
As downgrading java version 8.0.6.10 also works here to solve this issue , but it is not good approach.
For testing scenario , we modified the code of versionInfo.java and replace the 255 check to some higher version such as 301 or 311.
And our testing results show no error afterwards.
 
As per your old communication i.e this issue fixed in 68 version of icu4j.jar
I tried to find out the references ,
https://unicode-org.atlassian.net/browse/ICU-21295
https://github.com/unicode-org/icu/pull/1245
https://unicode-org.atlassian.net/browse/ICU-21219
[...]

PR #665, opened 2021-08-15
Version number may be negative or greater than 255 #660

SteveSchafer-Innovent:
java.lang.IllegalArgumentException: Invalid version number: Version number may be negative or greater than 255 #660

Specify com.ibm.icu version 63.1 in target orbit R20190226160451 location and explicitly specify icu version every place it's included in ant build scripts (it doesn't appear in any maven dependencies).


ruspl-afed:
LGTM

wimjongman:
Why is it needed to explicitly name that version? Why would it not automatically use the highest version?

viewer/org.eclipse.birt.report.viewer/birt/WEB-INF/web.xml
```diff
@@ -247,23 +247,23 @@ Automatically created by Apache Tomcat JspC.
 
 
     <servlet>
-        <servlet-name>org.apache.jsp.webcontent.birt.pages.common.Error_jsp</servlet-name>
-        <servlet-class>org.apache.jsp.webcontent.birt.pages.common.Error_jsp</servlet-class>
+        <servlet-name>org.apache.jsp.webcontent.birt.pages.common.processing_jsp</servlet-name>
```
wimjongman, line 250: Steve, this does not seem related to issue #660, does it?

ruspl-afed:
> Why is it needed to explicitly name that version? Why would it not automatically use the highest version?

I never heard about such kind of functionality for Apache ANT.

wimjongman:
> > Why is it needed to explicitly name that version? Why would it not automatically use the highest version?
> 
> I never heard about such kind of functionality for Apache ANT.

I mean, why should this work

`				<include name="org.w3c.css.sac_*.jar" />`
`				<include name="com.ibm.icu_63.1.0.v20181030-1705.jar" />`

and not this:

`				<include name="org.w3c.css.sac_*.jar" />`
`				<include name="com.ibm.icu_*.jar" />`

ruspl-afed:
> I mean, why should this work
> 
> ` <include name="org.w3c.css.sac_*.jar" />`
> ` <include name="com.ibm.icu_63.1.0.v20181030-1705.jar" />`
> 
> and not this:
> 
> ` <include name="org.w3c.css.sac_*.jar" />`
> ` <include name="com.ibm.icu_*.jar" />`

Yes, I understand. I have no idea which ICU4J jar will be used by ANT if we have several of them that matches the pattern. All of them are just files to be added to classpath, right?
Since we cannot guarantee that one and only one ICU4J jar will be available during classpath resolution - I think it is good idea to mention the full jar name explicitly.
Of cause this is not the best solution, since we "hardcoding" the exact version in the dozen of places - but this is the price we will pay for non-optimized build scripts that we currently have.

ruspl-afed:
@wimjongman am I answered your question or should I try again?

wimjongman:
> 
> 
> @wimjongman am I answered your question or should I try again?

Hi Alexander. I read your answer but I don't understand why there would be any other ICU version to copy than the most recent one.

ruspl-afed:
> > @wimjongman am I answered your question or should I try again?
> 
> Hi Alexander. I read your answer but I don't understand why there would be any other ICU version to copy than the most recent one.

To be honest I don't know how many ICU jars we have available during the build. But I assume that after the "tycho part" of the build we _may_ have more than one ICU jar as a result of dependencies resolution.
Then we launch several ANT build that points to the directory with jars and says "pick an ICU jar for me, any version will match"
If we have several ICU jars there - which one will be used? To resolve this ambiguity the needed ICU jar specified precisely.

wimjongman:
Steve is this still needed?

wimjongman:
@SteveSchafer-Innovent is this still needed?

wimjongman:
Closing because of no reaction fof @SteveSchafer-Innovent. Please reopen if needed.

Issue #674, opened 2021-10-30
Do not embed fop but add it as a dependency


wimjongman:
Apache fop is embedded but it should be included.

wimjongman:
IP REVIEW: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23815#c3

ruspl-afed:
@wimjongman Wim, I pinged them but still no movement there. Should we ask @waynebeaton to intensify this a bit?

waynebeaton:
> Should we ask @waynebeaton to intensify this a bit?

I guess that this question answers itself. :-)

I'll see what I can do.

ruspl-afed:
@wimjongman please have a look at IP Team reply https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23815#c9

TBH I don't know the origin of the JAR that we currently have in our repository. But anyaway, are you OK if we remove some files from there?

wimjongman:
the review has been completed.

PR #675, opened 2021-10-30
Do not embed fop but add it as a dependency #674

wimjongman:
Moved FOP to target platform


wimjongman:
Researching the CQ for fop, I found this:

https://cwiki.apache.org/confluence/display/XMLGRAPHICSBATIK/PdfTranscoder

PR #678, opened 2021-11-01
Fix "Import=Package" directives to avoid compilation errors

ruspl-afed:
Add
org.dom4j.tree;version="2.1.1" 
to "org.eclipse.birt.build" bundle

Add 
org.apache.fop.svg;version="2.6.0" 
to "org.eclipse.birt.chart.device.pdf" bundle

Signed-off-by: Alexander Fedorov <alexander.fedorov@arsysop.ru>


wimjongman (commented):
Please hold. fop is currently embedded. When you run the build locally fop 2.6.0 will be automatically added. I have a separate PR aiming to fix this:

https://github.com/eclipse/birt/pull/675

ruspl-afed:
@wimjongman I removed the change regarding FOP, but it seems that we still need the change regarding `org.dom4j.tree`

PR #684, opened 2021-11-02
Do not embed fop but add it as a dependency #674

wimjongman:
Trying with inclusion of fop


wimjongman:
Alexander, this is a solution that includes fop jar without having to build first. Maybe we can accept this as a temp solution so that you can also release the oomph setup. WDYT?

ruspl-afed:
> Alexander, this is a solution that includes fop jar without having to build first. Maybe we can accept this as a temp solution so that you can also release the oomph setup. WDYT?

My only concern is CQ, do we have it for Apache FOP 2.6?

The statement [here](https://xmlgraphics.apache.org/fop/license.html) says 

> Apache FOP comes with a number of libraries all of which are currently coming from within the Apache Software Foundation. These libraries are all licensed under one of the Apache License versions. There is one exception which is "xml-apis.jar" which contains code licensed by the W3C. Please consult the license files included with the libraries in the "lib" directory for details.

"one of the Apache License versions" sounds good but I know nothing about W3C.

wimjongman:
We can't release it before we have a CQ. However, we are in the development phase. I suggest filing a CQ once we know how we consume it.

ruspl-afed:
> We can't release it before we have a CQ. However, we are in the development phase. I suggest filing a CQ once we know how we consume it.

@wimjongman I created [CQ](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23815) for Apache FOP 2.6 binary since it may take time to get an approval. I think we will need to include FOP in any case. If you don't think so you can use "Withdraw this CQ" action.

wimjongman:
Brilliant! Thanks. Maybe add the link to the issue?

ruspl-afed:
@wimjongman I think we should label this issue somehow to not forget about fop 2.6 inside that is not yet approved by IP team.

Issue #685, opened 2021-11-02
Fix chart core tests


wimjongman:
The chart core tests require some magic to happen. See PR.

wimjongman:
I can't get this to work in CI but the tests are running locally. For now, I have disabled the test in the workflow.

I suggest taking the code as it is and trying to fix it for CI later.

PR #686, opened 2021-11-02
Fix chart core tests #685

wimjongman:
Added fragment to provide dependency [1]
Fixed some Locale dependent tests

[1]
The chart core tests were failing due to an incompatible
java/osgi construct in org.apache.batik.transcoder.

The transcoder "bundle" is doing a Class.forname(String)
on a class that is not listed as a dependency.

This happens in class
org.apache.batik.transcoder.image.PNGTranscoder#getWriteAdapter
which is looking for
"org.apache.batik.ext.awt.image.codec.png.PNGTranscoderInternalCodecWriteAdapter"

To fool this "bundle", a fragment bundle is introduced with
o.a.b.transcoder as the master plugin that depends on the
"org.apache.batik.codec" bundle and in this way provides the dependency.


wimjongman:
Alexander, I can't get it to work in the CI build. However, with this patch, we can at least run the test locally.

PR #687, opened 2021-11-04
Revert "Do not embed fop but add it as a dependency #674"

wimjongman:
Reverts eclipse/birt#684

This somehow breaks the build


wimjongman:
Closing, since I forgot what this was.

Issue #699, opened 2021-11-10
Add resource filters to parent projects


wimjongman:
In BIRT there are several parent projects which cause projects to be seen by eclipse at multiple levels. This is causing freezes and a general confusion inside Eclipse.

merks:
@wimjongman 

I looked at your video and understand how to create a fork.  But please excuse my ignorance.  What's not clear is how work with that in Eclipse.  I found this question with a description:

https://stackoverflow.com/questions/53024406/how-to-work-on-a-project-fork-when-having-the-original-in-eclipse

So I think I could follow those instructions (add a remote to pushes to my fork URI) but is this really the simplest/correct approach?

My concern is that I make a bunch of changes, commit them to my fork, but maybe I forget something and then I have multiple commits.  Will multiple commits be a problem?  

Also, what happens if someone commits other changes after I have forked?  How can I rebase my fork/branch if that happens?

Again, sorry if these are stupid questions...

wimjongman:
> Again, sorry if these are stupid questions...

After all these years of learning from you, I'm happy to return the favor. :)

> but is this really the simplest/correct approach?

Yes, after you have already cloned the main repo, this is the easiest approach. Detailed steps are:

* Fork birt main repo to your own account on the Github website
* git remote add fork https://github.com/merks/birt.git
* git remote -v

If you did not already clone the master birt repo then you should just clone your fork:

* git clone https://github.com/merks/birt.git

### Will multiple commits be a problem?
Just make sure you put all your commits in a topic branch and you can push as many commits as you like. After you made a pull request, even pushes after that will be automatically added to the pull request.

We can squash before we do the final merge.

### what happens if someone commits other changes after I have forked
The easiest way is to go to your fork on the GitHub website and rebase your forked repository. Then fetch and merge/rebase locally.

![image](https://user-images.githubusercontent.com/180969/141124910-043d58ea-b7b4-419e-b18a-0b065055166c.png)

merks:
Thank you!  That wasn't so hard!!

I created this pull request to test the process before adding filters to the various projects with nested projects.

https://github.com/eclipse/birt/pull/700

merks:
I created this pull request for this issue:

https://github.com/eclipse/birt/pull/702

Here's how it looks:

![image](https://user-images.githubusercontent.com/208716/141267665-ab273fd4-e897-4ea8-95e2-82c31d9aa357.png)

[...]

PR #702, opened 2021-11-11
Add resource filters to parent projects #699

merks:
In general, for projects that contain nested projects, filter the
folders corresponding to such nested projects such that the nested
contents of those folders are not shown in the resource tree, although
the folder itself is shown.  Of course when the nested contents are not
in the resource tree, they are not shown in any explorers.

Note that this filtering is designed not to apply for folders starting
with a '.', e.g,. .settings folders.  These folders are hidden by
default in the explorers anyway but need must be present in order to
store project-specific preferences.

### DIFF
--- a/.project
+++ b/.project
@@ -14,4 +14,15 @@
 	<natures>
 		<nature>org.eclipse.m2e.core.maven2Nature</nature>
 	</natures>
+	<filteredResources>
+		<filter>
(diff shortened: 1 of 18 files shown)


wimjongman:
Thanks, Ed. I merged it and I am now running a test install.

PR #703, opened 2021-11-11
Use the generated image for the development environment creation button

merks:
This ensures that all the projects will use a consistent image style for their buttons now and in the future should we want to make central style changes...

It looks like this:

image


wimjongman:
Thanks, Ed. You're on a roll!

Issue #649, opened 2021-05-02
Report viewer failing again.


wimjongman:
The report viewer in the latest build is failing again. See attached stacktrace.
[log.txt](https://github.com/eclipse/birt/files/6411176/log.txt)

hvbtup:
Hmm, it seems like Jasper is using the Jasper JSP compiler (which sounds somewhat silly btw, because we are running BIRT, not JasperReports).
Do we need JSP at all for BIRT?
It seems like this can be disabled or configured, see https://www.eclipse.org/jetty/documentation/jetty-9/index.html#configuring-jsp
But I'm a complete noob in this regard.
Maybe we need some help from the experts of the Jetty project...

wimjongman:
I see in the build that we indeed have a step where jsp's are being compiled. I think that the webviewer is using jsp.

I don't know why but I have a feeling that [this issue](https://bugs.eclipse.org/bugs/show_bug.cgi?id=453379) might be related.

hvbtup:
Yes, that seems related.
Probably you should contact Alexander Kurtakov.

Just to clarify this.
The problem is not Jetty in general, but specifically compiling JSP pages?

Do we need JSP at all?

From what I see in the BIRT sources, there are only 3 *.jsp files:

  *   chartImage.jsp
  *   control.jsp
  *   Userpreference.jsp

All three only contain static content and could be replaced with pure HTML files, as it seems.
From this point of view we could just stop using JSP.

Disclaimer: I am not using the Web Viewer – maybe there are other reasons for JSP…

Henning

Von: Wim Jongman ***@***.***>
Gesendet: Montag, 3. Mai 2021 09:45
An: eclipse/birt ***@***.***>
Cc: Henning von Bargen ***@***.***>; Comment ***@***.***>
Betreff: Re: [eclipse/birt] Report viewer failing again. (#649)


I see in the build that we indeed have a step where jsp's are being compiled. I think that the webviewer is using jsp.

I don't know why but I have a feeling that this issue<https://bugs.eclipse.org/bugs/show_bug.cgi?id=453379> might be related.

SteveSchafer-Innovent:
There are 34 jsp files in the generated webapp.  It would not be possible to replace them with html files.  They are in viewer/org.eclipse.birt.report.viewer/birt/webcontent/birt.

hvbargen:
I see.

At this time this (the viewer not working) is the most annoying bug in BIRT - I'd call it a showstopper.

[...]

PR #733, opened 2021-11-21
Report viewer failing again. #649

claesrosell:
Please review


hvbtup:
This is great!
The preview is working. I tested this from within Exclipse, not with the Maven build.
The SLF4J still not working. Did you commit the changes to make it work as well?

wimjongman:
Thanks, Claes. Awesome work.

The local dev viewer is working but the product does not at this moment. It is failing with a 503.

Looking at the differences between my workspace and the product I see the following missing jetty items (in the product):

* org.eclipse.birt.jetty.overlay_4.9.0.qualifier
* org.eclipse.jetty.annotations_10.0.6
* org.eclipse.jetty.jmx_10.0.6
* org.eclipse.jetty.schemas_4.0.3  <-- ???
* org.eclipse.jetty.servlet-api_4.0.6 <--

There are other differences, but let's add these first.

wimjongman:
I have found the issue with the product. It seems that the extra entries in the web.xml which were removed by Claes, are automatically generated during the build.

Removing the generated data from the web.xml solves the previewer and allows to generate all formats.

The web.xml to change is here: plugins/org.eclipse.birt.report.viewer_4.9.0.v202111221047/birt/WEB-INF

claesrosell:
I will need to get the build environment up and running to solve this. I will try to do that tonight.
When I last tried that, three years ago, it was the precompilation of jsp that made me give up. I guess that it is needed for some of the deployment scenarios of the viewer though.

claesrosell:
I have tested this with the "org.eclipse.birt.designer-win32.win32.x86_64" product and it works for me.

wimjongman:
Claes, you are marvelous. Thank you!!

Issue #734, opened 2021-11-21
org.eclipse.birt.report.designer.internal.ui.editors.script.JSEditor is leaking Fonts


claesrosell:
JSEditor.setDescriptionText(String text) is creating a new Font on every call.

```
private void setDescriptionText(String text) {
	Font font = descriptionText.getFont();
	FontData[] fontData = font.getFontData();
	....
	descriptionText.setFont(new Font(font.getDevice(), fontData));
	descriptionText.setText(description);
}
```
We should probably use BIRTS FontManager or the backing JFaceResources.getFontRegistry() directly instead.

wimjongman:
Yes. good catch. In this case, something like this can also work.


      Font font = new Font(font.getDevice(), fontData)
      descriptionText.setFont(font);
      descriptionText.addListner(SWT.Dispose, e-> font.dispose());

PR #748, opened 2021-11-25
JSEditor is leaking fonts #734

claesrosell:
No longer generating fonts for this label, but using the FontRegistry to use JFace default fonts instead.

### DIFF
--- a/UI/org.eclipse.birt.report.designer.ui.editors.schematic/src/org/eclipse/birt/report/designer/internal/ui/editors/script/JSEditor.java
+++ b/UI/org.eclipse.birt.report.designer.ui.editors.schematic/src/org/eclipse/birt/report/designer/internal/ui/editors/script/JSEditor.java
@@ -1,10 +1,10 @@
 /*************************************************************************************
  * Copyright (c) 2004-2009 Actuate Corporation and others.
- * All rights reserved. This program and the accompanying materials 
+ * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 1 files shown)


UI/org.eclipse.birt.report.designer.ui.editors.schematic/src/org/eclipse/birt/report/designer/internal/ui/editors/script/JSEditor.java
```diff
@@ -1517,34 +1518,29 @@ private void updateMethodDescription(String methodName) {
...
-			for (int i = 0; i < fontData.length; i++) {
-				fontData[i].setStyle(fontData[i].getStyle() & ~SWT.ITALIC);
-			}
+			fontToUse = JFaceResources.getFontRegistry().get(JFaceResources.DEFAULT_FONT);
 			description = text;
 		} else {
-			for (int i = 0; i < fontData.length; i++) {
-				fontData[i].setStyle(fontData[i].getStyle() | SWT.ITALIC);
-			}
+			fontToUse = JFaceResources.getFontRegistry().getItalic(JFaceResources.DEFAULT_FONT);
```
ruspl-afed, line 1532: yes, this will remove the leak, but the Font will not be ITALIC anymore

ruspl-afed (commented):
Why not to keep the existing font styles during the fix? We can use registry to keep our own font for reuse.

ruspl-afed:
> Why not to keep the existing font styles during the fix? We can use registry to keep our own font for reuse.

Please ignore this comment

ruspl-afed (approved):
LGTM

Issue #757, opened 2021-11-29
Generated Report designer 4.9 start is very slow


hvbtup:
I tested the Report Designer All-in-One (which I compiled myself) from the 4.9 master on 2021-11-26.

It works fine once it is started, however I noticed that **starting** the all-in-one-designer **takes 2-3 minutes** on my machine (Core i5, 8GB RAM, SSD) with high cpu on Windows 10 64bit.

This is *before* the progress bar shows.

The 4.2.1 all-in-one-designer only takes a few seconds to start on the same machine.

Colin Sutton reported the same issue on the birt-dev mailing list with the downloaded eclipse4.9-20211127, thus I don't think that something is wrong with my build process.

The second start is also slow.

hvbtup:
I used sampling with JVisualVM during the startup (and stopped when I thought that should be enough).
This is the output:
![grafik](https://user-images.githubusercontent.com/7644776/143831415-5067230b-cf1d-421d-940f-60442a124929.png)

I'm not an expert but to me it seems the "Refresh Thread" is causing most of the CPU load.

Opening it, it looks like this:

![grafik](https://user-images.githubusercontent.com/7644776/143831676-5f8dd4d0-45d2-46b2-8700-adbaedd8fb41.png)

![grafik](https://user-images.githubusercontent.com/7644776/143831928-188cba3c-7e61-4674-984c-728ddfb27354.png)

From here on, I'm lost.

hvbtup:
A quick search for "org.apache.felix.resolver.ResolverImpl checkConsistency slow" finds https://issues.apache.org/jira/browse/FELIX-6140 which was closed only this year and https://issues.apache.org/jira/browse/FELIX-5648

Is someone able to take a look at this?

I'm running out of time for this week and I'm definitely not an OSGI expert.

wimjongman:
Henning, you are getting better by the minute :)

Henning, could it be that you have hit this bug: https://bugs.eclipse.org/bugs/show_bug.cgi?id=548443

My all-in-one takes a long time only the first time I have unpacked it and run it. After the first run, it is fast as greased lightning.

hvonbargen:
Not sure.

Can't do further testing this week.

It sounds like a candidate for being the reason.

On my machine,  we are using GData Antivirus, and unfortunately the second start is as slow as the first one.
BTW I don't see this issue when starting the Oomph generated eclipse for development *of* BIRT.
But maybe this is because those JARs are well-known to the Anti-Virus software.

[...]

PR #761, opened 2021-11-30
Remove org.w3c.dom.events import from org.eclipse.birt.chart.device 

gab1one:
I was able to fix the resolution problems at startup (#757) with this change. Still need to investigate if this breaks anything.

### DIFF
--- a/chart/org.eclipse.birt.chart.device.pdf/META-INF/MANIFEST.MF
+++ b/chart/org.eclipse.birt.chart.device.pdf/META-INF/MANIFEST.MF
@@ -16,4 +16,4 @@ Bundle-ClassPath: .,
 Automatic-Module-Name: org.eclipse.birt.chart.device.pdf
 Import-Package: org.apache.batik.ext.awt.color;version="1.13.0",
  org.apache.batik.transcoder;version="1.14.0",
- org.w3c.dom.events;version="2.0.0"
+ org.w3c.dom.events


wimjongman:
Thanks, Gabriel! We will see if the tests are still running. PTL for unit tests.

gab1one:
They looked good on my machine, but that does not necessarily mean anything in my experience.. :)

wimjongman:
The server agrees.

wimjongman (approved):

hvonbargen:
Did anyone test to generate a PDF report with SVG chart ?

hvbtup:
It's a bit too much magic here. 

Gabriel, can you explain how you came to the conclusion that changing this one line could fix this issue?

Wim, was there a specific reason to add the version constraint '2.0.0' to the manifest file with #640?

wimjongman:
> Wim, was there a specific reason to add the version constraint '2.0.0' to the manifest file with #640?

There was no specific reason. If there was, I can't remember.

gab1one:
@hvbtup the debug output complained wiring conflicts with the org.w3c.dom.events package. This package is offered by the JDK and by the org.w3c.dom.events.3.0.0 bundle. That bundle is required  by the old batik bundle `org.w3c.dom.smil` which is required by the `org.apache.batik.anim` bundle.
I tried a lot of different changes, but in the end figured out, that changing this import statement was the minimal invasive change to prevent these conflicts.

gab1one:
BTW thes chart core tests are failing locally for me, because the fragments that provide the correct runtime wiring are missing during the test run: 
- org.apache.batik.dom.birt.ext
- org.apache.batik.transcoder.birt.ext
We need to add them to the runtime of the test environment

hvbtup:
In particular, generating a PDF report with an SVG chart and a PDF report with an SVG image (and SVG support enabled) should be tested.

Personally, I don't use the BIRT charts.  Can one of you test that?

PR #763, opened 2021-12-01
Enable chart tests in CI

gab1one:
The chart tests did not run with the CI as they used to not work with maven, only when started from eclipse.


wimjongman:
Gabe, you're the boss! Thank you.

PR #765, opened 2021-12-07
Start with Report Design perspective

hvbtup:
This starts the "all-in-one" designer and the RCP designer (which I actually never used) with the Report Design perspective.
The other topics mentionend in #694 (splash screen, icons, welcome page) are still TO DO.


gab1one (approved):

build/birt-packages/birt-report-all-in-one/BIRT.product
```diff
@@ -7,7 +7,10 @@
    </configIni>
 
    <launcherArgs>
-      <vmArgs>-Xms512m
+      <programArgs>-perspective
+org.eclipse.birt.report.designer.ui
```
ruspl-afed, line 11: I cannot find the perspective with identifier `org.eclipse.birt.report.designer.ui`
Please double check that we have the definition with such an identifier.

build/birt-packages/birt-report-rcp/BIRT.product
```diff
@@ -7,13 +7,18 @@
    </configIni>
 
    <launcherArgs>
-      <vmArgs>-Xms512m
+      <programArgs>-perspective
+org.eclipse.birt.report.designer.ui
```
ruspl-afed, line 11: I cannot find the perspective with identifier `org.eclipse.birt.report.designer.ui`
Please double check that we have the definition with such an identifier.

ruspl-afed (commented):
@hvbtup  I have concerns regarding the perspective identifer
if this PR is "work in progress" please mark it accordingly

hvbtup:
Somehow my comment got lost.
To put a long story short: Finding the ID was more or less a guess. It seemed to work for me, but that was only because I used an existing .metadata. The way to find the perspective ID described in #694 did not work. I found another way on https://stackoverflow.com/questions/3383470/how-to-get-an-id-perspective-of-eclipse/, re-tested with a very clean start script and confirmed that ID org.eclipse.birt.report.designer.ui.ReportPerspective is working as expected.

hvbtup:
Not sure if the same ID is correct for the RCP Designer, though. I don't know how the RCP Designer should look, because I never used it before.

wimjongman:
This was fixed with #900

Issue #710, opened 2021-11-11
Replace deprecated Navigator with Project Explorer in Report Design perspective


wimjongman:
The report design perspective uses the Navigator but it is deprecated. Use the Project Explorer instead.

This can be fixed in the ReportPerspective class.

qs-wang:
Interesting after leaving for such a long time, BIRT is still using navigator which I added. Let me try to fix it then. ^_^

claesrosell:
@qs-wang how far is you with this? I saw this comment just now. I have been worked on a PR for this on "after hours" for a couple of days. I was just about to commit it.

qs-wang:
> @qs-wang how far are you with this? I saw this comment just now. I have been worked on a PR for this on "after hours" for a couple of days. I was just about to commit it.

ah, ok. I haven't really started yet. I'll look around, and may try other items later on.

PR #766, opened 2021-12-07
Replace deprecated Navigator with Project Explorer in Report Design

claesrosell:
perspective #710

Replaces the ResourceNavigator with ProjectExplorer in the
ReportPerspective. Also changed all the previous "Action" contributions
to the navigator to Commands / Handlers.


UI/org.eclipse.birt.report.designer.ui.preview.web/src/org/eclipse/birt/report/designer/ui/ide/explorer/AbstractViewHandler.java
```diff
@@ -0,0 +1,43 @@
...
+import org.eclipse.core.resources.IFile;
+import org.eclipse.jface.viewers.IStructuredSelection;
+import org.eclipse.ui.ISources;
+
+/**
+ *
+ * Abstract handler used for all handlers in the explorer
+ *
+ */
+abstract class AbstractViewHandler extends AbstractHandler {
```
ruspl-afed, line 14: Honestly, I do not understand why we need this class
1) we have "visibleWhen" expression that checks for file (it could check more if needed)
2) we can use `HandlerUtil#getCurrentStructuredSelection(ExecutionEvent)` 
So, I would remove this class

UI/org.eclipse.birt.report.designer.ui.preview.web/src/org/eclipse/birt/report/designer/ui/ide/explorer/AbstractViewHandler.java
```diff
@@ -0,0 +1,43 @@
+package org.eclipse.birt.report.designer.ui.ide.explorer;
```
ruspl-afed, line 1: -1 for missed license header

ruspl-afed (changes requested):
please always provide license header for added source files (including manifests)

UI/org.eclipse.birt.report.designer.ui.preview.web/src/org/eclipse/birt/report/designer/ui/ide/explorer/AbstractViewHandler.java
claesrosell: I will remove the class.

hvbtup:
I noticed (*without this PR*) that the RCP Designer does not contain the Navigator/Project Explorer. Is that intended? I'm using grafik
for my ressources. How is the current project folder defined in the RCP Designer? This is hard to see without a navigator or project explorer view.

claesrosell:
@hvbtup : looking at the org.eclipse.birt.report.designer.ui.ReportRCPPerspective git history, it does not look like it has ever had the Navigator. Can you please open a "discussion" for this?

claesrosell:
I do not think that there are any outstanding issues with this path anymore. The new "unnecessary" class was removed so there are no need to add license headers

UI/org.eclipse.birt.report.debug.ui/plugin.xml
```diff
@@ -211,76 +211,31 @@
...
+               label="%Action.DebugReport"
+               icon="icons/full/ctool16/birtdebugger.gif"
+               style="push">
+
+            <visibleWhen>
+               <with variable="selection">
+                  <iterate operator="or" ifEmpty="false">
+                     <adapt type="org.eclipse.core.resources.IResource">
+                        <or>
+                           <test property="org.eclipse.core.resources.extension" value="rpttemplate" />
```
ruspl-afed, line 228: here it is better to rely on contentTypeId as it was before

UI/org.eclipse.birt.report.debug.ui/src/org/eclipse/birt/report/debug/internal/ui/script/handlers/DebugResourceHandler.java
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Add the debug report action in the navigate view
+ */
+
+public class DebugResourceHandler extends AbstractHandler {
+
+	protected IFile selectedFile;
+
+	@Override
+	public void setEnabled(Object evaluationContext) {
```
ruspl-afed, line 34: still seems excessive to me
we have "visibleWhen" expression that ensures it is a file - why should we calculate this here again and how often?

UI/org.eclipse.birt.report.debug.ui/src/org/eclipse/birt/report/debug/internal/ui/script/handlers/DebugResourceHandler.java
```diff
@@ -0,0 +1,67 @@
...
+				}
+			}
+		}
+
+		this.setBaseEnabled(this.selectedFile != null);
+	}
+
+	@Override
+	public Object execute(ExecutionEvent event) throws ExecutionException {
+
```
ruspl-afed, line 54: `IStructuredSelection selection = HandlerUtil.getCurrentStructuredSelection(event)`

ruspl-afed (commented):
Added comments for the first handler, but they are actual for all of them

ruspl-afed:
> The new "unnecessary" class

The point was not to inline its code (with duplications) but to remove its functionality. For me "visibleWhen" expression makes "!isEnabled" just impossible. Do you have a counter example?

claesrosell:
I do not have the code in front of me, but I tried to replicate the old behavior which had both "visible" and "enabled" code, if I remember correctly. I do not know if the old actions could be visible but still disabled.
It make since to have a "visible, hence enabled" behavior. I will implement it with XML. I may have questions though, the XML expression thing is totally new to me.

ruspl-afed:
> I will implement it with XML. I may have questions though, the XML expression thing is totally new to me.

You can try to check do you ever have this handler !isEnabled for the current impl.
I think we don't need more XML, we just need one additional method (in a dedicated type of in an abstract handler type) to extract `IFile` from `IStructuredSelection`. And then we can remove "setEnabled" override for all.

Why I'm thinking so? Because I never needed to override `setEnabled` for handler during my development experience. But I'm open to learn from others :)

claesrosell:
So. I have tried to alter the patch according to your suggestions. I did not understand some of your last comments but hopefully the changes capture the essence of it.

ruspl-afed (approved):
We need to move forward. Not all the code looks idiomatic, but it is definitely has been improved.

Issue #672, opened 2021-10-30
Move projects to Java 11


wimjongman:
Move all projects to Java 11

gab1one:
I recently updated a large Eclipse RCP codebase to java 11 and can give you some input on this:
- The javax.xml bundle is indeed the most common source of that problem.
- Be aware of bundles jars that contain copies of packages / classes from the jre
   - This is very common for some xml packages.
   - You will need to delete these packages

wimjongman:
>  I am a bit afraid that this issue is not "Easy", but I will try to make another stab at it this or next weekend.

You are right this is not that easy.

ruspl-afed:
Yes, it may be not that easy, and I would suggest to provide dedicated PR for each bundle moved.
In that case we can have at least "easy" bundles moved.
Another rationale behind that is that contribution should be less than 1000 LOC to be accepted without additional checks from EF IP team.

wimjongman:
@gab1one, can you help us with this? Setting up a dev environment is described in the [README.md](https://github.com/eclipse/birt/blob/master/README.md)

claesrosell:
Should we go for a "move project by project" strategy on this issue? It makes since to me.

ruspl-afed:
> Should we go for a "move project by project" strategy on this issue? It makes since to me.

yes, definitely, as suggested above

claesrosell:
Progress report:
I now have a compiling workspace where all but one project is migrated to Java-11. This is unfortunately the initial attempt / branch, so I need to chunk it up in smaller PRs. If it actually works, which it probably won´t.

The project that is not moved to Java-11 is the viewer, which has dependencies to Axis 1.4.

wimjongman:
It is not strictly needed to chunk it up. Only if it makes sense. Just create a PR and see if the tests are still ok?

gab1one:
I'll try to set up a dev environment tomorrow. 
@claesrosell if you create a PR with the progress you have so far, I'd be happy to take a look and help you with java 11 problems you are encountering.

claesrosell:
That would be great @gab1one .
There is a PR open.

wimjongman:
Thanks, Gabriel! We put a lot of effort into setting up a dev environment and we hope it is very simple. It is described on the [README](https://github.com/eclipse/birt/blob/master/README.md).

gab1one:
Yes, it is super easy to use 👍🏾 I was able to set it up very quickly!

[...]

PR #774, opened 2021-12-12
Move projects to Java 11 #672

claesrosell:
Moves org.eclipse.birt.viewer to JavaSE-11. Also adds javax.xml.rpc
v1.1.2 to our build since the inclusion of javax.xml.namespace in
javax.xml.rpc v1.1.0 is problematic. However, possibly faulty
Import-Package statements causes javax.xml.rpc to import another version
of javax.xml.rpc since v1.1.0 still exists in our target platform hence
this commit includes a javax.xml.rpc with modified Import-Package
statements.


claesrosell:
I know that adding a modified third party bundle is not ideal but the web viewer is a mess of old legacy stuff. When we get rid of Axis1, this bundle will loose its purpose. The web viewer will, with current technologies, be hard to maintain.

wimjongman (commented):
Looks good to me.

wimjongman:
Claes, can this be submitted?

claesrosell:
Yes

wimjongman:
Thanks, Claes!

hvbtup:
Probably I missed a simple manual config step in Eclipse, but for me this doesn't build in Eclipse.
I get some dozens of errors, but I think it boils down to 

Unsatisfied version constraint: 'javax.xml.rpc: 1.1.2'

for two MANIFEST.MF files in org.apache.axis/META-INF and Unsatisfied version constraint: 'javax.xml.rpc: 1.1.2'.

claesrosell:
Have you imported the new "javax.xml.rpc" project into your workspace?

hvbtup:
Thanks Claes, that fixed 19 of the 69 errors. I could get rid of one of them by Refresh and Clean.
I still got 49 errors then first, most of them saying that org.eclipse.birt.core.internal.util.EclipseUtil could not be imported.
These magically disappeared after 15 minutes!? Anyway, now it's working.

wimjongman:
> Have you imported the new "javax.xml.rpc" project into your workspace?

That has the name javax.xml.rpc2 in the .project file. Was that intended? Should we not put it in our own namespace?

claesrosell:
My intention was not to keep the project name "javax.xml.rpc2". That is probably a leftover from when I recreated the plugin so that it only contained the .jar instead of all .class files. My intention was to rename the project to "javax.xml.rpc".
I was looking at the axis and derby projects for inspiration when I created that project and they are keeping there original namespaces in their project names. Or am I missing something?

wimjongman:
I think it is fine. Potentially it will break an already existing jacax.xml.rpc

Issue #683, opened 2021-11-01
Move license to EPL 2


wimjongman:
The current license is EPL1 and we must move to EPL2.

We need to replace the following files:

license.html
epl-v10.html

and fix all texts in feature.xml and feature.properties.

See examples of the new files here:

https://github.com/eclipse/nebula/tree/master/widgets/oscilloscope/org.eclipse.nebula.widgets.oscilloscope.feature

SteveSchafer-Innovent:
I'll do this.  It's about my speed.

SteveSchafer-Innovent:
What about all the copyright comments in the java files?

ruspl-afed:
1) java files header should be modified to comply with EML2.0
2) about.html should be replaced with its EPL2.0 version
3) other textual files should obtain a license header if not yet

@SteveSchafer-Innovent please select one small bundle to apply EPL2.0 and we will discuss the required changes in details.

wimjongman:
Alexander, is point 3 really needed or should we file a new issue and solve that post 4.9?

ruspl-afed:
Wim, it should not be a big deal while it removes a lot of questions. Typically this is +several files per bundle, nothing notable comparing with amount of .java files.

wimjongman:
Ok. Let's see. 

The java files can be changed using a smart regex find/replace. Same for the about.html.

SteveSchafer-Innovent:
My plan is to change all the files with scripting of one form or another.

wimjongman:
Maybe you can ask on cross project mailing list if someone already made these tooling. @vogella do you know.if there is eplv1-eplv2 tooling?

SteveSchafer-Innovent:
How do I access the cross project mailing list?

ruspl-afed:
https://www.eclipse.org/lists/cross-project-issues-dev/2021/Dec/index.html

vogella:
I used regex expression for the main work and manual fixes then needed,  see here: https://bugs.eclipse.org/bugs/show_bug.cgi?id=535802

wimjongman:
> How do I access the cross project mailing list?

There is this thing called google ;)

SteveSchafer-Innovent:
I've created awk scripts for license.html, about.html, feature.properties, epl-2.0.html files, and java files.  Should I delete the epl-v10.html files?  Still searching for more instances of the license but I should be ready to submit a PR soon.

ruspl-afed:
[...]

PR #777, opened 2021-12-17
Move license to EPL 2 #683

SteveSchafer-Innovent:
Change all occurrences of Eclipse Public License version 1 to version 2
in UI/org.eclipse.birt.report.debug.core.

### DIFF
--- a/UI/org.eclipse.birt.report.debug.core/about.html
+++ b/UI/org.eclipse.birt.report.debug.core/about.html
@@ -11,25 +11,18 @@
 <p>June 28, 2007</p>	
 <h3>License</h3>
 
-	<p>
-		The Eclipse Foundation makes available all content in this plug-in
-		(&quot;Content&quot;). Unless otherwise indicated below, the Content
-		is provided to you under the terms and conditions of the Eclipse
-		Public License Version 2.0 (&quot;EPL&quot;). A copy of the EPL is
-		available at <a href="http://www.eclipse.org/legal/epl-2.0">http://www.eclipse.org/legal/epl-2.0</a>.
-		For purposes of the EPL, &quot;Program&quot; will mean the Content.
-	</p>
+<p>The Eclipse Foundation makes available all content in this plug-in (&quot;Content&quot;).  Unless otherwise 
+indicated below, the Content is provided to you under the terms and conditions of the
+Eclipse Public License Version 1.0 (&quot;EPL&quot;).  A copy of the EPL is available 
+at <a href="http://www.eclipse.org/legal/epl-v10.html">http://www.eclipse.org/legal/epl-v10.html</a>.
+For purposes of the EPL, &quot;Program&quot; will mean the Content.</p>
 
-	<p>
-		If you did not receive this Content directly from the Eclipse
-		Foundation, the Content is being redistributed by another party
-		(&quot;Redistributor&quot;) and different terms and conditions may
-		apply to your use of any object code in the Content. Check the
-		Redistributor's license that was provided with the Content. If no such
-		license exists, contact the Redistributor. Unless otherwise indicated
-		below, the terms and conditions of the EPL still apply to any source
-		code in the Content and such source code may be obtained at <a
-			href="http://www.eclipse.org/">http://www.eclipse.org</a>.
-	</p>
+<p>If you did not receive this Content directly from the Eclipse Foundation, the Content is 
+being redistributed by another party (&quot;Redistributor&quot;) and different terms and conditions may
+apply to your use of any object code in the Content.  Check the Redistributor's license that was 
+provided with the Content.  If no such license exists, contact the Redistributor.  Unless otherwise
+indicated below, the terms and conditions of the EPL still apply to any source code in the Content
+and such source code may be obtained at <a href="http://www.eclipse.org">http://www.eclipse.org</a>.</p>
+
 </body>
 </html>
--- a/UI/org.eclipse.birt.report.debug.core/about.html
+++ b/UI/org.eclipse.birt.report.debug.core/about.html
@@ -11,18 +11,25 @@
 <p>June 28, 2007</p>	
 <h3>License</h3>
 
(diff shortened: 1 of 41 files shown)


UI/org.eclipse.birt.report.debug.core/about.html
```diff
@@ -11,18 +11,25 @@ <h2>About This Content</h2>
...
+	<p>
+		The Eclipse Foundation makes available all content in this plug-in
+		(&quot;Content&quot;). Unless otherwise indicated below, the Content
+		is provided to you under the terms and conditions of the Eclipse
+		Public License Version 2.0 (&quot;EPL&quot;). A copy of the EPL is
+		available at <a href="http://www.eclipse.org/legal/epl-2.0">http://www.eclipse.org/legal/epl-2.0</a>.
+		For purposes of the EPL, &quot;Program&quot; will mean the Content.
+	</p>
 
+	<p>
```
wimjongman, line 23: @waynebeaton is this allowed? Can the content be distributed under different terms and conditions?

UI/org.eclipse.birt.report.debug.core/src/org/eclipse/birt/report/debug/core/i18n/nls.properties
```diff
@@ -1,12 +1,16 @@
...
+# * 
+# * This program and the accompanying materials are made available under the
+# * terms of the Eclipse Public License 2.0 which is available at
+# * https://www.eclipse.org/legal/epl-2.0/.
+# * 
+# * SPDX-License-Identifier: EPL-2.0
+# * 
 # * Contributors:
 # *  Actuate Corporation  - initial API and implementation
+# *  Others: See git history
```
wimjongman, line 13: This should not be added if there is already an existing EPL header. Only add it if the header does not exist. This applies to all entries,

UI/org.eclipse.birt.report.debug.core/src/org/eclipse/birt/report/debug/core/i18n/nls.properties
```diff
@@ -1,12 +1,16 @@
 #/*******************************************************************************
 # * Copyright (c) 2006, 2008 Actuate Corporation.
-# * All rights reserved. This program and the accompanying materials
-# * are made available under the terms of the Eclipse Public License v1.0
-# * which accompanies this distribution, and is available at
-# * http://www.eclipse.org/legal/epl-v10.html
-# *
+# * All rights reserved.
```
wimjongman, line 3: This line is not needed. If we remove it, it will be in line with other projects.

wimjongman (changes requested):
Thank you, Steve. I have a couple of comments.

UI/org.eclipse.birt.report.debug.core/src/org/eclipse/birt/report/debug/core/i18n/nls.properties
SteveSchafer-Innovent: I resolved these two issues a couple of days ago.  Can you review?

wimjongman (approved):

UI/org.eclipse.birt.report.debug.core/plugin.xml
```diff
@@ -1,5 +1,19 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?eclipse version="3.2"?>
+<!--
+ *******************************************************************************
+ * Copyright (c) 2021 Contributors to Eclipse Foundation
```
wimjongman, line 5: Steve, should this not be "the" Eclipse Foundation? I'm not a native speaker but this looks like a spelling error. What do you think?

wimjongman (commented):
Almost there, Steve. It looks great. I have one last comment.

UI/org.eclipse.birt.report.debug.core/plugin.xml
SteveSchafer-Innovent: You're right.  I'll change the scripts.

SteveSchafer-Innovent: As I exercise the scripts on more projects, should I continue with this PR or create a new ones?

wimjongman: I don't understand. Let's focus on this project first. If all is fine then you can make separate pull requests for other projects.

Please also note the build.properties file. It should contain entries for the new license files and entries for removed files should be removed.

SteveSchafer-Innovent: @wimjongman When I say project, I'm referring to eclipse projects within the BIRT repo.  There are 297 of them.  @ruspl-afed suggested we do one PR per bundle, which is roughly equivalent to an eclipse project.  We are done with this bundle, correct?

wimjongman: We are not done yet. There is still this one:


Steve, should this not be "the" Eclipse Foundation? I'm not a native speaker but this looks like a spelling error. What do you think?
Member Author
@SteveSchafer-Innovent SteveSchafer-Innovent 3 days ago

You're right. I'll change the scripts

SteveSchafer-Innovent: I needed to correct them manually but the scripts will pick them up going forward.

SteveSchafer-Innovent: Just noticed another wrinkle.  There are 29 instance of "about.properties" that contain license text in a property value.  The scripts don't detect this because they are looking for comments.  I should fix that before we merge this PR.

SteveSchafer-Innovent: I notice that the file "about.properties" contains a copyright notice in a property value, but my script added a copyright notice as a comment to the file itself because it didn't have one, and these two notices don't match exactly.  I'm not sure if that's a problem since the comment I added is a copyright for the about.properties file and the copyright notice in the property value is a copyright for the eclipse project as a whole.

wimjongman (approved):

Issue #767, opened 2021-12-07
Replace com.lowagie.text


claesrosell:
The version of com.lowagie.text (iText) is old and uses "Illegal reflective access" which causes warnings in Java11 and makes it hard to use in Java16+ 
iText has switched license from MPL/LGPL to AGPL but there are an active fork called OpenPDF which probably can be used without too much effort.
The latest OpenPDF artifact on Maven central has "Export-Package" headers so we should be able use that during build.

BIRT is currently including the com.lowagie.text plugin in some of the features. We would of cause need to change that as well.

patric-r:
+1

hvbtup:
+1

hvbtup:
Last year I tried to replace iText with Apache PDFBox. I can tell you that iText is very deeply integrated with BIRT - and not only with the PDF emitter. In fact, it seems that BIRT uses iText functions to calculate text widths etc.

Thus, OpenPDF is the only possible alternative to iText.

I was concerned about possible legal issues, but I recently I found that open source software from the EU (to be more specific: the "Digital Signature Service") also uses OpenPDF, so I think that possible legal issues are probably not a problem. 

Nevertheless, before deciding to use OpenPDF, a legal expert should take a look at this. Can the Eclipse foundation help here?

hvbtup:
Regarding the illegal reflective access, maybe the code which is causing this is still present in OpenPDF.
AFAIK it is a place where sun.misc.unsafe is used, to optimize memory access for very large PDFs.

claesrosell:
The code is still present but it is called only if the call to sun.misc.Unsafe.invokeCleaner() fails:
https://github.com/LibrePDF/OpenPDF/blob/master/openpdf/src/main/java/com/lowagie/text/pdf/MappedRandomAccessFile.java

Can you describe which legal issues that you are concerned about?
My reasoning was that OpenPDF is a fork from a version of iText, that uses MPL/LGPL, the. It has with the same license as the iText we currently using. Any legal issues with OpenPDF should apply to the version of iText that we are using. But I am probably naive regarding this.

hvbtup:
Ah, yes, MappedRandomAccessFile.java was what I meant.
The code there looks slightly suspicious to me.
I mean, will it have the desired effect of actually freeing memory in Java 17 without sun.misc.Unsafe and without reflection? The exceptions are just ignored.
Couldn't this result in the process eating up all RAM if it processing PDFs for a long time?

[...]

PR #780, opened 2022-01-04
Replace com.lowagie.text #767

claesrosell:
Switched to the latest version of OpenPDF (1.3.26)


hvbtup:
I saw that the CI build failed with

java.lang.NoClassDefFoundError: org/apache/batik/util/SVGConstants
	at org.eclipse.birt.chart.tests.device.render.ImageOutputBaseTest.runTest(ImageOutputBaseTest.java:101)

Is this related to this change?

claesrosell:
Yes. This PR is not done yet. That is way it is still a draft. I have not had time to look into why yet, but I think it has to do with OpenPDF's optional dependencies to XML-graphics and PDFBox and the automatic generation of OSGI Manifest headers.

claesrosell:
I messed this pull request up. I had it working but then I rebased against master and force pushed.
One should never do that.
Sorry, this will probably not be solved today.

claesrosell:
Closing this PR, take a look at #814 instead

Issue #781, opened 2022-01-05
BIRT POJO runtime not working


hvbtup:
I used the Eclipse setup to download and compile the current "master" today.

This generated the file birt-runtime-4.9.0-20220105.zip (in build\birt-packages\birt-runtime\target).
Unzipped this file and added \Path\to\ReportEngine\lib\* to the CLASSPATH environment variable.
Then I tested creating reports from within our application.

This fails with an exception like

org.eclipse.birt.report.engine.api.EngineException: Error happened while running the report.
	at org.eclipse.birt.report.engine.api.impl.EngineTask.handleFatalExceptions(EngineTask.java:1826)
	at org.eclipse.birt.report.engine.api.impl.RunAndRenderTask.doRun(RunAndRenderTask.java:149)
	at org.eclipse.birt.report.engine.api.impl.RunAndRenderTask.run(RunAndRenderTask.java:65)
	at com.tp.server.lisa.birt.report.BirtReportGenerator.run(BirtReportGenerator.java:433)
Caused by: java.lang.ExceptionInInitializerError
	at org.eclipse.birt.report.engine.emitter.pdf.PDFRender.createPageDevice(PDFRender.java:62)
	at org.eclipse.birt.report.engine.layout.emitter.PageDeviceRender.start(PageDeviceRender.java:111)
	at org.eclipse.birt.report.engine.layout.emitter.PageEmitter.start(PageEmitter.java:50)
	at org.eclipse.birt.report.engine.nLayout.LayoutEngine.start(LayoutEngine.java:265)
	at org.eclipse.birt.report.engine.emitter.CompositeContentEmitter.start(CompositeContentEmitter.java:171)
	at org.eclipse.birt.report.engine.api.impl.RunAndRenderTask.doRun(RunAndRenderTask.java:140)
	... 2 more
**Caused by: java.lang.NullPointerException
	at org.eclipse.core.internal.runtime.InternalPlatform.getBundles(InternalPlatform.java:232)
	at org.eclipse.core.internal.runtime.InternalPlatform.getBundle(InternalPlatform.java:222)
	at org.eclipse.core.runtime.Platform.getBundle(Platform.java:1442)
	at org.eclipse.birt.report.engine.internal.util.BundleVersionUtil.getBundleVersion(BundleVersionUtil.java:33)
	at org.eclipse.birt.report.engine.emitter.pdf.PDFPageDevice.<clinit>(PDFPageDevice.java:84)**
	... 8 more

[...]

PR #788, opened 2022-01-24
BIRT POJO runtime not working #781

wimjongman:
Fixed a faulty access to Platform that did not guard
against a possible non-osgi environment.


ruspl-afed (approved):

ruspl-afed:
LGTM

core/org.eclipse.birt.core/src/org/eclipse/birt/core/util/EclipseUtil.java
```diff
@@ -0,0 +1,27 @@
...
+public class EclipseUtil {
+
+	/**
+	 * Safely gets a bundle from Eclipse
+	 *
+	 * @param bundleId
+	 * @return the required bundle or null if the bundle could not be found or the
+	 *         platform is not running.
+	 */
+	public static Bundle getBundle(String bundleId) {
```
ruspl-afed, line 21: Optional<Bundle> to avoid `null`
Then we can have clear understanding what to do on call site. For some cases we already have branches regarding exisitng/absent bundle, but not for all.

core/org.eclipse.birt.core/src/org/eclipse/birt/core/util/EclipseUtil.java
```diff
@@ -0,0 +1,27 @@
+package org.eclipse.birt.core.util;
```
ruspl-afed, line 1: -2 for missed license header

core/org.eclipse.birt.core/src/org/eclipse/birt/core/util/EclipseUtil.java
```diff
@@ -0,0 +1,27 @@
...
+import org.eclipse.core.runtime.Platform;
+import org.osgi.framework.Bundle;
+
+/**
+ * Eclipse and OSGi utilities
+ *
+ * @since 4.9
+ *
+ */
+public class EclipseUtil {
```
ruspl-afed, line 12: `EclipseUtil` is too generic. 

This could be 
```
public final classs FindBundle implements Function<String, Optional<Bundle>> {
...
}
```

ruspl-afed (changes requested):
Could not go without license header

core/org.eclipse.birt.core/src/org/eclipse/birt/core/util/EclipseUtil.java
wimjongman: It is meant to hold more platform access classes in the future. one-class-one-method is overkill. WDYT?

core/org.eclipse.birt.core/src/org/eclipse/birt/core/util/EclipseUtil.java
wimjongman: This means changing the code everywhere instead of creating a simple redirect (which is a refactor). I don't have time for that.

core/org.eclipse.birt.core/src/org/eclipse/birt/core/util/EclipseUtil.java
ruspl-afed: I think we need to use platform facilities directly and improve the platform capabilities rather than invent a (potentially big) utilities class with static method. Of cause this is the question of style and the style of PR is closer to current BIRT source base. Let's go forward.

core/org.eclipse.birt.core/src/org/eclipse/birt/core/util/EclipseUtil.java
ruspl-afed: true, and it should not hold this fix from being applied

ruspl-afed (approved):

wimjongman:
Alexander, should we release this or leave it and use the patch in platform? If we target 22/03 we can change the build to use the IBUILDS repository.

ruspl-afed:
I like the idea of retargeting to 22/03 much more. But I don't want to block anything. If we can move this new class from `public` to `internal` - we are free to either use I build or not.

wimjongman:
That makes sense. Thanks.

Issue #799, opened 2022-01-30
Fix report runner windows batch file


wimjongman:
This batch file will not resolve the classpath correctly and also does not know about the OSGi runtime.

colinsutton:
pdf output produces the correct result but gives a warning:
D:\BIRT\4.9\birt-runtime-4.9.0-20220201\ReportEngine>genreport.bat -p "sample=This is the value of the test parameter" -f pdf .\samples\hello_world.rptdesign
OSGI-Mode, BIRT_HOME=D:\BIRT\4.9\birt-runtime
Java command=java -cp "D:\BIRT\4.9\birt-runtime-4.9.0-20220201\ReportEngine\lib\*" -DBIRT_HOME=D:\BIRT\4.9\birt-runtime org.eclipse.birt.report.engine.api.ReportRunner -p "sample=This is the value of the test parameter" -f pdf .\samples\hello_world.rptdesign
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.lowagie.text.pdf.MappedRandomAccessFile$1 (file:/D:/BIRT/4.9/birt-runtime-4.9.0-20220201/ReportEngine/lib/com.lowagie.text_2.1.7.v201004222200.jar) to method java.nio.DirectByteBuffer.cleaner()
WARNING: Please consider reporting this to the maintainers of com.lowagie.text.pdf.MappedRandomAccessFile$1
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release

colinsutton:
Also fixes:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=473450
https://bugs.eclipse.org/bugs/show_bug.cgi?id=484453
https://bugs.eclipse.org/bugs/show_bug.cgi?id=547526
I will close them when BIRT is released.

PR #800, opened 2022-01-30
Fix report runner windows batch file #799

wimjongman:
* Added help
* Classpath issues were resolved
* Made the script OSGi aware

### DIFF
--- a/build/org.eclipse.birt.releng.birtbuilder/extras/genReport.bat
+++ b/build/org.eclipse.birt.releng.birtbuilder/extras/genReport.bat
@@ -1,4 +1,9 @@
 @ECHO off
+setlocal
+
+PUSHD %~dp0
+set THIS_DIR=%CD%
+POPD
 
 if [%1]==[help] goto usage
 if [%1]==[/?] goto usage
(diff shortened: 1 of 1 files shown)


wimjongman:
@hvbtup can you take a look at this?

hvbtup (changes requested):
Please take a look at https://github.com/hvbtup/birt/tree/Fix_report_runner_windows_batch_file_%23799

Your help output shows the quotation marks.
The right way to escape special characters in CMD ist to escape them with '^'.

Other things I changed:
* The script can never run into the "BIRT_HOME not set" branch, so I removed it.
* Used the IF DEFINED syntax.
* Set the BIRT_HOME environment variable and the specifiy the corresponding system property if and only if running in the OSGI runtime.
* Used the little THIS_DIR trick to get rid of the trailing/double backslash. This little change means that the script can not be called with an UNC path (it needs a "drive letter"), but the script is more or less for local testing anyway, so I don't consider this to be an issue.

wimjongman:
Yes, this makes sense since it is in my repo. 

Let me know if you are good to commit this.

PR #814, opened 2022-02-06
Replace iText with OpenPDF #767

claesrosell:
This is an attempt to do a cleaner solution to #767 and, if it compiles, it will replace #780


ruspl-afed:
@claesrosell we may need to go through CQ process for this dependency if we want to bundle it.

What I can see [here](https://github.com/LibrePDF/OpenPDF#license) requires LGPL2.1+MPL2.0

@waynebeaton please advice

claesrosell:
Ok. If I remember correctly, the version of iText we budled and this version of OpenPDF has the same licenses. I hope that it works out.

wimjongman (approved):
Looks good to me. Thanks, Claes!

PR #819, opened 2022-02-08
Export bookmarks #797

hvbtup:
### PDF: Export information about the generated bookmarks to the appContext

With this PR the PDF emitter will create an entry "Bookmarks" in the appContext, if there are any bookmarks.
The content of this entry is a Map<String, BookmarkInfo>
The BookmarkInfo object contains the following information about the bookmark target: name (also used as the key), page number, x & y of the top-left corner and width & height.

This information can be used by PDF post-processors e.g. to "stamp" content at a location where the location itself is dynamic.

One example would be to add a visible digital signature whose location depends on the report output:
Create an empty 1x1-grid of fixed size where the signature should be places. Set the bookmark property of this grid to "signature".
After BIRT created the PDF, examine the appContext with

     BookmarkInfo signatureInfo = appContext.get("Bookmarks").get("signature")

`signatureInfo.pageNumber, .x, .y, .width, .height` then give you the information where to place the visible signature exactly.

Note: The BookmarkInfo also has a useful toString method, so you could also write the bookmark info to a text file in the report's afterRender event and parse that textfile, if the post-processing program is not written in Java.

Other use cases:
- Add PDF content (from other PDFs) as kind-of images, for example to include diagrams with mermaid or plantuml.
- Add a text like "This report contains 5 pages" (where the number is dynamic, of course) inside the body
- For automatic testing of the PDF emitter (in particular, useful for regression tests)


engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/BookmarkInfo.java
```diff
@@ -0,0 +1,85 @@
...
+ * terms of the Eclipse Public License 2.0 which is available at
+ * https://www.eclipse.org/legal/epl-2.0/.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   See git history
+ *******************************************************************************/
+
+package org.eclipse.birt.report.engine.emitter.pdf;
```
wimjongman, line 14: Is this API? If not please move this to an internal class.

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/BookmarkInfo.java
```diff
@@ -0,0 +1,85 @@
...
+ * Copyright (c) 2021 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * https://www.eclipse.org/legal/epl-2.0/.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   See git history
```
wimjongman, line 11: Henning von.. Initial API and implementation

wimjongman (changes requested):
Thanks, Henning. Please take a look at my requested changes.

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/BookmarkInfo.java
hvbtup: It is API, because is meant to be available to BIRT users from Java and JS in the appContext.

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/BookmarkInfo.java
```diff
@@ -0,0 +1,85 @@
+/*******************************************************************************
+ * Copyright (c) 2021 Contributors to the Eclipse Foundation
```
wimjongman, line 2: Henning, you need to add you or your company here.

wimjongman (changes requested):

wimjongman (approved):

wimjongman:
You are good to Squash and merge Henning. Then you can throw away your branch and sync your fork.

Issue #706, opened 2021-11-11
Fix Warnings


wimjongman:
We have 130K warnings which are not good for the performance and image of the project.

Either:

* Suppress the warnings in the project preferences
* Fix them

SteveSchafer-Innovent:
I'll work on these as I have time.  If someone else wants to help we should agree on some kind of division of areas so we don't collide.

hvbtup:
Regarding NLS-warnings: I had started with adding NON-NLS comments on individual lines in the Word and Excel emitters, but that is rather tedious work, wouldn't recommend that. I think it's better to switch those warning off for whole classes. Most of the Java classes in the engine do not contain translatable strings.

SteveSchafer-Innovent:
Eclipse will automatically add the NON-NLS comments everywhere in quick-fix.  That removes the tedious aspect.  Would this then be a better solution?

wimjongman:
No because it will also non-nls the text that has to be translated.

Switching off the error in the preferences would be the only quickfix

hvbtup:
I still think that after skim reading a *.java source file (scanning for translatable messages) in most cases SuppressWarnings can be added to whole classes.

As a non-english developer and user I think that only texts *inside the IDE* should be translated, whereas all those text that only occur *at runtime* (debug, warning, info, error messages) should strictly *not* be translated.

The reasoning behind this is: If you are not a developer, then those texts are all gibberish for your anyway. 

I know this from ~30 years.  experience. Only in a handful of cases has any of our users understood the root cause of a stack trace.

Those messages are understandable only for developers. And even then, you often have to search the internet for more information. Since we don't have unique message ids (such like ORA-01403 for Oracle databases with the meaning "No data found" or "Keine Daten gefunden" in German), it would be counter-productive to have a e.g. a German translation. Developers are expected to understand a basic subset of English anyway, and when searching the internet, the chances to find some useful information are much higher with an english search phrase.

wimjongman:
I do not agree with the first paragraph. The rest I do agree with.

SteveSchafer-Innovent:
@wimjongman if you don't agree with suppressing warnings on whole classes then you wouldn't agree to changing the severity level to "ignore" across the board.  Are we left with examining individual string constants?

wimjongman:
If there is no plan to externalize translatable strings then I'm fine with changing the severity level. If the plan does come up in the future it is easy to bump the severity level back to a warning.

[...]

PR #834, opened 2022-02-16
Fix Warnings #706

SteveSchafer-Innovent:
Change severity level of compiler compliance and JRE compatibility to "ignore".
Change severity level of unused thrown, import, local, parameter, and private member to "ignore".


hvbtup:
I'm -1 on this. 
Most of these warnings indicate possible programming errors with a non-neglectable probability of being actual errors.
Instead I think we could get rid of quite a lot of warnings by disabling all those warnings about raw types (e.g. untyped collections like List without <...>).

claesrosell:
I am -1 on these severity changes as well. Most of these warnings should be resolved by changing the code. I am -1 on disabling warnings for untyped collections as well. It will be a lot of work but the project will be a lot more maintainable when fixed.

Warnings are our friends! They tell us what we should do to improve the code and the product.
I know that there are annoyingly many of them in this project but we shouldn't try to hide that. We should resolve them by fixing them. It will take a lot of time but until they are solved let them remain to remind us of what to do.

wimjongman:
Steve, you can also take a project and apply the "Clean source" action to it. This can upgrade a lot of old code and will remove warnings.

e.g. fix the raw types, use modern for loops, use Lambda's, remove unused fields from the code, etc... 

Maybe you can play with that.

SteveSchafer-Innovent:
Okay I'll see what I can do.

PR #836, opened 2022-02-22
Fix Warnings #706

SteveSchafer-Innovent:
Execute the following clean-up actions:

Convert control statement bodies to block
Combine nested 'if' statement in 'else' block to 'else if'
Simplify lambda expression and method reference syntax
Pull up assignment
Use instanceof keyword instead of Class.isInstance()
Exit loop earlier
Replace String concatenation by StringBuilder
Convert StringBuffer to StringBuilder for local variables
Use String.replace() instead of String.replaceAll() when possible
String.isBlank() rather than String.strip().isEmpty()
Use lazy logical operator (&& and ||)
Primitive comparison
Primitive parsing
Primitive serialization
Primitive type rather then wrapper class
Precompile reused regular expressions
Remove unnecessary String creation
Use boolean literals instead of Boolean.TRUE/FALSE when used as a primitive
Remove unused imports
Remove unused private methods
Remove unused private constructors
Remove unused private types
Remove unused private fields
Add missing '@Override' annotations
Add missing '@Override' annotations to implementations of interface methods
Add missing '@Deprecated' annotations
Remove redundant modifiers
Remove redundant semicolons
Create array with curly
Remove useless return
Remove useless continue
Convert loop into if
Remove unnecessary '$NON-NLS$' tags
Organize imports
Format source code
Remove trailing white spaces on all lines
Remove redundant String.substring() parameter
Use Arrays.fill() when possible
Evaluate without null check
Boolean value rather than comparison
Avoid double negation
Remove overridden assignment
Remove redundant comparison statement
Remove unreachable block
Operate on Maps directly
Initialize collection at creation
Initialize map at creation
Factorize operands
Single 'if' statement rather than duplicate blocks that fall through
Remove redundant if condition
Use Comparator.comparing()
Use try-with-resource
Multi-catch
Use Java method instead of system property 'FILE_ENCODING'
Use Java method instead of system property 'PATH_SEPARATOR'
Use Java method instead of system property 'FILE_SEPARATOR'
Use Java method instead of system property 'LINE_SEPARATOR'
Use Java method instead of system property 'BOOLEAN_PROPERTY'
Use diamond operator
Use Objects.hash()
Use Objects.equals() in the equals method implementation
Use Autoboxing


claesrosell:
I think that these build problems are related to the lines:
javacSource=1.5
javacTarget=1.5
... in the build.properties file of org.eclipse.birt.report.engine.emitter.config .

There are other projects with these, or similar, lines as well. I think that it is safe to remove these lines. @wimjongman what do you think?

claesrosell:
I have created PR #837, which probably solve these build errors.

claesrosell:
@SteveSchafer-Innovent, if you "Fetch upstream" your [SteveSchafer-Innovent:issue-706-4](https://github.com/SteveSchafer-Innovent/birt/tree/issue-706-4) branch, I think that org.eclipse.birt.report.engine.emitter.config will compile now.

viewer/org.eclipse.birt.report.viewer/birt/WEB-INF/web.xml
```diff
@@ -255,4 +255,353 @@
 	</jsp-config>
 <context-param><param-name>org.eclipse.jetty.jsp.precompiled</param-name><param-value>true</param-value></context-param>
 
+<!--
```
claesrosell, line 258: These additions to web.xml is done by the JSP pre-compiler during build and should not commited since they cause problems with the web viewer when BIRT is executed from Eclipse IDE.

claesrosell (changes requested):

claesrosell (approved):

claesrosell:
Yes, of cause. I thought that I just approved the web.xml changes. Did I approve the complete pull request?

SteveSchafer-Innovent:
Some of the failures are of this form:  
junit.framework.ComparisonFailure: expected:<null> but was:<null>
Hmmm.  Maybe one is a real null and the other is "<null>"?
Looking too see if any of the clean-ups would have affected these tests.

claesrosell:
I posted this on the issue but is copy / pasting it here as well : 

On the theme "less then perfect clean-ups", this is the reason why #836 currently fails :
The "Replace String concatenation by StringBuilder" clean up converts

```
	public String getDisplayLabel(Module module, int level) {
		String displayLabel = super.getDisplayLabel(module, level);
		if (level == IDesignElementModel.FULL_LABEL) {
			String name = getStringProperty(module, IReportItemModel.DATA_SET_PROP);
			name = limitStringLength(name);
			if (!StringUtil.isBlank(name)) {
				displayLabel += "(" + name + ")"; //$NON-NLS-1$//$NON-NLS-2$
			}
		}
		return displayLabel;
	}
```

to

```
	@Override
	public String getDisplayLabel(Module module, int level) {
		StringBuilder displayLabel = new StringBuilder().append(super.getDisplayLabel(module, level));
		if (level == IDesignElementModel.FULL_LABEL) {
			String name = getStringProperty(module, IReportItemModel.DATA_SET_PROP);
			name = limitStringLength(name);
			if (!StringUtil.isBlank(name)) {
				displayLabel.append("(").append(name).append(")"); //$NON-NLS-1$//$NON-NLS-2$
			}
		}
		return displayLabel.toString();
	}
```

The old implementation will return null if super.getDisplayLabel() returns null and level != IDesignElementModel.FULL_LABEL , and the new implementation will return "null" under the same circumstances.

This is of cause suspicious code to begin with since the original code will throw a NPE if super.getDisplayLabel() returns null and level == IDesignElementModel.FULL_LABEL. but it is not a good sign that the clean up changes the behavior of the method.

I was initially very optimistic about these "automatic clean-ups" but I am starting to get scared instead.

wimjongman:
Apart from the fact that this conversion should work as advertised, we should not do the StringBuilder conversion because that is being taken care of by the JVM. StringBuilder makes badly readable code which we can avoid.

SteveSchafer-Innovent:
Thanks for finding that Claes.  I just re-did the clean-ups without the StringBuilder one.  Fingers crossed.

wimjongman:
Great, Steve!! Please Squash and Merge when you are ready.

PR #837, opened 2022-02-22
Removed javacSource and javacTarget from build.properties files.

claesrosell:
This PR aims to remove the JavaSE-1.5 source level restriction still available on some projects which prevents PR #836 to build


wimjongman (approved):
Yes, this looks good.

PR #842, opened 2022-02-28
Fix compiler compliance issues.

SteveSchafer-Innovent:


wimjongman:
Steve, let's move this to 4.10.

SteveSchafer-Innovent:
Do I need to do anything with this to move to 4.10?  I notice the maven version in master is still 4.9.0.

wimjongman:
With "move to 4.10" I meant to not include it in 4.9

Issue #838, opened 2022-02-25
Some reports created with BIRT 4.2.1 don't work with 4.9.0


hvbtup:
Some of our exisiting reports are using the CONCATENATE aggregate function.

When running such a report with BIRT 4.9, the defined properties such as separator, maximum length etc are ignored and the default values are uses (an empty separator and a maximum length of 1000).

This results in incorrect output.

I needed a lot of time to find out what's causing the issue.

In the original report, the properties are defined as follows:

                <structure>
                    <property name="name">all_products</property>
                    <property name="dataType">string</property>
                    <property name="aggregateFunction">CONCATENATE</property>
                    <list-property name="arguments">
                        <structure>
                            <property name="name">Expression</property>
                            <expression name="value">row["PRODUCTNAME"]</expression>
                        </structure>
                        <structure>
                            <property name="name">Separat&amp;or</property>
                            <expression name="value">", "</expression>
                        </structure>
                        <structure>
                            <property name="name">Ma&amp;x length</property>
                            <expression name="value">1000000</expression>
                        </structure>
                        <structure>
                            <property name="name">Sho&amp;w all values</property>
                            <expression name="value">true</expression>
                        </structure>
                    </list-property>
                </structure>

If I open the report in the *new* designer, this is silently converted to

                <structure>
                    <property name="name">all_products</property>
                    <property name="dataType">string</property>
                    <property name="aggregateFunction">CONCATENATE</property>
                    <list-property name="arguments">
                        <structure>
                            <property name="name">Expression</property>
                            <expression name="value">row["PRODUCTNAME"]</expression>
                        </structure>
                        <structure>
                            <property name="name">Separator</property>
                            <expression name="value">", "</expression>
                        </structure>
                        <structure>
                            <property name="name">Maxlength</property>
                            <expression name="value">1000000</expression>
                        </structure>
                        <structure>
                            <property name="name">Showallvalues</property>
                            <expression name="value">true</expression>
                        </structure>

Note that ampersands (to mark the keyboard shortcut key) are present in the old file, but no longer in the new file.

At runtime, the implementation is looking for the "Separator" property but can't find it.

This can only be solved by manually "migrating" the existing rptdesign files to the new format.

Of course, the new format is better, but the necessity of manual migrating existing reports is a PITA.

Question:
Is there a migration guide - or a migration tool?
Does anybody know if there are other such issues?

claesrosell:
I will dig in the git commit log and see if I find the reason for this.

[...]

PR #845, opened 2022-03-04
Some reports created with birt 4.2.1 don't work with 4.9.0 #838

claesrosell:
Added migration of old OdaDataSet that used translated strings
(Message.getString()), to the new property names introduced with commit
cac6e65a766ecbf1086728cebdbb5d54b4755728

I know that we are close to release, and introduce model migration for old issues may be a bad idea. - But here it is.

### DIFF
--- a/model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/parser/OdaDataSetState.java
+++ b/model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/parser/OdaDataSetState.java
@@ -201,7 +201,7 @@ public class OdaDataSetState extends SimpleDataSetState {
 		mergeResultSetAndResultSetHints(tmpElement);
 		doCompatibleRemoveResultSetProperty(tmpElement);
 		doCompatibleRemoveResultSetHitProperty(tmpElement);
-		doCompatibleConvertComputedColumnsConcatenateProperites(tmpElement);
+		doCompatibleConvertComputedColumnsConcatenateProperties(tmpElement);
 
 		TemplateParameterDefinition refTemplateParam = tmpElement.getTemplateParameterElement(handler.getModule());
 		if (refTemplateParam == null) {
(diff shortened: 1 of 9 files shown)


model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/parser/OdaDataSetState.java
```diff
@@ -198,6 +201,7 @@ public void end() throws SAXException {
 		mergeResultSetAndResultSetHints(tmpElement);
 		doCompatibleRemoveResultSetProperty(tmpElement);
 		doCompatibleRemoveResultSetHitProperty(tmpElement);
+		doCompatibleConvertComputedColumnsConcatenateProperites(tmpElement);
```
wimjongman, line 204: Typo in the method name ...Properties

wimjongman (changes requested):
Thank you, Claes. That looks great. We should maybe create a test case for that. I only found one minor issue.

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/parser/OdaDataSetState.java
claesrosell: Fixed with latest commit.

claesrosell:
I am not good with unit tests, but I will try to add a test that uses the compareDesignModel() test util to determine if a migration is properly read/parsed.

wimjongman:
Ok, that would be great. Maybe @hvbargen can help you or design a proper test. It should not be hard to copy an existing test. Let me know if you need help.

wimjongman (approved):

hvbtup:
Sorry, I'm totally busy ATM, won't find the time to test this in the next few days. But skimming over the code changes, it looks good to me.

claesrosell:
Test added

wimjongman (approved):
Wonderful.

wimjongman:
Thanks, Claes. Please Squash and Merge.

hvbargen:
Thanks Claes. I'm looking forward to merge this into our product and test 
it with our patched BIRT and a complex report before April. This is very 
helpful.

Issue #852, opened 2022-03-08
Fix Launch configurations


wimjongman:
Several launch configurations have become stale.

hvbtup:
+1 I did not test it, but your changes look quite similar to those I had in my repo master branch to make the all-in-one designer (what confused me a bit in PR #851).

SteveSchafer-Innovent:
@wimjongman I'm getting this error again with 4.10 master:

!ENTRY org.eclipse.equinox.launcher 4 0 2022-06-02 12:01:35.047
!MESSAGE Exception launching the Eclipse Platform:
!STACK
java.lang.ClassNotFoundException: org.eclipse.core.runtime.adaptor.EclipseStarter
	at java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:476)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:589)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
	at org.eclipse.equinox.launcher.Main.invokeFramework(Main.java:656)
	at org.eclipse.equinox.launcher.Main.basicRun(Main.java:596)
	at org.eclipse.equinox.launcher.Main.run(Main.java:1467)
	at org.eclipse.equinox.launcher.Main.main(Main.java:1440)

I'd appreciate it if you could describe the steps you take to update a stale run configuration.

wimjongman:
Please open a new issue. This one is closed in 4.9

PR #853, opened 2022-03-08
Fix Launch configurations #852

wimjongman:
Stale launch configurations were fixed.


wimjongman:
Steve, this fixes the all-in-one run problems that you were facing.

Issue #856, opened 2022-03-09
Aggregation button disabled


jochenberger:
Hi,

we tried to upgrade to 4.9. The results look very promising (:clinking_glasses:), but it looks as if there is still an issue with the aggregation button. See https://bugs.eclipse.org/bugs/show_bug.cgi?id=511212 for reference.

Cheers,
Jochen

_Originally posted by @jochenberger in https://github.com/eclipse/birt/discussions/855_

claesrosell:
I have made a fix for this, that is building now. The build will be available here: https://github.com/eclipse/birt/actions/runs/1959493255
Please try it out.

jochenberger:
@claesrosell, the fix works for us. Thanks for the quick action.

speckyspooky:
@claesrosell 
I tested the original-report with rcp-version-4.9 and confirm too. The fix works and the function editor is now usable.
Thanks to your side!

![rcp-designer-4 9](https://user-images.githubusercontent.com/41593722/157748968-e470cb53-a5fe-484a-85f0-31f9303dac56.png)

wimjongman:
Thanks, Claes. Another quality patch.

PR #857, opened 2022-03-09
Aggregation is disabled in charting #856

claesrosell:
Enabled aggregation again. Seems to have been a negation wrong after refactoring. Also expanded
the default IChartUIHelper.useDataSetRow() implementation to a simple startsWith("data") check previously used in the enableAddEditor() method


claesrosell:
The change in ChartUIHelperBase.useDataSetRow(Object reportItem, String expression) is taken from the old test used in the BaseDataDefinitionComponent.enableAggEditor() prior to the refactoring done in commit b40bee012aa2e29b00c138d1adb91ff998a7043b , however the useDataSetRow() method is now used in other places as well and may effect those. According to the javadoc the method should:
"Return true if expression uses binding that contains data set row directly or indirectly"
but did only return "false" in the only implementation.
I think that this is a minimal start for a useDataSetRow() implementation, we can expand on it in the future when our knowledge of the code base has increased.

claesrosell:
[qs-wang](https://github.com/qs-wang) Can you provide any input on this? I got the feeling that you have worked on Birt previously.

wimjongman (approved):

wimjongman:
Claes, we have to close master until we have released 4.9. I am fine taking this in but you have to do it now. 

Otherwise, we retarget this for 4.10

Issue #817, opened 2022-02-08
Sign our builds


wimjongman:
Before we can deliver our final release, we need to sign the binaries. For mac we need to notarize and created DMG's. 

The signing process is described here:

https://wiki.eclipse.org/IT_Infrastructure_Doc#Sign_my_Jar.2Fplugins.2FWindows_exe.2FmacOS_App_files.3F

wimjongman:
Part 1 is done. The update site and executables are signed. Now I have to take a look at signing/notarizing the mac applications. It is not quite clear from the description ^^ how to do that.

wimjongman:
There are still some signing issues when installing BIRT from our snapshot update site [1]

![image](https://user-images.githubusercontent.com/180969/155364164-c4b61b50-79d8-4dc1-a770-d510ce08b596.png)

[1] https://download.eclipse.org/birt/update-site/snapshot/

wimjongman:
@claesrosell any idea why the plugins above are unsigned? I would not expect that the jetty bundles are unsigned, nor any other content we fetch from maven.

claesrosell:
I think they are unsigned because the eclipse-jarsigner-plugin only signs bundles that are built and not dependencies. The above bundles are included in our features but pulled from the bundle pool ( our target platform ).
To solve this we would need to either ask the maintainers of the projects to sign the bundles (hard) or sign them our self via some Maven magic. 
I found this, very old, Stackoverflow post: https://stackoverflow.com/questions/22541301/jarsigner-doesnt-sign-plugin-dependencies. Which mentions a workaround, if the problem indeed is what I think.

I am curious regarding the .source bundles in the screenshot. That must be a problem in some of our features / plugins.

wimjongman:
Closing for 4.9 follow-up issue is  _Sign our builds #871_

PR #868, opened 2022-03-16
 Sign our builds #817 

wimjongman:
Move to Jetty 10.0.8 in the hope that these are signed.


wimjongman:
Abandoned because platform is using 10.0.6.

Issue #869, opened 2022-03-16
Master is not building due to broken target platform


claesrosell:
Maven / Tycho says:
No repository found at https://download.eclipse.org/eclipse/updates/4.23-I-builds.
Changing to https://download.eclipse.org/eclipse/updates/4.24-I-builds.

wimjongman:
Claes, maybe we should switch to 4.23 proper instead [1] of 4.24i? 

The reason we used i-builds was because of jetty IIRC. 

[1] https://download.eclipse.org/eclipse/updates/4.23/

claesrosell:
Yes it might be better.
I actually did that first, but then I remembered the "master is open for 4.10 development" statement and I thought that 4.24-i was more appropriate.
How do we plan to go forth? I foresee 4.9.x releases ( support for SVG is broken in 4.9.0, #865. More bugs will certainly show up ). Should we branch of for 4.9.x ? If so, that branch should use https://download.eclipse.org/eclipse/updates/4.23/ and master could perhaps continue with 4.24-i.
If 4.9.x releases will be done from master, we should for sure change to 4.23 there.

wimjongman:
I think we should just continue counting to 4.10. We can release it next week if we want, we don't have to go through release reviews anymore if released within one year of the previous release.

claesrosell:
We will need to synchronize with the release train at some point of development. Eclipse is released every third month and we would need to update our target platform to adopt to the changes of our dependencies well before that. If not, the Birt designer may be uninstallable in the IDE.

wimjongman:
It depends on how strict our version numbers are. For example, we now build against jetty 10.0.6 but it will run fine when the platform updates to jetty 10.0.8. If we don't follow the minor updates of the platform then BIRT 4.10/11/12/13 will also be installable in Eclipse 4.23

Question is, do we want to follow Eclipse or do we keep targeting Eclipse 4.23 for the next x period.

Targeting Eclipse 4.23 gives adopters some breathing room in building a product that they don't have to update every three months if they need to incorporate new BIRT fixes.

There are many Eclipse-based products that use BIRT (e.g IBM in the rational suite, SAP in their development environment, and probably many others) that can't keep up with this pace.

Maybe we can start a dis

claesrosell:
Yes. 
Since my company sells and develops an RCP based product which uses Birt. I would very much like to keep Birt "up to date" with the rest of Eclipse.
[...]

PR #870, opened 2022-03-16
Master is not building due to broken target platform #869

claesrosell:
Changing https://download.eclipse.org/eclipse/updates/4.24-I-builds/ for
now.


ruspl-afed:
Thanks @claesrosell !

Issue #875, opened 2022-03-17
Report with Oracle JDBC ORA-01000 Too many open cursors


hvbtup:
We are facing a serious issue with BIRT 4.9 when used with an Oracle DB.

One of our reports has a structure like this:
![grafik](https://user-images.githubusercontent.com/7644776/158799538-b8648320-5fd3-4b65-8f62-306767af6faa.png)
That is, we have a simple master-detail structure (with multiple details).

The table "Reports" is bound to a JDBC dataset with the same name.
For each report, there are some detail tables, one of these is table "Ausgabegeräte", which is bound to a JDBC dataset with the name.
Some systems call this a "sub-report".

The dataset "Ausgabegeräte" uses a dataset parameter to select the detail data for the given report from the layout context (DataSet Parameter Binding)
![grafik](https://user-images.githubusercontent.com/7644776/158800061-2ec3619b-9931-4623-bb1e-190b3eb8a5e7.png)

The datasets contain beforeOpen events just for debugging and performance tuning purposes (showing that the dataSet is about to opened for a given REPT_ID).

This is all very straightforward und it worked for > 10 years.

But now, with BIRT 4.9, the report crashes with "ORA-01000 too many open cursors" error message in the exception trace
if I restrict
By using Oracle's tracing tools, I could verify that indeed the cursors for the SQL statements in the reports' datasets are closed only at the end of the report instead of when the processing is done.

I added logging statements to the `afterOpen` and `afterClose` events of the dataset "Ausgabegeräte" and this explains the error:

With BIRT 4.3, the sequence of the events was

     afterOpen afterClose
     afterOpen afterClose
     afterOpen afterClose
     ...

but with BIRT 4.9 the sequence is 

    afterOpen afterOpen afterOpen ...
    afterClose afterClose afterClose

**So, the JDBC cursors are freed only at the end of the whole report. It seems like something has changed significantly in the processing.**

Maybe something has changed regarding caching (or options have been renamed or have different default values?)

Note: The report uses `RunAndRenderTask`.

claesrosell:
@hvbtup. Can you try tro reproduce this behavior with a small example report and the sample data base? That way I could look into without much effort. Time is really limited right now.

hvbtup:
I created a sample reporting demonstrating the different behavior.
[issue-875.zip](https://github.com/eclipse/birt/files/8286156/issue-875.zip)

Both log files were created by running the report directly in the all-in-one designer and then renaming the file.

[...]

PR #876, opened 2022-03-17
Report with Oracle JDBC ORA-01000 Too many open cursors #875

claesrosell:
IServiceForQueryResults.close() is now closed upon QueryResults.close(). This is currently the only place in the frame work were IServiceForQueryResults.close() is called.

@hvbtup : Since you have problems with your dev.environment I took the liberty to make a PR for you to test.


hvbtup:
Works for me, but what about the problem that the to-be-reverted commit wanted to solve? See my comment in #875

hvbtup (commented):
Works for me, but probably we have to be more flexible? See my comment in #875

wimjongman (approved):
Let's go!

Issue #894, opened 2022-03-23
Disable pull_request_target


wimjongman:
pull_request_target and pull_request are running simultaneously. According to [1] pull_request_target is dangerous. 

[1] https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

PR #895, opened 2022-03-23
Disable pull_request_target #894

wimjongman:
Removed pr_target


wimjongman:
@gab1one WDYT? See also #894

gab1one:
Totally makes sense

Issue #882, opened 2022-03-22
Include Java in products


wimjongman:
It is now possible to embed Java in our products. It makes it easy for users to not have to fool around with Java installations but run immediately.

wimjongman:
The RCP and All-in-one products now include a jre. A separate install is no longer needed.

PR #900, opened 2022-03-27
BIRT executable story #694 + Include Java in products

wimjongman:
https://github.com/eclipse/birt/issues/694
https://github.com/eclipse/birt/issues/882

Added a java jre to the products
Added icons to the producst


hvbtup (commented):
I like the images.

Issue #885, opened 2022-03-22
NullPointerException using COUNT aggregate


hvbtup:
With some reports we are facing a NullPointerException.

The relevant part of the stack trace is this:
...
Caused by: java.lang.NullPointerException
	at org.eclipse.birt.data.engine.executor.aggregation.ProgressiveAggregationHelper$JSColumnBindingRow.get(ProgressiveAggregationHelper.java:508)
...

I can narrow it down to the following binding:
![grafik](https://user-images.githubusercontent.com/7644776/159457398-e3417287-f3ba-4ee2-9e36-d65f82c0f7c8.png)

I don't know if that's relevant, but actually the "AnzahlKarten" binding is not at all used inside the list item.
So I could fix my report (in this particular case) by just removing the binding "AnzahlKarten".

It seems that the problem is related to the fact that the expression to count is not a column from the dataset row.
The issue arises when the expression is a constant expression or row.__rownum.

A general workaround is to count any not-null column instead of row.__rownum or a constant expression.

The very same report with the very same data works with BIRT 4.3.0.

As a side note: The COUNT aggregate is somewhat weird anyway. 
I would expect that it should be possible to use COUNT and COUNTDISTINCT also without an expression.
COUNT should then be the number of rows and COUNTDISTINCT should be the number of distinct rows (considering all columns from the database). That would be in line with the corresponding SQL keywords COUNT and COUNT (DISTINCT ...).

claesrosell:
@hvbtup , Can you provide a small test report with the example database which recreates this problem?

hvbtup:
Voila!
[issue-885.zip](https://github.com/eclipse/birt/files/8331351/issue-885.zip)

hvbtup:
I created a modified version of the report.

[issue-885.zip](https://github.com/eclipse/birt/files/8409747/issue-885.zip)
Here, I created to tables with a filter condition based on PRODUCTLINE.
The first table filters PRODUCTLINE LIKE "T%", the second one PRODUCTLINE LIKE "X%".
Otherwise the tables are identical:
The aggregate "CountRows" uses COUNT to count row.__rownum.
The visibility expression hides the table if CountRows == 0.
The two tables are identical except for the filter expression.

With BIRT 4.2.1, the output shows only the "T" table, with a correct value of 2 for CountRows (Trucks and Trains&Busses).
With BIRT 4.9.0, the report fails,
With the Designer from #914, the output shows _both_ tables, with a value of 1 for CountRows for the X table.

[...]

PR #914, opened 2022-04-02
Null pointer exception using count aggregate #885

claesrosell:
Added special case for ScriptConstants.ROW_NUM_KEYWORD in org.eclipse.birt.data.engine.executor.aggregation.ProgressiveAggregationHelper$JSColumnBindingRow as seen in org.eclipse.birt.data.engine.script.JSResultSetRow also made some clean up.

@hvbtup, can you do some hardcore testing on this? This seems like a small fix, and I think that it is done correctly, but one never knows.


wimjongman:
Can you make a test for it, Claes?

claesrosell:
I can make a try.

hvbtup:
Well, at least it's not failing anymore. But I'm not sure if the result is correct. I'll explain in #885.

hvbtup:
Can you explain fd342f7dd15d3a963175ab43f10979c62bc08b57?

claesrosell:
The problem is that the Accumulator.onRow(Object[] args) is called even if the resultset is empty.
It is the groupCalculator that is causing this faulty call, and since it seems like an error to call the groupCalculator when there are no data in the resultset I looked on the behaviour of how the final row in the resultset is treated. (see doNext() in SimpleResultSet).
This is not neccesserily right though, since this.groupCalculator.next(0); seems to have been called from prepareFirstRow() even before you reported https://bugs.eclipse.org/bugs/show_bug.cgi?id=438257 . I will continue my investigation. The commit was more of a "unit test run" since my local engine tests does not work.

Do you know the exact version this stopped working?

wimjongman:
The build is working again. Seems to be a temp error.

claesrosell:
This PR now only contains the fix for the NPE when using row.__rownum.
For the original problem described in https://bugs.eclipse.org/bugs/show_bug.cgi?id=438257, see #922

claesrosell:
I am working on a unit test for this (using data extraction api), but I do not have much time right now so it will probably take another week.

wimjongman:
@claesrosell can we take this for 4.10 which I plan to release soon?

wimjongman (approved):
Looks good to me! Thanks, Claes.

Issue #922, opened 2022-04-07
SimpleResultSet.groupCalculator() called for index(0) even if the RowResultSet is empty


claesrosell:
The group calculator and hence the aggregator function is called even on empty result sets, this leads to erroneous results for certain aggregator functions, like COUNT.

hvbtup:
Old pages which describe the behavior:

https://www.eclipse.org/forums/index.php/t/848630/
Aggregation Count returns "1" when 0 row returned

https://bugs.eclipse.org/bugs/show_bug.cgi?id=438257 Visibility expression on table based on COUNT aggregate not working since BIRT 4.3.0

PR #923, opened 2022-04-07
SimpleResultSet.groupCalculator() called on empty RowResultSet #922

claesrosell:
Added a check to see if there are any data in the result set before
calling IGroupCalculator.next() on the first row.

Will try to add unit test for this.


claesrosell:
I am working on a unit test for this (using data extraction api), but I do not have much time right now so it will probably take another week.

claesrosell:
@hvbtup can you test this one with one of your data sets. This should fix the problem where "count" returns 1 for empty result sets. As a special bonus, this PR also includes a unit test for this case. This bug probably affected all aggregations but probably only had a real effect on "count".

hvbtup:
> can you test this one with one of your data sets.

That will have to wait for some days, but I will do it.

claesrosell:
@hvbtup: Do you think that you can test this in the coming week? This PR will hopefully solve this old issue, that you filed 8 years ago. I think that it would make a good addition to 4.10.
If you accept this fix, and it is committed, I plan to go forth with the fix for #885 (#914), which will add "access to row data in the progressive aggregator helper"

wimjongman:
Fetch the reference to the pull request based on its ID number, creating a new branch in the process.

(#923 

$ git fetch origin pull/#923/head:BRANCHNAME

Switch to the new branch that's based on this pull request:

[main] $ git checkout BRANCHNAME
> Switched to a new branch 'BRANCHNAME'

hvbtup:
Same error:

[ERROR] Failed to resolve target definition C:\Users\hvb\birt-master\git\birt\build\org.eclipse.birt.target\org.eclipse.birt.target.target: Could not find "org.eclipse.justj.openjdk.hotspot.jre.full.feature.group/11.0.14.v20220314-1349" in the repositories of the current location -> [Help 1]

hvbtup (approved):
I tested with the count-bug.rptdesign report (and with the other report there) and can confirm that this PR fixes the bug.

PR #925, opened 2022-04-11
QR-Code feature

hvbtup:
I created a new plugin extension for creating QR-Codes.

The important properties can be set in dialogs.
Other properties such as "QR-Code version" (aka dimension) and "Error Correction Level" can be set in advanced properties.

The foreground color is black, the background color is transparent. 

The size of the image is depending on the "imageDpi" property of the report. This is similar to the RotatedText item.

The code uses the ZXing library internally.

Questions / discussions:

Should we integrate this into BIRT directly (as "batteries included") or should I create a separate repository for this?

Is using the zxing library OK (license)?

Any important features missing?

How to add translations for the menu entries?

I'm not satisfied with the imageDpi dependency - or to be more precise - unsatisfied with the fact that at runtime, the size is depending on the imageDpi property, but the size in the designer seems to assume a fixed dpi value (of 96?). For example, if I set imageDpi to 300, then the image is much large in the designer than at runtime.
Note: The RotatedTextItem has a similar issue - the font size must be set much bigger depending on imageDpi.


hvbtup:
I still don't understand why - when creating a PR - GH wants to "append" commits to older commits. Only https://github.com/eclipse/birt/pull/925/commits/983e9cf4a8f280d7c9343ed46a70edff03ef330e is relevant for the QR-Code plugin.

wimjongman:
Because you are working on the wrong fork. I suggest you move your flugtiger fork to another github account and make a new fork from our repo.

ruspl-afed:
> Is using the zxing library OK (license)?

What exactly the license there? ASL-2.0 is OK by default unless we have some surprises either in dependencies or in sources.

hvbtup:
To the best of my knowledge, the license for ZXing is Apache Software License, Version 2.0

extension-qr-code/src/org/eclipse/birt/extension/qrcode/QRCodeBuilder.java
```diff
@@ -0,0 +1,179 @@
...
+		super(shell);
+
+		this.qrItem = qrItem;
+	}
+
+	@Override
+	protected void configureShell(Shell newShell) {
+		super.configureShell(newShell);
+
+		newShell.setText("QRCode Builder"); //$NON-NLS-1$
```
ruspl-afed, line 86: Please don't mark strings that requires translation with NON-NLS

extension-qr-code/src/org/eclipse/birt/extension/qrcode/QRCodeBuilder.java
```diff
@@ -0,0 +1,179 @@
...
+
+	@Override
+	protected void okPressed() {
+
+		try {
+			qrItem.setText(txtText.getText());
+			qrItem.setDotsWidth(Integer.parseInt(spDotsWidth.getText()));
+			qrItem.setEncoding(txtEncoding.getText());
+		} catch (Exception ex) {
+			ex.printStackTrace();
```
ruspl-afed, line 174: We should not expect any exceptions here. If some input is invalid - "OK" should be disabled

extension-qr-code/src/org/eclipse/birt/extension/qrcode/views/QRCodePageGeneratorFactory.java
```diff
@@ -0,0 +1,33 @@
...
+import org.eclipse.birt.report.designer.ui.views.IPageGenerator;
+import org.eclipse.core.runtime.IAdapterFactory;
+
+/**
+ * RotatedTextPageGeneratorFactory
+ */
+public class QRCodePageGeneratorFactory implements IAdapterFactory {
+
+	public Object getAdapter(Object adaptableObject, Class adapterType) {
+		return new QRCodePageGenerator();
```
ruspl-afed, line 26: It should return this only in case if `org.eclipse.birt.report.designer.ui.views.IPageGenerator` was requested

extension-qr-code/.project
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>extension-qr-code</name>
```
ruspl-afed, line 3: Why don't we have folder name and project name in sync with `Bundle Symbolic Name?

ruspl-afed (changes requested):
This should be included to the build to be visible for our pipeline

extension-qr-code/src/org/eclipse/birt/extension/qrcode/views/QRCodePageGeneratorFactory.java
hvbtup: Can you be more precise? I don't know what I should test here (the interface is too generic)

extension-qr-code/.project
hvbtup: Thanks for looking at the code. I'll try to change the code accordingly. BTW most of what you say is because I took the RotatedText plugin as a template.

extension-qr-code/src/org/eclipse/birt/extension/qrcode/QRCodeBuilder.java
hvbtup: Changed.

hvbtup:
> This should be included to the build to be visible for our pipeline

Sorry, I don't know what to do here.

extension-qr-code/src/org/eclipse/birt/extension/qrcode/views/QRCodePageGeneratorFactory.java
ruspl-afed: Perhaps you are not familiar with this contract of `IAdaptapleFactory`
It is all about adapting `adaptableObject` to a type `adapterType`

So, what I would expect to see is like this
```
if(IPageGenerator.class.isInstance(adapterType)) {
    return new QRCodePageGenerator();   
}
return null;
```

extension-qr-code/src/org/eclipse/birt/extension/qrcode/views/QRCodePageGenerator.java
```diff
@@ -0,0 +1,79 @@
...
+	public void createTabItems(List input) {
+		if (generalPage == null || generalPage.getControl().isDisposed()) {
+			tabFolder.setLayout(new FillLayout());
+			generalPage = AttributesUtil.buildGeneralPage(tabFolder,
+					new String[] { null, AttributesUtil.BORDER, AttributesUtil.MARGIN, AttributesUtil.SECTION,
+							AttributesUtil.VISIBILITY, AttributesUtil.TOC, AttributesUtil.BOOKMARK,
+							AttributesUtil.USERPROPERTIES, AttributesUtil.NAMEDEXPRESSIONS,
+							AttributesUtil.ADVANCEPROPERTY },
+					new String[] { "General" //$NON-NLS-1$
+					}, new String[] { "General" //$NON-NLS-1$
```
ruspl-afed, line 61: Please remove NON-NLS

extension-qr-code/src/org/eclipse/birt/extension/qrcode/views/QRCodeGeneralPage.java
```diff
@@ -0,0 +1,243 @@
...
+			if (children[i] instanceof Composite) {
+				adaptFormStyle((Composite) children[i]);
+			}
+		}
+
+		toolkit.paintBordersFor(comp);
+		toolkit.adapt(comp);
+	}
+
+	protected QRCodeItem getItem() {
```
ruspl-afed, line 175: It is better to return Optional<QRCodeItem> if it could be unavailable

extension-qr-code/src/org/eclipse/birt/extension/qrcode/views/QRCodeGeneralPage.java
```diff
@@ -0,0 +1,243 @@
...
+			spDotsWidth.setLayoutData(gd);
+			spDotsWidth.addFocusListener(new FocusAdapter() {
+
+				@Override
+				public void focusLost(org.eclipse.swt.events.FocusEvent e) {
+					updateModel(QRCodeItem.DOTS_WIDTH_PROP);
+				}
+			});
+
+			toolkit.createLabel(contentpane, "Encoding:"); //$NON-NLS-1$
```
ruspl-afed, line 111: Please remove NON-NLS

extension-qr-code/src/org/eclipse/birt/extension/qrcode/views/QRCodeGeneralPage.java
```diff
@@ -0,0 +1,243 @@
...
+
+		if (children != null && children.length > 0) {
+			contentpane = (Composite) children[children.length - 1];
+
+			GridLayout layout = new GridLayout(3, false);
+			layout.marginLeft = 8;
+			layout.verticalSpacing = 12;
+			contentpane.setLayout(layout);
+
+			toolkit.createLabel(contentpane, "Text Content:"); //$NON-NLS-1$
```
ruspl-afed, line 69: Please remove NON-NLS

extension-qr-code/src/org/eclipse/birt/extension/qrcode/util/SwtGraphicsUtil.java
```diff
@@ -0,0 +1,122 @@
...
+		try {
+			dest = new Image(display, width, height);
+			gc = new GC(dest);
+
+			gc.setAdvanced(true);
+			gc.setAntialias(SWT.OFF);
+
+			qrw = new QRCodeWriter();
+			HashMap<EncodeHintType, Object> hints = new HashMap<EncodeHintType, Object>();
+			hints.put(EncodeHintType.CHARACTER_SET, encoding != null ? encoding : "utf-8");
```
ruspl-afed, line 87: In Java it is more common to see "UTF-8" (in capital case), but I'm not sure regarding this particular case

extension-qr-code/src/org/eclipse/birt/extension/qrcode/util/SwingGraphicsUtil.java
```diff
@@ -0,0 +1,67 @@
...
+import com.google.zxing.common.BitMatrix;
+import com.google.zxing.qrcode.QRCodeWriter;
+import com.google.zxing.qrcode.decoder.ErrorCorrectionLevel;
+
+/**
+ * SwingGraphicsUtil
+ */
+public class SwingGraphicsUtil {
+
+	public static BufferedImage createQRCodeImage(String text, int dotsWidth, int dotsHeight, String encoding,
```
ruspl-afed, line 33: "create" method should never return `null`, it should throw exception instead if creation failed

extension-qr-code/src/org/eclipse/birt/extension/qrcode/util/SwtGraphicsUtil.java
```diff
@@ -0,0 +1,122 @@
...
+import com.google.zxing.common.BitArray;
+import com.google.zxing.common.BitMatrix;
+import com.google.zxing.qrcode.QRCodeWriter;
+import com.google.zxing.qrcode.decoder.ErrorCorrectionLevel;
+/**
+ * SwtGraphicsUtil
+ */
+public class SwtGraphicsUtil {
+
+	public static Image createQRCodeImage(String text, int dotsWidth, int dotsHeight, String encoding,
```
ruspl-afed, line 38: "create" method should never return `null`, it should throw exception instead if creation failed

extension-qr-code/src/org/eclipse/birt/extension/qrcode/QRCodeBuilder.java
```diff
@@ -0,0 +1,191 @@
...
+	protected void okPressed() {
+
+			try {
+				qrItem.setText(txtText.getText());
+				qrItem.setEncoding(txtEncoding.getText());
+				qrItem.setDotsWidth(Integer.parseInt(spDotsWidth.getText()));
+				super.okPressed();
+			} catch (SemanticException e) {
+				// Should not happen, because the dialog controls should prevent invalid data.
+				// If it happens nevertheless, we log this and we do not leave the dialog.
```
ruspl-afed, line 182: Sorry, but this is not the way how JFace Dialog should work. No `try-catch` should be in `okPressed`, it is too late to validate. We have button `enabled` state to control the ability to "apply and close the dialog", controls should be validated on value change and the result should be propagated to the `enabled` state

extension-qr-code/src/org/eclipse/birt/extension/qrcode/views/QRCodePageGeneratorFactory.java
hvbtup: Added this test.

extension-qr-code/src/org/eclipse/birt/extension/qrcode/views/QRCodeGeneralPage.java
hvbtup: I don't want to be the first to use newer language features in the code base, in particular when I'm not used to them myself.

extension-qr-code/src/org/eclipse/birt/extension/qrcode/util/SwingGraphicsUtil.java
hvbtup: I considered changing this, but I didn't, because the code wouldn't be cleaner IMHO. 
Besides, returning null and more or less silently ignore many exceptions seems to be the traditional BIRT coding style.

hvbtup:
> Because you are working on the wrong fork. I suggest you move your flugtiger fork to another github account and make a new fork from our repo.

Do you think it could help if I create a GH organization account, create a BIRT fork there, and still use the hvbtup account?

wimjongman:
> Do you think it could help if I create a GH organization account, create a BIRT fork there, and still use the hvbtup account?

Yes, that is a good idea.

hvbtup:
I'll create a new PR and close this one.

PR #929, opened 2022-04-12
Fix data regression tests

claesrosell:
Work in progress

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,6 +54,8 @@ jobs:
           ${{ runner.os }}-
     - name: Build with Maven
       run: mvn clean install --no-transfer-progress -U -DskipTests
+    - name: Data Test
+      run: mvn -B -pl :org.eclipse.birt.data.tests integration-test
     - name: Model Test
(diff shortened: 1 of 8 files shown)


claesrosell:
This adds another 1191 tests to the build.
I have disabled two tests.

- "FeaturesTest.test16" is ignored. It is time zone dependent and it seems like Derby does not store time zone date, so the "check against golden file" check fails if the test is run if the time zone changes.
- "CursorNavigatorTest.testNavigatorOnSubCursor" is removed. I will open an issue to fix this one, but currently, it is not for me comprehensible what the issue is. I think the test is ok, it is the implementation that fails. But why?

wimjongman:
Thanks, Claes. This is another great step forward.

> * "FeaturesTest.test16" is ignored. It is time zone dependent and it seems like Derby does not store time zone data for date and time types, so the "check against golden file" check fails if the test is run if the time zone changes.

I remember that I changed many tests to deal with the date formats. These tests were failing for different people and for others they were fine. 

I think I  explicitly set a locale before the tests, but I have to revisit it.

I created the issue for the two disabled tests.

wimjongman (approved):
Great work.

wimjongman:
Unless you want to retain the commit history, I suggest to Squash and Merge.

Issue #948, opened 2022-04-19
JavaScript IScriptEngineFactory is not found by ScriptEngineFactoryManagerImpl in non OSGI environments


claesrosell:
This was found while investigating issue #941. master currently fails in non-OSGI environments when running reports that use javascript:
```
org.eclipse.birt.report.engine.api.EngineException: Cannot create data engine.
   at org.eclipse.birt.report.engine.executor.ExecutionContext.openDataEngine(ExecutionContext.java:782)
   at org.eclipse.birt.report.engine.api.impl.RunTask.doRun(RunTask.java:199)
...
```
The script engines does not use the extension point attribute scriptID during look-up, but the scriptName instead. The scriptName attribute was recently made translatable and moved to a bundle.properties file. -Which triggers this problem.

In org.eclipse.birt.report.engine.script.javascript/plugin.xml :
```
      <scriptEngineFactory
            factoryClass="org.eclipse.birt.report.engine.javascript.JavascriptEngineFactory"
            scriptID="org.eclipse.birt.report.engine.javascript"
            scriptName="%scriptEngineFactory.scriptName">
      </scriptEngineFactory>
```
"scriptName" needs to be changed back to "javascript"

claesrosell:
Closed with #950

PR #950, opened 2022-04-20
JavaScript IScriptEngineFactory is not found by ScriptEngineFactoryManagerImpl in non OSGI environments #948

claesrosell:
scriptName is not translatable since it used for lookup in
org.eclipse.birt.core.internal.plugin.ScriptEngineFactoryManagerImpl


claesrosell:
Confirmed to work with birt-runtime-artifact

Issue #952, opened 2022-04-25
Build fails with messages "Package 'org.dom4j.bean' does not exist exist in this plugin" etc


hvbtup:
I am trying to build master, but this fails with errors:

![grafik](https://user-images.githubusercontent.com/7644776/165129053-ed298522-5c8e-41b4-930e-eb3baaf53586.png)

This seems to be related to changes to /build/org.eclipse.birt.build/META-INF/MANIFEST.MF by commit 91d6e8ff510a3e98b224df9241422878fa09dbb2

I really have no clue, but was adding these lines intended?

    +Export-Package: org.dom4j;x-internal:=true,
    + org.dom4j.bean;x-internal:=true,
    + org.dom4j.datatype;x-internal:=true,
    ...

claesrosell:
I have these errors locally as well but since the Maven build works I haven't looked into it. However, the commit that you reference probably added the "Export-Package" headers by mistake.
The org.eclipse.birt.build project states this in the readme.txt:

```
BIRT Verisoning Tools 
===========================

BIRT Versioning Tools is used for versioning control in current BIRT nightly build. 
It is an extended ant task and should be used with ant script. 
```
I can see no reason for why that project should export the dom4j packages.

PR #953, opened 2022-04-25
Build fails with "Missing dom4j messages" #952

claesrosell:
Removed the "Export-Package" headers for dom4j packages.


hvbtup:
Thanks Claes!

PR #984, opened 2022-05-27
New Feature: QR-Code item

hvbtup:
Note: This PR replaces PR #925 .

@ruspl-afed made some comments on #925. It seems he is cut off from GitHub at the moment.

I tried to take some of his remarks into account, but I'm not able to react to all of them, as my Java knowledge is a bit rusty.
So, to put things short, the Java code could probably use some polishing (which is true for 99% of the old code base as well), but I don't know what exactly I should do, so I suggest someone can improve it if it seems necessary. That way I could learn, too.


hvbtup:
BTW, now, with the new fork triestram-partner/birt *directly* from eclipse/birt, I can finally use the normal and easiest way to create a PR directly from the GH web site 😄.

claesrosell:
I am very excited about this feature and I will take a look at this in the coming weeks.

hvbtup:
I just noticed that the plugin works when I run the all-in-one designer as a .launch file, but it is is not part of the build/birt-packages directory generated by `mvn package -D skipTest`. What's necessary to make this happen?

ruspl-afed:
@hvbtup please have a look how over bundles are layed out:

Yes, you can place your "component" to form another subfolder "qrcode" in the repository root.
But then, please introduce a dedicated folder for bundle itself
`qrcode\.project` -> `qrcode\org.eclipse.birt.report.item.qrcode\.project`

After that I can help you to make it known to releng.

hvbtup:
I have to admit that I don't quite understand what exactly I should do.
This might all seem trivial to you, but I'm at a loss here. I don't see the pattern to follow.
Initially, I took the folder birt\tutorial\extension-tutorial-2 as a template.

ruspl-afed:
Thank you for your comment @wimjongman 
I think we will need to define a feature to include this new bundle. What it your preference?

claesrosell:
I just want to warn you: If this bundle ends up in a new feature and we want qr-code to be part of the Birt-Runtime, we will need to extend the some ant scripts. Ant scripts pull jars from a couple of different places to create both the Birt-Runtime and Birt-Runtime-OSGI artifacts. The jar "scrapping" is done in a couple of stages if I recall correctly.

wimjongman:
> Thank you for your comment @wimjongman I think we will need to define a feature to include this new bundle. What it your preference?

If we are going to ship it always then a new feature is not needed. Just put it in the feature where the rest of the UI goes.

wimjongman:
> I just want to warn you: If this bundle ends up in a new feature and we want qr-code to be part of the Birt-Runtime, we will need to extend the some ant scripts. Ant scripts pull jars from a couple of different places to create both the Birt-Runtime and Birt-Runtime-OSGI artifacts. The jar "scrapping" is done in a couple of stages if I recall correctly.

Noted. Thanks Claes.

ruspl-afed:
> If we are going to ship it always then a new feature is not needed. Just put it in the feature where the rest of the UI goes.

Do you mean [this one](https://github.com/eclipse/birt/blob/master/features/org.eclipse.birt.feature/feature.xml) @wimjongman ?

wimjongman:
> > If we are going to ship it always then a new feature is not needed. Just put it in the feature where the rest of the UI goes.
> 
> Do you mean [this one](https://github.com/eclipse/birt/blob/master/features/org.eclipse.birt.feature/feature.xml) @wimjongman ?

That looks good.

wimjongman:
Can this branch be rebased?

hvbtup:
I noticed that for the QR Code Item - as well as for the Rotated Text Item - the `onCreate` etc. events do not work.
The Javascript scripts are just not called.

I think it is important to have the opportunity to adjust properties dynamically.
Otherwise several typical use cases for the item would not work.

I tried to debug this and I found out that during `RunAndRenderTask.prepareDesign`, the item is visited (see call stack)
grafik
and then `ExtendedItemManager.getExtendedItem` puts a null entry into `extItems`.
Thus the following calls to `getExtendedItem` always return null and JS events won't be called.

Does anyone know more about how this is supposed to work?

Is there anything I can add to the Java code (for the QR code item or for the RotatedTextItem) to make JS onCreate, onPrepare, onRender events work?

The only other existing extended item I've heard of is the chart item. But this is really, really complex and thus not an option to use it as as code template.

I guess that there should be a recommended way to make `createExtendedItemFactory` return a non-null object.

grafik

hvbtup:
> > > If we are going to ship it always then a new feature is not needed. Just put it in the feature where the rest of the UI goes.
> > 
> > 
> > Do you mean [this one](https://github.com/eclipse/birt/blob/master/features/org.eclipse.birt.feature/feature.xml) @wimjongman ?
> 
> That looks good.

When I add the plugin to that file like this
grafik

Then the Maven build still fails:
ERROR Failed to execute goal org.eclipse.tycho:tycho-packaging-plugin:2.7.5:package-feature (default-package-feature) on project org.eclipse.birt: Execution default-package-feature of goal org.eclipse.tycho:tycho-packaging-plugin:2.7.5:package-feature failed: **eclipse-plugin artifact with ID "org.eclipse.birt.report.item.qrcode" and version matching "0.0.0" was not found in the target platform** -> [Help 1]

wimjongman:
Yes, you also have to add it to the modules to be compiled. I think chart/pom.xml

wimjongman:
Apologies, this needs to go into the extensions folder, so the pom to update is

Let me fix this for you.

hvbtup:
> Let me fix this for you.

@wimjongman Yes, please.
BTW did you see my question regarding JS events for ExtentedItems not working?

wimjongman:
We continue here: Qrcode feature #1087

Issue #974, opened 2022-05-19
CVE-2015-1832


SteveSchafer-Innovent:
Vulnerability in Apache Derby prior to 10.12.1.1.  
https://nvd.nist.gov/vuln/detail/CVE-2015-1832#match-6066870
https://issues.apache.org/jira/browse/DERBY-6807
it is fixed in 10.12.1.1

SteveSchafer-Innovent:
Derby 10.12.1.1 is included in the target file and additionally derby.jar is included in the common/org.apache.derby.core plugin.
All the persisted orbit repositories have derby 10.11 and the milestone builds don't have derby at all.   I don't know how to get derby re-included in orbit.  It would probably be easiest to include it as our own plugin, which would replace the existing org.apache.derby.core plugin, but I need to find out why it was done that way.

wimjongman:
Steve, whenever possible, artifacts should be consumed from the upstream providers. In this case, we can hopefully pull derby from maven. Inclusion should be the very last resort.

Tycho is able to consume bundles from maven, however, we need to make sure that they are OSGified (e.g. proper export package and bundle name)

SteveSchafer-Innovent:
I will change common/org.apache.derby.core to use the latest derby from maven and then try to replace the version of derby referenced in the target file.  I'm not exactly sure how to do that but there are examples to follow, e.g. org.apache.axis.

SteveSchafer-Innovent:
When you consume a jar file with maven, the jar file goes into your .m2 directory.  In the case of derby, there are ant scripts that copy its jar file.  I need to figure out how to specify derby as a dependency in a pom file and then locate the jar file in an ant script.

wimjongman:
Do you mean the term "releng"? If so, it is short for release engineering. Anything that has to do with build/package/deploy/install/etc..

wimjongman:
I would assume that the packagers create the standalone runtimes. Maybe a lot of that code got replaced by maven tycho.

You could try to delete the build.xml and buildall.xml and run the build to see where it fails.

SteveSchafer-Innovent:
The maven tycho builds create everything as far as I know, so I was wondering if the releng packages did anything useful.  The readme files are very old -- 2007, for BIRT 2.2.  I'll try your suggestion.  Maybe we can get rid of them altogether.

SteveSchafer-Innovent:
[...]

PR #995, opened 2022-06-08
CVE-2015-1832 #974

SteveSchafer-Innovent:
Replace common/org.apache.derby.core with common/org.apache.derby using
the latest derby in maven


merks:
I can look at incorporating derby into Orbit but the maven artifacts look to be effectively garbage.   E.g., the MANIFEST.MF mentions jars that are not nested in the artifact jar:

image

Also sources are missing:

image

Apache is able to bypass the normal publication rules.    So much for open source.

merks:
BIRT 4.17 is using derby 10.16.1 and provided by Orbit.   Orbit produces the jars by processing the direct downloads of the derby project:

https://github.com/eclipse-orbit/orbit-simrel/tree/main/maven-derby

Issue #989, opened 2022-06-01
Can we get rid of releng?


SteveSchafer-Innovent:
In trying to track down where scripts are executed (both ant builds and other scripts), I see there are some scripts in build/org.eclipse.birt.releng.birtbuilder and build/org.eclipse.birt.releng.util that are not called anywhere.  I deleted those files and my local build still succeeds.   There is no pom.xml in birtbuilder or util.  There is a pom.xml in birtbuilder/extras but it isn't referenced as a module in any other pom so it may not be called either.  There are readme files in birtbuilder and util but they are very old (2007).  Has all this releng code been replaced by maven and can we delete it?

wimjongman:
Does this build also still succeed if you do not have a local .m2 repo? Do the bundles get referenced in the build output?

SteveSchafer-Innovent:
The local build still succeeds if I delete the local .m2 directory.

As for the other question:  Which bundles, and which build output are you referring to?   Neither build/org.eclipse.birt.releng.birtbuilder nor build/org.eclipse.birt.releng.util have any bundles (MANIFEST.MF files) nor any build output (target folders).

wimjongman:
Sorry, I meant: do you see any reference to the scripts in the build output. If not I say go ahead and delete them. We can always pull them back from git.

SteveSchafer-Innovent:
The util project doesn't seem to be referenced anywhere other than projectSet.psf so I deleted it.  Some files in the birtbuilder project are used as follows:

projectSet.psf
build/org.eclipse.birt.build/repoGen.properties:readmeFile=../org.eclipse.birt.releng.birtbuilder/extras/runtime_readme.txt
build/birt-packages/birt-wtp-integration-sdk/pom.xml: copy html files from extras
build/birt-packages/birt-sample-plugins/pom.xml: copy html files from extras
build/birt-packages/birt-runtime-osgi/build.xml: copy genReport.sh from extras
build/birt-packages/birt-runtime-osgi/build.xml: copy genReport.bat from extras
build/birt-packages/birt-runtime-osgi/build.xml: copy html files from extras
build/birt-packages/birt-runtime-osgi/build.xml: copy runtime_readme.txt from extras
build/birt-packages/birt-charts/ChartSDK/pom.xml: copy html files from extras
build/birt-packages/birt-charts/pom.xml: copy html files from extras
build/birt-packages/birt-charts/ChartRuntime/pom.xml: copy html files from extras
[...]

PR #1004, opened 2022-06-28
Can we get rid of releng? #989

SteveSchafer-Innovent:
Delete build/org.eclipse.birt.releng.util.
Delete .sh and .xml files from build/org.eclipse.birt.releng.birtbuilder.


merks:
I believe we can throw much of it away, but there are still references to its content from elsewhere to take into consideration:

image

That's the case after all the unused ant scripts have been removed.

So I think "we need to selectively delete the folder/parts that are unused, confirming that the builds still work until we are left with only the parts still used.  Then we could consider how to factor those out...

merks:
I believe I've deleted all the unused content at this point.

wimjongman:
Thank you, Ed. That stuff was scaring me for years.

PR #1015, opened 2022-07-21
ClassCastException while emitting "postscript" format #1010

claesrosell:
During reflective access to the "trueTypeFonts" field of
"FontFactoryImpl" the type "Properties" was assumed, but the type has
changed to "HashMap" in OpenPDF. Stopped using reflective access to the
field since there are now API to access the content of "trueTypeFonts".


ruspl-afed (approved):
LGTM

hvbtup (commented):
Regarding the change to "getFontPath":
The old code used `fontName.toLowerCase()` while the new code just uses `fontName`.

Is that intented and what's the reason?
This could perhaps cause a regression on case-sensitive file systems.

claesrosell:
> Regarding the change to "getFontPath": The old code used `fontName.toLowerCase()` while the new code just uses `fontName`.
> 
> Is that intented and what's the reason? This could perhaps cause a regression on case-sensitive file systems.

The .getFontPath() method on FontFactoryImpl is doing the .toLowerCase() call.

hvbtup (approved):
OK then.

Issue #1042, opened 2022-09-12
BIRT Report Designer 4.9.0. - Chart don't work in View Report as PDF


rankdev01:
I've installed BIRT report designer 4.9.0. traing to test some reports, and I notice that reports that contains Chart elements are not well generated in PDF format.

The browser shows the message: "An error occurred while loading the PDF document."

Windows 10 x64: birt-rcp-report-designer-4.9.0-20220315-win32.win32.x86_64.zip
Example:
[chart_test.zip](https://github.com/eclipse/birt/files/9546500/chart_test.zip)
Tested with JVM 11 and 18

speckyspooky:
Hello Together,

I confirm there is an issue with the rcp-designer. I started some retests with the same configuration like above with JVM11.
The issue is located at the services of POSTSCRIPT and PDF. Both output formats cann't be processed.

**Processing of PS error message:**

![BIRT-PS-error](https://user-images.githubusercontent.com/41593722/189985555-097df51a-14d7-4575-a139-be65756dd96d.png)

**Processing PDF - result is an empty PDF:**

![BIRT-PDF-error](https://user-images.githubusercontent.com/41593722/189985708-f13e5605-098b-4755-a46b-d06a984b5cbe.png)

Both errors are reproducable with the BIRT-viewer and with the direct processing call from the designer (without designer).
The other output formats can be produced correctly.


**Regression test (manual) BIRT 4.9 with the "all-in-one"-designer (JVM11):**
- PDF-document created correctly
- PS-document ended in error above

**Regression test (manual) BIRT 4.8 with the "all-in-one"-designer (JVM8):**
- PDF-document created correctly
- PS-document created correctly

Can somebody assists with it. It seams to be an issue with the SOAP-service requests or the emitter service registration.

Best regards

hvbtup:
I can confirm the PS issue with a (custom) 4.10 I built April 2022.

Part of the stack trace is:

Caused by: **java.lang.ClassCastException: class java.util.HashMap cannot be cast to class java.util.Properties (java.util.HashMap and java.util.Properties are in module java.base of loader 'bootstrap')**
	at org.eclipse.birt.report.engine.emitter.postscript.PostscriptWriter.getFontPath(PostscriptWriter.java:741)
	at org.eclipse.birt.report.engine.emitter.postscript.PostscriptWriter.applyFont(PostscriptWriter.java:682)
[...]

PR #1070, opened 2022-09-27
BIRT Report Designer 4.9.0. - Chart don't work as PDF #1042

wimjongman:
Created an alternative configuration for the RCP designer Move embedded jre to 17


wimjongman:
This has been fixed.

PR #1071, opened 2022-09-29
Fix BIRT.setup to avoid installing unsupported m2e connectors

merks:
https://github.com/eclipse/birt/discussions/1060


merks:
@wimjongman 

There are a few simple changes to make it work better.

Note that I noticed a major problem related to this:

https://bugs.eclipse.org/bugs/show_bug.cgi?id=580807

Here there is more than one `-Dosgi.requiredJavaVersion` in the eclipse.ini and if the last one is for Java 11, then the m2e plugins fail to start and the target platform fails to resolve fully resulting in a sea of red errors.

@jonahgraham  FYI.

wimjongman:
Ed, I am super thankful for your support. I will give this a spin.

PR #1084, opened 2022-10-12
Fix duplicate Rotated text item in run configurations and fix SWT war…

hvbtup:
…ning about missing dispose


hvbtup:
Can someone with a Mac test the MAC all-in-one run configuration, please?
The intention was to only remove the line for `org.eclipse.birt.report.extension.tutorial_1`.
I don't know why the IDE changed several other lines in the `BIRT all-in-one-MAC.launch` file.

hvbtup:
Part of this was solved with #1191 .
I'll create an issue for the SWT warning.

wimjongman:
`public static Image createRotatedTextImage(String text, int angle, Font ft) {
`

I think this method should have javadoc explaining what it does AND that the caller is responsible for disposing of the created image.

PR #1112, opened 2022-11-17
Bug 580994 - BIRT 4.10.0-20221001 - CVE-2021-34427 fix bypass

wimjongman:
Fixes an exploit in the BIRT viewer to block file extensions with special characters.


wimjongman:
See https://bugs.eclipse.org/bugs/show_bug.cgi?id=580994

Issue #1123, opened 2022-12-01
Generated PDF ignores "line-height" CSS property for "line-height" < 100%


luzhanov:
We're using BIRT 4.9.0 for generating PDF reports. 
The issue we've faced is that generated PDFs doesn't respect the CSS property `line-height` in cases when this property is < 100%.

Debugging shows that the reason is logic in `LineArea` and `InlineStackingArea` classes. I will provide a PR which fixes this. 

Attached file is a sample` .rptdesign` template which shows the problem. 
[line-height-test-20221201-1915.zip](https://github.com/eclipse/birt/files/10135161/line-height-test-20221201-1915.zip)

This issue was present in BIRT 4.4.2 also.

luzhanov:
PR with issue fix: https://github.com/eclipse/birt/pull/1124

PR #1124, opened 2022-12-01
Fix for the PDF "line-height" < 100% bug #1123

luzhanov:


wimjongman:
Thanks for the patch, Igor. It looks good to me. 

Before we can accept your patch we need you to sign the ECA.
https://accounts.eclipse.org/user/eca

luzhanov:
Thank you, Wim! 
Here is my Eclipse profile with ECA signed: https://accounts.eclipse.org/users/luzhanov

PR #1147, opened 2022-12-21
Remove version from splash screen

wimjongman:


wimjongman:
already committed in another pr

Issue #1164, opened 2023-01-03
Fix report path vulnerability


wimjongman:
In some cases, it is possible to execute javascript code on the server.

wimjongman:
Fixed in 4.13

PR #1165, opened 2023-01-03
Fix report path vulnerability #1164

wimjongman:
Hardened the capability of running reports with an absolute file path.


wimjongman:
Verified!

PR #1180, opened 2023-01-12
Add special handling for Soft Hyphen (SHY) unicode symbol to DOCX emi…

hvbtup:
…tter and PDF emitter.

Added support for correctly handling the SHY symbol (Unicode 173) to the PDF emitter and the DOCX emitter.

### DIFF
--- a/engine/org.eclipse.birt.report.engine.emitter.wpml/src/org/eclipse/birt/report/engine/emitter/wpml/writer/AbstractWordXmlWriter.java
+++ b/engine/org.eclipse.birt.report.engine.emitter.wpml/src/org/eclipse/birt/report/engine/emitter/wpml/writer/AbstractWordXmlWriter.java
@@ -32,6 +32,10 @@ import org.eclipse.birt.report.engine.emitter.wpml.WordUtil;
 import org.eclipse.birt.report.engine.layout.pdf.util.PropertyUtil;
 import org.w3c.dom.css.CSSValue;
 
+/**
+ * This is used for writing WordML by the DocxEmitter and by the old Word 2003
+ * emitter.
+ */
 public abstract class AbstractWordXmlWriter {
 
 	protected XMLWriter writer;
@@ -50,7 +54,29 @@ public abstract class AbstractWordXmlWriter {
 
 	public static final int INDEX_NOTFOUND = -1;
 
-	protected final char SHY_CHAR = (char) 173;
+	/**
+	 * <p>
+	 * The soft hyphen Unicode symbol is intended to be visible only when a line
+	 * break occurs there.
+	 * </p>
+	 * <p>
+	 * This hiding logic of the SHY symbol needs special attention in many emitters.
+	 * </p>
+	 * <p>
+	 * SOFT HYPHEN is often abbreviated as SHY, which also is very descriptive,
+	 * because this symbol is hiding inside the surrounding words most of the time.
+	 * </p>
+	 * <p>
+	 * In most fonts, its width is defined as zero, which of cause is correct only
+	 * if it is hidden. If it is rendered, it looks similar to the minus sign.
+	 * </p>
+	 * <p>
+	 * The Unicode standard also defines a HYPHEN symbol, which should look the same
+	 * as the SHY symbol, but doesn't have the hiding logic. However, the HYPHEN
+	 * symbol is rarely defined in TTF fonts.
+	 * </p>
+	 */
+	public static final char SOFT_HYPHEN = '\u00ad';
 
 	protected int imageId = 75;
 
@@ -556,8 +582,8 @@ public abstract class AbstractWordXmlWriter {
 					start++;
 				}
 				end = start + 1;
-			} else if (ch == SHY_CHAR) {
-				// output previous text
+			} else if (ch == SOFT_HYPHEN) {
+				// Output a special WordML tag for the SHY symbol.
 				writeText(txt.substring(start, end));
 				writer.closeTag("w:t"); //$NON-NLS-1$
 				writer.cdata("<w:softHyphen/>"); // $NON-LS-1$
@@ -1012,11 +1038,8 @@ public abstract class AbstractWordXmlWriter {
 	 * @param cellWidth  the width of the container in points
 	 * @return String with truncated words that surpasses the cell width
 	 */
-	public String cropOverflowString(String text, IStyle style, String fontFamily, int cellWidth) {// TODO: retrieve
-																									// font type and
-																									// replace plain
-																									// with
-																									// corresponding
+	public String cropOverflowString(String text, IStyle style, String fontFamily, int cellWidth) {
+		// TODO: retrieve font type and replace plain with corresponding
 		Font font = new Font(fontFamily, Font.PLAIN, WordUtil
 				.parseFontSize(PropertyUtil.getDimensionValue(style.getProperty(StyleConstants.STYLE_FONT_SIZE))));
 		Canvas c = new Canvas();
--- a/engine/org.eclipse.birt.report.engine.emitter.wpml/src/org/eclipse/birt/report/engine/emitter/wpml/writer/AbstractWordXmlWriter.java
+++ b/engine/org.eclipse.birt.report.engine.emitter.wpml/src/org/eclipse/birt/report/engine/emitter/wpml/writer/AbstractWordXmlWriter.java
@@ -50,6 +50,8 @@ public abstract class AbstractWordXmlWriter {
 
 	public static final int INDEX_NOTFOUND = -1;
 
+	protected final char SHY_CHAR = (char) 173;
+
 	protected int imageId = 75;
 
 	protected int bookmarkId = 0;
@@ -554,9 +556,17 @@ public abstract class AbstractWordXmlWriter {
 					start++;
 				}
 				end = start + 1;
-				continue;
+			} else if (ch == SHY_CHAR) {
+				// output previous text
+				writeText(txt.substring(start, end));
+				writer.closeTag("w:t"); //$NON-NLS-1$
+				writer.cdata("<w:softHyphen/>"); // $NON-LS-1$
+				writer.openTag("w:t"); //$NON-NLS-1$
+				start = end + 1;
+				end++;
+			} else {
+				end++;
 			}
-			end++;
 		}
 		writeText(txt.substring(start));
 
--- a/engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/layout/pdf/hyphen/Word.java
+++ b/engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/layout/pdf/hyphen/Word.java
@@ -18,6 +18,19 @@ public class Word {
 	protected int end;
 	protected String text;
 
+	private boolean keepLastSHY = true;
+
+	/**
+	 * Should a trailing SHY symbol be kept or omitted?
+	 */
+	public boolean isKeepLastSHY() {
+		return keepLastSHY;
+	}
+
+	public void setKeepLastSHY(boolean keepLastSHY) {
+		this.keepLastSHY = keepLastSHY;
+	}
+
 	public Word(String text, int start, int end) {
 		this.text = text;
 		this.start = start;
--- a/engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/layout/pdf/hyphen/Word.java
+++ b/engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/layout/pdf/hyphen/Word.java
@@ -10,27 +10,72 @@
  *
  * Contributors:
  * Actuate Corporation - initial API and implementation
+ * Henning von Bargen - Added at least a bit of JavaDoc, added SOFT HYPHEN support.
  ***********************************************************************/
+
 package org.eclipse.birt.report.engine.layout.pdf.hyphen;
 
+/**
+ * <p>
+ * Despite its name, this describes a <em>fragment</em> of a word of text.
+ * </p>
+ * <p>
+ * If the word does not contain possible hyphenation / line-breaking points,
+ * then it is a whole word. But if the word contains Unicode MINUS or HYPHEN or
+ * SOFT HYPHEN symbols, then the {@link BreakIterator} splits this whole word
+ * into more than one Word instances.
+ * </p>
+ * <p>
+ * For example, "extra-ordinary" will be split into two Word instances "extra-"
+ * and "ordinary".
+ * </p>
+ */
 public class Word {
 	protected int start;
 	protected int end;
 	protected String text;
 
-	private boolean keepLastSHY = true;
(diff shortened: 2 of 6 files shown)


hvbtup:
The code is not yet perfect, but I think it is a big step in the right direction.

Some remarks:

1) There is another unicode symbol (hyphen, "\u2010") which I did not test.

2) MS Excel does not support the soft hyphen symbol "\u00ad". I wonder if it would be better to remove the SHY symbol for the Excel output.

3) I did not test other emitters except PDF, DOCX and HTML and XLSX.

4) The code for removing the soft hyphen symbol from text could probably use some performance tuning, but frankly I didn't see an obvious way to do so. As long as hyphenation is relatively rare, this is not an issue.

5) Now that we found the places in the code, **this opens the door for automatic hyphenation**.

6) This is unrelated, but it seems like great parts of BIRT's source code is unused. Eg. there are two classes called `TextCompositor` which look very similar. Also there's `o.e.b.report.engine.layout` and `o.e.b.report.engine.nLayout`.

More details regarding 5):

In modern HTML, there are some CSS properties related to line-breaking:
 * `hyphens` 
 * `overflow-wrap`
 * `word-break`
 * `hyphenate-character`

These are described in more detail here: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Text/Wrapping_Text

I did not examine the other properties, but regarding the `hyphens` property, the status quo of BIRT is:

**BIRT behaves as if `hyphens: manual` was specified.**

Without this PR, however, the behavior is incorrect, because the SHY symbol does not work as expected.
Minus symbols cause a correct line-break. Part of this logic already happens in ICU (see `WordRecognizerWrapper` and `com.ibm.icu.text.BreakIterator`.

With this PR, the behavior is correct regarding the SHY symbol.

I think it would be possible (but of course not part of this PR) to support `hyphens` in a better way in BIRT:

Firstly, document that BIRT's default value `manual` differs from HTML CSS default value `none`.

Next steps:

Make BIRT recognize and use the `hyphens` CSS property.

Add support for `hyphens: none`?
The HTML default value `hyphens: none` does not make much sense to me.
Since the `WordRecognizerWrapper` already splits words, it would be necessary to "collect and join" the `Word` instances which it delivers when `hyphens: none` is specified. This should be possible, but I think this is not an important feature.

Add support for `hyphens: auto`:
Automatic hyphenation can only work if the language is known, because hyphenation is language-specific. CSS requires that the `lang` attribute is present. I don't know if we already have something similar to the `lang` attribute in BIRT. If not, then it would make sense to add it anyway, because it is required for accessibility, too.
CSS specifies that a pre-hyphenated word always overrules automatic hyphenation, e.g. if a word contains SHY symbols or hyphen symbols (not minus-signs), then only the locations marked by these symbols must be taken into account for word-breaking. OK, that should be easy.
To support automatic hyphenation, there exist open source libraries. e.g. hunspell/hyphen (I'm pretty sure there are Java wrappers or pure Java libraries as well).
As there are different libraries and different languages, probably it makes no sense trying support all libraries and languages.
Instead, automatic hyphenation support for a specific should use the existing plugin concept.

engine/org.eclipse.birt.report.engine.emitter.wpml/src/org/eclipse/birt/report/engine/emitter/wpml/writer/AbstractWordXmlWriter.java
```diff
@@ -50,6 +50,8 @@ public abstract class AbstractWordXmlWriter {
 
 	public static final int INDEX_NOTFOUND = -1;
 
+	protected final char SHY_CHAR = (char) 173;
```
wimjongman, line 53: It took me a while to understand that SHY is SOFT_HYPHEN. Can you be more explicit about the name? You can remove CHAR from the name. Maybe you can also be more explicit in the declaration:

  protected final char SOFT_HYPHEN  = "\u00AD".charAt(0); // did not test this

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/layout/pdf/hyphen/Word.java
```diff
@@ -18,6 +18,19 @@ public class Word {
...
+	private boolean keepLastSHY = true;
+
+	/**
+	 * Should a trailing SHY symbol be kept or omitted?
+	 */
+	public boolean isKeepLastSHY() {
+		return keepLastSHY;
+	}
+
+	public void setKeepLastSHY(boolean keepLastSHY) {
```
wimjongman, line 30: No Javadoc.
rename to setKeepLastSoftHyphen
add @since 4.13 tag

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/layout/pdf/hyphen/Word.java
```diff
@@ -18,6 +18,19 @@ public class Word {
 	protected int end;
 	protected String text;
 
+	private boolean keepLastSHY = true;
+
+	/**
+	 * Should a trailing SHY symbol be kept or omitted?
+	 */
+	public boolean isKeepLastSHY() {
```
wimjongman, line 26: isKeepLastSoftHyphen ?

Please add @since 4.13 tag
Please add javadoc for every input and output param

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/nLayout/area/impl/TextArea.java
```diff
@@ -115,12 +121,27 @@ public int getTextLength() {
 		return textLength;
 	}
 
-	private String calculateText() {
+	public String getRawText() {
```
wimjongman, line 124: Does this really need to be public (API)

Always add proper javadoc for public methods

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/nLayout/area/impl/TextArea.java
```diff
@@ -28,6 +28,12 @@ public class TextArea extends AbstractArea implements ITextArea {
 
 	protected String cachedText = null;
 
+	private final char SHY_CHAR = (char) 173;
+
+	private boolean removeShy = true; // FIXME This should be configurable, depending on the emitter.
```
wimjongman, line 33: Get it from a property for now?

e.g. 

    rivate boolean removeSoftHyphen = "true".equals(System.getProperty("org.eclipse.birt.softhyphen.default.??", "true"));

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/nLayout/area/impl/TextArea.java
```diff
@@ -28,6 +28,12 @@ public class TextArea extends AbstractArea implements ITextArea {
 
 	protected String cachedText = null;
 
+	private final char SHY_CHAR = (char) 173;
```
wimjongman, line 31: Duplicate definition. Can we move this to a central place?

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/nLayout/area/impl/TextArea.java
```diff
@@ -115,12 +121,27 @@ public int getTextLength() {
 		return textLength;
 	}
 
-	private String calculateText() {
+	public String getRawText() {
+		return text.substring(offset, offset + textLength);
+	}
+
+	private String calculateText(boolean removeShy) {
```
wimjongman, line 128: add javadoc for input parm

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/nLayout/area/impl/TextArea.java
```diff
@@ -237,4 +258,14 @@ public void setWhiteSpaceNumber(int whiteSpaceNumber) {
 	public boolean needClip() {
 		return needClip;
 	}
+
+	public boolean isKeepLastSHY() {
```
wimjongman, line 262: Add javadoc
make method name more clear SoftHyphen

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/nLayout/area/impl/TextCompositor.java
```diff
@@ -157,8 +159,9 @@ private TextArea getNextTextArea(int maxLineWidth) {
 			textArea.setMaxWidth(maxLineWidth);
 			textArea.setWidth(0);
 			addWordIntoTextArea(textArea, remainWord);
+			textArea.setKeepLastSHY(remainWord.isKeepLastSHY());
 			remainWord = null;
-			return textArea;
+			return textArea; // Why do we return here already?
```
wimjongman, line 164: Why add your confusion to the source code?

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/nLayout/area/impl/TextCompositor.java
```diff
@@ -39,6 +39,8 @@ public class TextCompositor {
 	private FontInfo fontInfo;
 	private int runLevel;
 
+	private final String SHY_STRING = "\u00ad";
```
wimjongman, line 42: A third time SHY is defined. Please move this to a central place (CONSTANTS). Please do not add STRING but make the name more clear SOFT_HYPHEN

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/nLayout/area/impl/TextCompositor.java
```diff
@@ -250,13 +253,19 @@ private void addWordsIntoTextArea(TextArea textArea, IWordRecognizer words) {
...
 		// get the word's size
 		int textLength = word.getLength();
 		int wordWidth = getWordWidth(fontInfo, word);
 		// append the letter spacing
 		wordWidth += textStyle.getLetterSpacing() * textLength;
 		int adjustWordSize = fontInfo.getItalicAdjust() + wordWidth;
-		if (textArea.hasSpace(adjustWordSize)) {
+		int hyphenWidth = 0;
+		if (word.getValue().endsWith(SHY_STRING)) {
+			hyphenWidth = getTextWidth(fontInfo, "-"); // We are using the minus for computing the hyphen size, because
```
wimjongman, line 265: Suggestion to move that long comment to the Javadoc of getTextWidth??

why call getTextWidth at all if you know that it returns 0?

wimjongman (changes requested):
I have made some comments.

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/nLayout/area/impl/TextCompositor.java
hvbtup: I think this return statement is a weakness in the algorithm used.
It results in the next line consisting of *two* TextAreas.
Say you have the word "super-cali-fragi-listic-expli-ali-docious" (I'm using minus signs instead of SHY here for clarity).
Assume this word appears at the end of line 1, and only part of it fits into line 1.
So line 1 consists of one TextArea ["..... This is an super-"]
Line 2 consists of two TextAreas: ["cali-", "fragilisticexplialidocious idea"].
For PDF rendering, we have to remove the trailing "-" from the first TextArea, which was quite easy to achieve.

But this also results in *two* text runs in the PDF for line 2 (didn't verify this).
So in the resulting PDF line 2 is logically ["cali", "fragilisticexplialidocious idea"].

This probably makes it a little harder for accessibility software to understand that in fact this should read
"califragilisticexplialidocious idea".

Also the PDF is a few bytes longer than needed.

The rest of the code tries to stuff as many "words" as possible into one TextArea, but in this place, it results in an additional TextArea containing only one "word" ("cali-" in my example).

So this return statement kind of contradicts the idea of the remaining code.

I didn't dare to change it, though, because I needed to focus on the task at hand.

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/nLayout/area/impl/TextCompositor.java
hvbtup: We could do that. However, I don't see the point adding an additional level of indirection and a reference to another class just to avoid a single-character constant. Practicality beats purity.

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/nLayout/area/impl/TextCompositor.java
hvbtup: The trick is that the argument for getTextWidth is actually the good old ASCII MINUS character, not the SHY character.
A font will return a width of 0 for the SHY character. However, when it's visible at the end of the line, its width cannot be 0 actually.
Why should that comment go into the Javadoc comment for getTextWidth?
I think it's better to just use additional comment lines inside the code (it's really an implementation detail).

engine/org.eclipse.birt.report.engine.emitter.wpml/src/org/eclipse/birt/report/engine/emitter/wpml/writer/AbstractWordXmlWriter.java
wimjongman: If this is the correct and accepted term then it is fine with me.

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/nLayout/area/impl/TextCompositor.java
wimjongman: Yes, that is a great analysis which I would not have guessed from `// Why do we return here already?`

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/nLayout/area/impl/TextCompositor.java
wimjongman: Fine with me if there is no Constants class around.

hvbtup:
I renamed some attributes and method names and added a lot of comments in reaction to Wim's suggestions.

hvbtup:
@wimjongman @claesrosell Can we merge this?

wimjongman (approved):
LGTM!

wimjongman:
Henning, please squash and merge but make sure you make a good commit comment. Something like this comment https://github.com/eclipse/birt/pull/1180#issuecomment-1381534507 but maybe a bit more condensed.

wimjongman:
Thank you for this contribution, Henning!

PR #1195, opened 2023-01-30
Using the new UserProperty "defaultDpi" we can now ovveride the size …

hvbtup:
…of the rendered image, at least for the PDF emitter.

First change:
For background images (bitmap and SVG), the report's "imageDpi" setting is used instead of 96 dpi.
So e.g. with a setting of 300 thes background images are rendered smaller but sharper

Second change:
This can be overridden per item with an Integer UserProperty "imageDpi".

Note: I don't know if this is the way to go - it's just one way which allows me to workaround a regression in BIRT:

Earlier BIRT versions (4.2.1 and 4.3.0) "just worked" for *bitmap* images (I don't know why and how: Did they read the DPI metadata from the image or did they take background-image-height and -width into account?).

It would probably be better to use "background-image-height" and "background-image-width" properties for the size calculation.

### DIFF
--- a/engine/org.eclipse.birt.report.engine.emitter.html/src/org/eclipse/birt/report/engine/emitter/html/HTMLReportEmitter.java
+++ b/engine/org.eclipse.birt.report.engine.emitter.html/src/org/eclipse/birt/report/engine/emitter/html/HTMLReportEmitter.java
@@ -3221,10 +3221,15 @@ public class HTMLReportEmitter extends ContentEmitterAdapter {
 			String uriString = EmitterUtil.getBackgroundImageUrl(imageStyle, design,
 					this.report.getReportContext() == null ? null : this.report.getReportContext().getAppContext());
 
+			Integer dpi = null;
+			// As we are at the report level, "overriding" (as seen in e.g. CellArea.java)
+			// does not make sense here.
 			backgroundImage = new BackgroundImageInfo(uriString,
 					imageStyle.getProperty(StyleConstants.STYLE_BACKGROUND_REPEAT), 0, 0, 0, 0, rl, module,
-					imageStyle.getProperty(StyleConstants.STYLE_BACKGROUND_IMAGE_TYPE));
-
+					imageStyle.getProperty(StyleConstants.STYLE_BACKGROUND_IMAGE_TYPE), dpi);
+			if (dpi != null) {
+				saveImageDpiOverride(backgroundImage.getImageData(), dpi);
+			}
 			if (backgroundImage.getSourceType().equalsIgnoreCase(CSSConstants.CSS_EMBED_VALUE)) {
 				uri = backgroundImage.getDataUrl();
 			}
@@ -3485,6 +3490,25 @@ public class HTMLReportEmitter extends ContentEmitterAdapter {
 			}
 		}
 	}
+
+	protected void saveImageDpiOverride(byte[] imageData, Integer dpi) {
+		if (dpi != null) {
+			saveImageDpiOverride(imageData, dpi, dpi);
+		}
+	}
+
+	@SuppressWarnings("unchecked")
+	protected void saveImageDpiOverride(byte[] imageData, int dpiX, int dpiY) {
+		Map<String, Object> appContext = report.getReportContext().getAppContext();
+		Map<byte[], int[]> dpiOverrides = (Map<byte[], int[]>) appContext.get("dpiOverrides");
+		if (dpiOverrides == null) {
+			dpiOverrides = new HashMap<byte[], int[]>();
+			appContext.put("dpiOverrides", dpiOverrides);
+		}
+		dpiOverrides.put(imageData, new int[] { dpiX, dpiY });
+
+	}
+
 }
 
 class IDGenerator {
--- a/engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPage.java
+++ b/engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPage.java
@@ -80,6 +80,8 @@ public class PDFPage extends AbstractPage {
 
 	protected PDFPageDevice pageDevice;
 
+	private Map<byte[], int[]> dpiOverrides;
+
 	/**
 	 * font size must greater than minimum font . if not,illegalArgumentException
 	 * will be thrown.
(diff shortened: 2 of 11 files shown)


speckyspooky:
Could you please add a test report with your image and pdf-examplse with the result before and after your change to compare it.

What was your result of the reportproperty on "report level" because there is an "Image DPI"-property.
I would think that the property not work in your case or you figured out other BIRT-handling that this wasn't an option for your.
(The defautl of that property is 96 dpi.)

hvbtup:
[background_image.zip](https://github.com/eclipse/birt/files/10899689/background_image.zip)

The ZIP contains the following files:
 - background_image.rptdesign: A test report with an imageDPI setting of 762.
 - gaerqualitaet.png: An image with the same DPI resolution in its metadata. The report contains a grid with a cell of 156mm x 109mm which is more or less exactly the size of the image at this resolution.
 - image-properties.png: Properties of above image as shown by IrfanView.
 - background_image-master.pdf: Output of the report with the master branch as of 1263d9e "Fixed SWT (#1203)"
 - background_image-fixed-by-report-dpi-setting.pdf: Output of the same report with this PR.
 - background_image_with_user_property.rptdesign: An alternative report, with a imageDPI setting of 96 and a UserProperty `imageDpi` set to 762 for the cell containing the background image.
 - background_image-fixed-by-user-property.pdf: Output of this report with this PR.

Direct observations:
- With the master branch, the image is enlarged, thus showing only a small part of the image.
- The `backgroundSizeHeight` and `backgroundSizeWidth` properties are ignored.
- The other 2 PDFs are rendered correctly, because the PR allows to read the DPI setting from a user property or from the imageDPI setting of the report.

What cannot be seen here:
- This issue is only with background images, not with image items.
- This used to work with BIRT 4.3.0 and Oracle Java 6/8, without specifying the `backgroundSizeHeight` and `backgroundSizeWidth` properties (and IIRC without setting the `imageDPI` property of the report.
- The test report is for a cell background only. Other locations in a report also support background images, including Master Pages. I did not test the behavior of these other locations.

**This PR is draft, because I think that it is just an ugly workaround. I think a proper solution would be to take the  `backgroundSizeHeight` and `backgroundSizeWidth` into account.**

speckyspooky:
Thanks for your example, good documentation.
And the first research result is that the original implementation implemented the properties backgroundSizeWidth & backgroundSizeHeight, but the properties not implemented thorugh the process.
(It si little bit like the diagonal properties, added but not finished.)

I will take a look what I can do to get access to both properties (1st target).
And your idea is good, I would check if we can change it to the "Image DPI"-property at report level.
Here is the same issue like size and width of the background property, added but not finished at all.

speckyspooky:
Short update, I invested lot of investigation of the image handling
and refactored/enhanced the background image handling of the designer, preview, pdf and pptx.

Attached one test case without specific user property. The base class will calculate the dpi by himself
and create the basic resolution. Also the both height & width properties are added to work.

The open tasks on my side currently are scripting & testing.

Here an example of the cell-background-image:

01 Screen:
2023-04-02 22_13_36-background_image_testcase01_cell_#1195 pdf - Adobe Acrobat Reader (64-bit)


02: PDF-document:
[background_image_testcase01_cell_#1195.pdf](https://github.com/eclipse/birt/files/11133253/background_image_testcase01_cell_.1195.pdf)

wimjongman:
Great stuff Thomas!! Does it also support auto-resize of images? For example, the aspect ratio gets warped in your PDF when the width or height changes. It would be nice if it would also shrink to keep the aspect ratio.

hvbtup:
@speckyspooky This looks very promising! Great work so far!
And of course this is much better than my attempt.

speckyspooky:
I checked the option of aouto-scaling depending of the container which use the backrgound-image. The topic is that the information of the rendered height & width only directly given it the properties explicit set. But in most cases the content like cell of table row includes dynamic data without fixed/explicit height (for example) and then the information isn't given.

The option of "getComputedStyle()" doesn't calculate these values and the "getStyle()" shows only the direct set properties.
It would be helpfull if we had some properties like a browser "clientHeight & clientWidth" then we had a chance for auto-scaling container based.

hvbtup:
Not sure if I understand your last comment.
The use cases for background images that I can think of are:

UC1: For me, personally, it absolutely suffices if I can explicitly specifiy background width and height in physical units like mm, pt or whatever.

UC2: Specifiy a width and height of 100%, for color gradient background.

UC3: Specify only width or only height in physical units, and keep the aspect ratio intact.

UC4: The background image should automatically fit (100%) in one dimension, keeping the aspect ration intact (leave free space as necessary in the other dimension)  (CSS: "background-size: contain").

UC5: This is already exotic IMHO: Similar to UC4, but cropping the image (CSS: "background-size: cover").

UC6: Do not specify anything. AFAIK the CSS specs say that this should be handled like UC4.

Can you explain which of the use-cases can and cannot be handled?

speckyspooky:
Yes, my explanation are very simple the topic is the eclipse-engine and also the aoutput emitters.
So yes, I know that contain & cover are valid CSS-properties but the images will be drawn with physical height and width at the end.
The CSS-properties in this case will work only on HTML but definitely not on eclipse and PDF side.
There the calculated values would be needed the browser provie here the properties "clientHeight" and "clientWidth",

**Use case list:**
  - UC1: ok
  
  - UC1b: ok, scipting of the values with JavaScript

  - UC2: percentage usage, I will take a look here (currently only with physical units)

  - UC3: not exactly this way, according to HTML if one dimension is set on the other is unset then "auto" will be used

  - UC4, UC5: unsupported (contain, cover)

  - UC6: ok, without any specific values the image will be calculated dependend on the size and dpi


The following elements are improved:
  - Simple Page Master
  - Table
  - Grid
  - Row
  - Cell
  - Container

The following emitters are improved:
  - eclipse-UI (no emitter)
  - HTML
  - PDF
  - PPTX

hvbtup:
> UC3: not exactly this way, according to HTML if one dimension is set on the other is unset then "auto" will be used

I looked at a [CSS demo page](https://www.w3schools.com/cssref/playdemo.php?filename=playcss_background-size) yesterday, and "auto" is not well defined. The say "original size", but the effect is that the browser keeps the aspect size.

Anyway, for me, UC1 ist most important, and I don't really care if the others work or not.
I'm looking forward to your PR.

With UC4, what I think is nice to have is the *effect* of "background-size: contain", not necessarily the CSS property itself.
A variation which is only slightly more inconvenient is if one could specify 100% for one dimension, and don't specify the other dimension. The difference is that - depending on the cell aspect ratio / image aspect ratio - the effect is either "background-size: contain" or "background-size: cover".

speckyspooky:
I have added the pull request #1256 which includes all changes.
I added in my comment some of my test-reports too based on the original report & layout.

hvbtup:
Closing this pull request in favor of #1256

Issue #996, opened 2022-06-10
Add slash in Grid


sanshigongA:
like Excel

wimjongman:
Thanks for reporting and welcome. Can you be a bit more specific?

wimjongman:
You can use any style, as long as you remain friendly and clear. 

It is not clear what you want. 

What do you mean by "add slash to grid like in excel"

speckyspooky:
Hello Together,
it seems for me that sanshigongA means the diagonal line on a table/grid cell, e.g.:
![slash-diagonal-line-at-cell](https://user-images.githubusercontent.com/41593722/173225538-c08d6479-c4e6-4336-9e25-8d6f20ad57d1.png)

This is an option in Excel on border level. It would be a good improvement.
I figured out "Stack Overflow" some solutions for HTML (CSS3 and svg) - [https://stackoverflow.com/questions/28425574/slanted-diagonal-line-in-html-or-css](url) (the example image is from there).

The improvement sounds simple but it seems to me a challange on BIRT-engine due to CSS3 / SVG-handling on emitters (Excel, PDF, ...) - my favorit would be CSS3-solution.

Have we perhaps an expert in our group on it?

hvbtup:
It seems like MS Word and MS Excel support this. It is a border property there.

I think there isn't a CSS standard for this. 
The accepted solution in https://stackoverflow.com/questions/28425574/slanted-diagonal-line-in-html-or-css with background-image and linear-gradient is working, but I think we cannot use it as-is, at there is a conflict with the existing background-image property.

So I think this cannot be support (reasonably) for the HTML emitter. 

IMHO, for a solution we could add custom CSS border-properties like `border-slash` and `border-backslash` (?) just similar to `border-top`, eg `border-slash-width`etc.
These could fully be supported by PDF, Excel and Word emitters.
For the HTML emitter, there would be some restrictions, the most important being that it doesn't work in conjunction with `background-image`.
Of course, we'd need to add indivual code to the emitters to support this.

So, there'd be three tasks:
1) Add a specification to the ROM model
2) Document the restrictions
3) Add emitter-specific implementations

I think it would be doable for someone with enough time.

sanshigongA:
[...]

PR #1224, opened 2023-03-02
Enhancement to add new options of diagonals and antidiagonals #996

speckyspooky:
**Enhancement of diagonal lines - diagonal & antidigonal line**

ROM	added antidiagonal, properties available in the designer at page advanced properties
		
PDF		added antidiagonal with color
		diagonal line implementation: PDF diagonal border lines
		FIEXED: fixed issue with cell background & diagonal line at the same time

HTML	added digonal and antidiagonal lines with color
		diagonal line implementation: png image
		FIEXED: fixed issue with cell background & diagonal line at the same time

PPTX	added antidiagonal with color
		diagonal line implementation: PPTX diagonal border lines

DOCX	added diagonal and antidiagonal with color
		diagonal line implementation: DOCX diagonal border lines

XLSX	added diagonal and antidiagonal with color
		diagonal line implementation: XLSX diagonal border lines
		new implementation of diagonal methode because the POI apache api doesn't support directly digonal lines

DOC		added diagonal and antidiagonal with color
		diagonal line implementation: DOC diagonal border lines

PPT		proritary Actuate format, none Microsoft PPT format
		added diagonal and antidiagonal
		diagonal line implementation: DOC diagonal border lines

XLS		unsupported
		
ODP		added antidiagonal (no color support)
		diagonal line implementation: ODP diagonal draw line
		
ODS		unsupported

ODT		unsupported

**Documentation:**

1. Class changes: 
[996-Change-Class-Changes.xlsx](https://github.com/eclipse/birt/files/10876157/996-Change-Class-Changes.xlsx)

2. Test reports:
[diagonal_test_case_01_default.zip](https://github.com/eclipse/birt/files/10876164/diagonal_test_case_01_default.zip)

3. Test results (output):

[test_documents.zip](https://github.com/eclipse/birt/files/10876163/test_documents.zip)


hvbtup:
Please use one PR for background images and a second PR for the diagonals.

speckyspooky:
The pull request #1224 includes only the diagonal changes.
My failur was at the beginning because I used my background-branch again. But the 62 files are only the new topic without any mix.

speckyspooky:
At this moment I see that "Maven Build" finished with error like this night.
And I cannot see any reason what the error cause. Could somebody please assist to figure out the problem.

Built-Failed-without-any-hint

wimjongman:
> Built-Failed-without-any-hint

Hey Thomas, there is a hint. Just scroll down to the last page of the build log. Our build is crazy. It builds for over one hour.

The problem is below, and I have restarted the workflow.

`
Error:  Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:2.10:copy (download-jars) on project org.eclipse.birt.build: Unable to resolve artifact.: Could not transfer artifact dom4j:dom4j:jar:1.6.1 from/to central (https://repo.maven.apache.org/maven2): Connection timed out (Read failed)`

speckyspooky:
Thanks Wim for your restart, but something seems to be strange because the build failed again due to the same issue. 
I checked the maven repo and can find it witj the path:
https://repo.maven.apache.org/maven2/dom4j/dom4j/1.6.1/

I'm not a maven build expert, the only think I see is that the error message have an additional ":jar:" at detail message "dom4j:dom4j:jar:1.6.1", would it be possible that the repo has a changed path (I think this would be very strange) or is there an to do on our path/dependency...?

wimjongman:
I'm preparing for the 4.14 release. There I also updated the dependency version numbers, which might automagically fix this. 

@claesrosell are we caching the workspace? Could it be a problem in the maven cache?

Mailaender:
Try updating Tycho and downgrading Maven. https://github.com/eclipse/chemclipse/pull/1331

wimjongman:
I have downgraded maven for now in master. I am currently working on the 4.14 releng and as a part of this I will upgrade to the latest and the greatest dependencies.

I think the build will use whatever is in master. If not then please update your branch with the latest from master.

wimjongman:
see https://github.com/eclipse/birt/commit/69c3ca6adaf404304c8112a30d21713f88a2715f

wimjongman:
Deleted the previous comment. The correct commit is: https://github.com/eclipse/birt/commit/fa05ecd7b586e170a3525d4b47bc5d65e480ca71

wimjongman:
We have a master build running here: https://github.com/eclipse/birt/actions/runs/4372471029/jobs/7649432283

speckyspooky:
@wimjongman thanks for your hint. I updated my dev branch from your master.
The good new on it the build step runs sucessfully but therefor stopped the "Model Test" with many errors.
(But I didn't understand the issues because I haven't changed the something and the errors have nothing to do with my changes...)

speckyspooky:
@wimjongman Have you perhaps an idea to solve the issues with the failed "model tests"? The build of this week of the latest pull request was sucessful, therefore I didn't understand the error messgaes because without my changes there is no difference from master-branch.

wimjongman:
I have looked in the logs, and there is one test failure:

`GridItemParseTest.testParser:175 expected:<0> but was:<1>`

It looks related to the stuff that you are doing. You can run this test locally and see why it fails.

speckyspooky:
@wimjongman Thanks and I will take a look.
But the test framing doesn't work fine beause there are many issue with dependencies.
I have refreshed the local "Maven Build" of model tests but the dependency issues wasn't resolved and so currently the test frame is not runnable locally on my side.

This is also the reason why I haven't use the test framing active due to the big dependency issues on my local side.
Some hints to solve the issues so that I could use them would be helpful.

*Dependency issue*
test-frame-set - org eclipse birt report model tests_test_org_eclipse_birt_report_model_pars

*Output log running*
GridItemParseTest-test-result-based-on-dependencies-issues

wimjongman:
You have to set the target platform with the birt.target file and run with the model launch configuration. They are in the birt.target plugin. Please take a look at the README.MD. All is described in a video: https://www.youtube.com/watch?v=FqfrG2I0AIw

speckyspooky:
Your first hint was helpfull because the test-class used a wrong setting due to my changes - test is fixed.
Now I got new issues that old test reports based on BIRT 2.1.1 not valid XML files.

It is very confusing the the problems. I will take a look into your link and README.MD and hope this will help me for local testing.
I hope it would be ok for our if I ask you again if I didn't got a positive result.

wimjongman:
> I hope it would be ok for our if I ask you again if I didn't got a positive result.

Sure!

wimjongman:
> I hope it would be ok for our if I ask you again if I didn't got a positive result.

You can also post a discussion question so that others can benefit as well.

speckyspooky:
The issues are solved. @wimjongman I used your instructions since I started but I saw on your video the small hint of unit tests and dependencies. The reason of my issue was thet the dependend PlugIns wasn't marked therefore the data are missing.
I added these an now it is running.

The issues on the test clases based on 2 changed property values based on my change
and one issue with the test class because in some cases the report design files for testing couldn't be loaded correctly.
In that case the object ws null an tried to convert like xml which case the message invalid xml file (which was correct).

Currently the build finished in full also with the test cases.

wimjongman:
Great!

> one issue with the test class because in some cases the report design files for testing couldn't be loaded correctly.

That is strange. why would they not be loaded correctly? I see that you have caught the issue. Does this hide the test failure?

speckyspooky:
Yes, the new check of the null object avoid the issue. The message "invalid xml file" is correct if a null-object is casted to xml.
The first issue was based on a changed property falue for the diagonal default.

speckyspooky:
@wimjongman Hello Wim do you see a way to review the changes of the new feature because it has an impact of further changes based on the implementation dependencies (no library dependencies). Thanks in advance

wimjongman (approved):
Sorry for the delay, Thomas! It looks good to me.

speckyspooky:
Thanks for your verification and merge!

hvbtup:
@speckyspooky 
In our customized `tup_main` branch of our fork https://github.com/triestram-partner/birt I tried to merge your changes yesterday.
Merging was difficult, because the branch uses Jim Talbut's latest enhancements to the Spudsoft Excel emitter, so the source code for this emitter looks quite different from that of the master branch.

*I noticed two issues:*

1) Diagonal/Antidiagonal lines do not work for Excel output in this branch. This is certainly because I made a mistake when manually resolving the many conflicts. So this is my problem, not yours. Anyway, since you know the topic much better, perhaps you can spend 10 minutes, maybe you spot the error quickly?

2) The emitter code in this fork seems to require that `PerfectHash.java` works correctly. I noticed that you changed the file, probably manually, and it doesn't work correctly for the "new" options.

*Regarding 2), I took the time to fix this in my branch;
and I recommend to merge these fixed files into the master branch (see below).*

Please take a look at the following files in my `tup_main` branch (https://github.com/triestram-partner/birt/commit/11e5af9d73b603b7348ef67fe13121acf8418fc9) and compare them to the master branch:

In directory engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/css/engine:
* PerfectHash.java
* StyleConstants.java
* token.cpp
* token.gperf (the comment at the beginning is wrong, please ignore it)

The changes cost my several hours of work yesterday: I noticed in the debugger that for the new options, an index -1 was returned, so this resulted in an ArrayIndexOutOfBounds exception when trying to call the Spudsoft emitter. So it was clear that either the PerfectHash class had to be fixed or the whole PerfectHash thing should be removed from the code.

Since the source code contained many magic constants, it was clear to me that the code was generated somehow.
Searching in the internet, I quickly found that the `gperf` command line tool was involved somehow.

I was disappointed that even newer versions of the tool can't generate Java source directly, but soon I noticed the files token.* in the directory.

I then recreated `token.gperf` by extracting and sorting the wordlist from `PerfectHash.java`, and with the help of the hints I managed to let `gperf` create a new version of `token.cpp` which had the exact same structure as the existing one (after fixing UTF-16 encoding and windows line ending).

So I copy-pasted the constants, wordlist etc. from `token.cpp` into `PerfectHash.java`.

To my disappointment, the Excel emitter crashed instantly, it was totally messing up the properties.

With further debugging, I noticed that the integer values inside `StyleConstants.java` have to match the index values inside the wordlist of `PerfectHash.java`. It was rather tedious work to fix this file.

Maybe in the future it would make sense to create a little script which magically fixes the numbers in `StyleConstants.java` based on `PerfectHash.java` and assures that the numbers are distinct...

speckyspooky:
@hvbtup 
Thanks for your information. I'm little bit shocked to see that you have such an impact on your side.
Some minutes ago I tested again some of my master test cases with the latests build with designer, PDF & XLSX.

The first result seems to me ok because I got no issue like you with the merge - here the result:
[diagonal_test_case_02_xlsx.zip](https://github.com/eclipse/birt/files/11219358/diagonal_test_case_02_xlsx.zip)

You have written that you use another version of the SpudsoftEmitter like the BIRT-master-branch.
But I can only work what the main branch has included so I doesn't know the final impact of your newer version to my change.

Therefore I don't know whether I can use your changes because you have another BIRT-version based on your branch.

Can you please egt a link to your commited file changes on your branch because in your previous link I didn't figured out the commit of your changes.
And would you be so kind and give me an instruction (detailed) how you genereated the new values because yes,
I changed PerfectHash together with StykleConstants, but the token.cpp is new for me and I don't know the usage of it.

But at the end it seems to me that I cannot copy your changed files directly because the dependencies of your branch in difference to the master. Your branch is many commits ahead of the BIRT-master-branch.

hvbtup:
Don't be shocked.
I'm used to having difficulties with merging changes from the master branch into the tup_main branch sometimes.
I wish I knew more git-fu to make merging easier.
But I just live with it.
The enhancements of the branch in comparison to the master branch are worth it for our company.

You can in fact copy the four files I mentioned verbatim from the tup_main branch. They are not depending on any of the other changes in the branch.

> And would you be so kind and give me an instruction (detailed) how you genereated the new values

I tried to do this more or less in my previous comment.

The basic idea is like this:

Input:
You have a list of property names.
It is not clear to me what the source of the list should be, though.
I took the existing wordlist from `PerfectHash.java`, modified it with the help of a good editor, sorted it, and saved it to token.gperf, then added the (misleading) comment to the top of the file.

So `token.gperf` is the input file (ASCII encoding, Linux line ending required).

Then you need to have a working version of the GNU `gperf` utility.
In my case (MS Windows), I found one in cygwin.

On the command line, inside the directory containing `token.gperf`, call

    gperf -d -7 -l -D -L C++ token.gperf > token.cpp

So, `token.cpp` is an intermediate file.

The next step is to copy texts from `token.cpp` into `PerfectHash.java`.
What has to be copied should be clear, as the files have a very similar structure.
I reckon that for the original PerfectHash.java someone translated the C++ code more or less 1:1 into Java.

The last (and tedious) step is to make the integer values in `StyleConstans.java` match the indexes of the corresponding words in the wordlist array in `PerfectHash.java`. Fortunately, that list contains the indexes as comments.

speckyspooky:
Thanks for your information.
So I will take a look into your files and will create an additional change to add these changes to a new branche
based on the BIRT-master.

And thanks again for your details of PerfectHash. I took a deeper look into your notes and files so I can do it at the next time.
I copied directly your steps for the next time (perhaps I would send you a question if I would have additional questions on it if I use it in future).

hvbtup:
> I copied directly your steps for the next time 

Probably it makes sense to include the instructions as a `build-instructions.md` file inside the source directory,
and to include inside the two *.java files comment with a link to the instructions and a fat notice that the files are very closely related and should only be modified following the instructions.

PR #1232, opened 2023-03-09
Removed old Maven build log

Mailaender:
Looks like this was committed accidentally in https://github.com/eclipse/birt/pull/196.


wimjongman:
Thanks!

PR #1233, opened 2023-03-13
Fix Birt Viewer Frameset inside a Iframe

vpease:
I wanted to run the frameset report viewer inside an iFrame but the error reported was about trying to set Connection header by an ajax call. According to many sources  that is not allowed for ajax calls, but the error is not reported when the viewer is run in a complete window.

After a lot of research, i've just tried the simplest: modify prototype.js in order to avoid sending the Connection header. 

The viewer is working now inside and out of an iframe with no collaterals.


wimjongman:
Hey,  Victor! Thanks for the PR.

vpease:
Results of my testing:
- The fix will run with no changes only in Firefox 
- All other navigators will report "The viewing session is not available or has expired"
  - To solve this, the host page must be in the same context as the birt viewer. That is the page with the IFRAME must be published in the same Tomcat context.  The origin of this is because the ajax client uses several cookies with origin, when using it inside a frame, those cookies are not accesible.

Without the change, the error "The viewing session is not available or has expired" is generated in all navigators.
Considering Firefox as the only way to run it may be valid for some cases.

Necessity: 
Birt viewer (/frameset) has navigation control and pagination, the only way to deal with a lot of records. 

Apply to:
- JSP viewer
- Ajax viewer

vpease:
> The question is how long will FF support it if it is the only one?

Doing some more research, the problem for the rest of navigators is about restrictions on iframe to read their own cookies 

Firefox display this warning: 
Cookie “JSESSIONID” does not have a proper “SameSite” attribute value. Soon, cookies without the “SameSite” attribute or with an invalid value will be treated as “Lax”. This means that the cookie will no longer be sent in third-party contexts. If your application depends on this cookie being available in such contexts, please add the “SameSite=None“ attribute to it. To know more about the “SameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite

So for now the workaround willl be:
 - Deploy the host page in the same context as the birt viewer. 
That's working now for all navigators


Alternatives:  
- implement in servlet /frameset  https://developers.google.com/search/blog/2020/01/get-ready-for-new-samesitenone-secure#:~:text=A%20New%20Model%20for%20Cookie%20Security%20and%20Transparency,-Today%2C%20if%20a&text=Developers%20must%20use%20a%20new,be%20accessed%20over%20HTTPS%20connections.
- Or a proxy hack https://stackoverflow.com/questions/45094712/iframe-not-reading-cookies-in-chrome

Note: The fix is still valid considering this https://stackoverflow.com/questions/7210507/ajax-post-error-refused-to-set-unsafe-header-connection/7210840#7210840

PR #1264, opened 2023-04-13
Renewing of the PerfectHash values due to new properties of diagonal and antidiagonal #1224

speckyspooky:
The added changes are the renew of the new designer properties of PerfectHash caused on #1224


speckyspooky:
I have added a README.md file. It is my first try with such kind of README-makefile.
Therefore I don't know exactly if I have reached all your requirements.

But the good thing is I installed cygwin and tested the steps with success(!)

hvbtup:
Thanks for the README.md file. I made some minor changes and added an example at the end.

Is the `BIRT Model Test (Engine).launch` file related to the change in any way?
If not, can you please remove it? It causes a conflict.

hvbtup (changes requested):
I think the run configuration (which is probably not related to the changes) should be restored.

hvbtup (approved):

PR #1296, opened 2023-05-29
Provide shared launchers for all tests

merks:
Also provide a group launcher that launches all the tests

Improve the Oomph set to add all the launchers as favorites.

Add to .gitinore to avoid so many test result artifacts dirtying the Git repository state.


merks:
This creates launchers for every test as well as a launcher for launching all tests:

image

The setup increases the launch history size to 50 so that one can review the test runs after running them all

image

Only the SpudSoft tests fail.

merks:
@wimjongman (or anyone else)

Please merge these changes so that we can all more-easily run tests locally when making changes to BIRT.

Issue #1286, opened 2023-05-21
Improve the release engineering


merks:
### Tasks
- [x]  Make Eclipse Marketplace work with the new location
- [X]  Update website download locations
- [X]  Update project download locations

### Description
The setup should produce a usable workspace amenable to contribution. The target platform should be modernized to use newer dependencies; ideally it should be easier to maintain it and to update it to newer dependencies in the future.  It should be easier (possible) to do local Tycho builds.  It should be easier to produce nightly, milestone, and release builds.  There appear to be too many products (*.product) and too many update sites (category.xml).  It's not clear why those are all needed and who all consumes those.  Simplification and reduction of these would help streamline the releng process. 

Perhaps the EPP Eclipse IDE for Java and Report Developers could be revived, but the build results are of a quality not consumable by SimRel:

https://download.eclipse.org/oomph/archive/reports-extra/birt-snapshot/download.eclipse.org/birt/update-site/snapshot/index.html

merks:
I've been trying to get the tests to run locally, with some success but also some failure, but now I see the build uses `-Dmaven.test.skip=true` which is most disconcerting!

wimjongman:
Do you mean the main build? I can imagine that because that one is only to publish the result. All PR checks are running with tests. E.g.:

https://github.com/eclipse/birt/actions/runs/4773353790/jobs/8486375628?pr=1271

merks:
My branch that I was working on, unknown to me, was badly out of date.   😢   So I'm reworking the changes to get back to the point I was at yesterday.   And there are many interrupts today we shall see how far I get today...

wimjongman:
Thanks, Ed. No rush. 

@eclipse/technology-birt-committers Please read this thread. We are going to improve our releng and with that our overall ability to execute.

merks:
@wimjongman 

To really use the latest of everything, e.g., Platform, GEF and so on, you'd need to use Java 17 for some things, e.g., for the JustJ JRE used for the products.  Does anything speak against switching to Java 17 which is the latest LTS version?

wimjongman:
We can go to Java 17.

wimjongman:
We have been moved to our own organisation:

https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/3185

merks:
FYI, I will wait for this IP review to complete

https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/8506

[...]

PR #1303, opened 2023-06-04
Update pom, feature, and bundle versions to 4.14.0

merks:
https://github.com/eclipse/birt/issues/1286


merks:
@wimjongman 

I assume that all these things should be incremented for 4.14.0 for the next release given that 4.13.0 was the previous release.  If that is correct, then this is ready to be merged.

merks:
@speckyspooky 

It will be easier to help with the emitter issue tomorrow if this is merged so I can work off master.

ruspl-afed (approved):
LGTM. I've looked through them all.

Issue #1322, opened 2023-06-09
High memory usage in SpudSoft Xlsx Emitter


speckyspooky:
### Discussed in https://github.com/eclipse-birt/birt/discussions/1318

<div type='discussions-op-text'>

<sup>Originally posted by **redkark** June  8, 2023</sup>
I observe high memory usage when using Spudsoft Emitter while generate XLSX output for larger data volumes. I am using the BIRT runtime v4.12 in my Java code to generate the reports. I looked at certain posts where it was suggested to use the ExcelEmitter.Extract in the render options to instruct the spudsoft emitter to use the "streaming" SXSSFWorkbook from Apache POI. However it did not have desired effect. Upon looking at the code for XslxEmitter, I noticed it uses the XSSFWorkbook with no option provided to switch to the SXSSFWorkbook. Would like to know if it is possible to configure spudsoft emitter to make use of the SXSSSFWorkbook, via the render options or other mechanisms?</div>

speckyspooky:
To have the miniumum of impact to the main project, we could check whether an additonal ExcelEmitter-option would be possible like "ExcelEmitter.StreamingXlsx" as boolean to setup the xlsx-streaming option.

The important thing is that the usage of XSSF and SXSSF isn't the same so the default would be the current handling with XSSF without streaming. The streaming must beactivated with ExcelEmitter.StreamingXlsx = "true".

Apache POI reference:
https://poi.apache.org/components/spreadsheet/

Apache POI details:
https://poi.apache.org/components/spreadsheet/how-to.html#sxssf

speckyspooky:
With the latest built of 4.14-snapshot the new emitter-option is available.

Issue #1307, opened 2023-06-05
Improvement of the BIRT-viewer, small styling renew


speckyspooky:
Hello together,

I started a very small sensitive improvement of the BIRT-viewer.
Since I started the BIRT usage 2017 I had the same preview with a little bit old fashioned style.

On my one viewer I had updated the coloring, the main function icons and the parameter icons.
Now I wants to push these changes to the BIRT-master.

But before I start a PR here you can see the sensitive styling changes.
There are only changes on coloring and 8 icons (no master renew).

The main change is that I removed the beige-light-brown through a blue color and a very light gray of the dialog.
I know that the most use and own integration and styling can be a very sensitive topic,
but I think it would be good to have small changes after a so long time.

Please let me know your comments - The icons created on my own (no foreign icons).

### Styling compare - Left side "current version"  / Right side "renewed version"

![birt-preview-sensitive-renew](https://github.com/eclipse-birt/birt/assets/41593722/90919d62-cb81-4561-839c-3ca554658d09)

hvbtup:
I like the new look, but OTOH we don't use the viewer in our application, we only see the parameter entry UI during development, when we're running the report from the All-In-One IDE.

speckyspooky:
Same on my side the viewer isn't part my application.
The idea is to get a little bit smarter viewer for the report developer with a little bit modern feeling.

speckyspooky:
The new layout is available with the latest snapshot of BIRT-4-14.

PR #1323, opened 2023-06-09
Excel streaming xlsx option with new configuration UserProperty "ExcelEmitter.StreamingXlsx" #1322

speckyspooky:
The new user property give the option to define the kind of workbook processing which should be used.
The new option: ExcelEmitter.StreamingXlsx is a boolean and has ethe options true, false (default: false).
The values are:
- true: activate the SXSSF usage to stream the data internally
- false: standard handling based on dom object

The difference of memory usage is given but the handling isn't the same in all steps.
My test with 1Mio records:
XSSF: 130s
SXSSF: 82s

Attached my test case with 1Mio records:

[Data7602DescendingYearOrder-shrinked-1Mio.zip](https://github.com/eclipse-birt/birt/files/11710418/Data7602DescendingYearOrder-shrinked-1Mio.zip)


speckyspooky:
@wimjongman 
Sorry, this PR includes the changes of #1307 again the changes includes only 3 files.
I don't know why this happens because I created a new branch and eclupse showed only the 3 changed files:

2023-06-10 00_18_43-Excel streaming xlsx option with new configuration UserProperty _ExcelEmitter St

merks:
@speckyspooky 

When you created the new branch you created it when the checked out branch at the time of creation had commits on it other than those on the remote master of the main repository.  I don't think you can commit the changes here without including both the commits...

Generally, if you want to keep things separate, you should check out master, and pull before you create a new branch. 

You could "fix" this by renaming the branch used for this PR, checking out master, pulling, creating a new branch with this same name as you used before, in the history view, cherry pick this commit onto that branch, and finally force push that branch to your fork which will update this PR...

speckyspooky:
Yes, your expanding make sense.
I created the new branch directly.

The good thing user both changes should be part of 4.14. and there are no further changes.

merks:
@speckyspooky 

In my opinion the entire set of changes looks fine and good!  I think they should be merged, and that might be a good idea to do sooner rather than later because I also have a commit pending which I don't believe will conflict, but you never know...

If @wimjongman has suggestions, they can always be applied after the fact...

speckyspooky:
I will merge the changes this weekend.
The risk is very small and I want to go ahead with viewer changes.

wimjongman:
Go ahead. I'm not at my computer right now to approve.

Best regards,

Wim Jongman
________________________________

speckyspooky:
@merks 
It seems to me that I have an issue with the new build process.
The build seems to be done but there is an error documented on Jenkins. Could it be that I need additona grants/privileges?

Could you please assist what I can do to fix it or get the additional grants.

2023-06-10 15_07_51-Eclipse Business Intelligence and Reporting Tools (BIRT) - build Scan Repository

merks:
There seemed to be a glitch (infrastructure problem) in a PR build as well and I just restarted the build manually and now that seems to be doing it's thing:

https://ci.eclipse.org/birt/job/build/job/PR-1325/2/console

I kicked off master again manually and it seems to be waiting for an executor:

https://ci.eclipse.org/birt/job/build/job/master/12/console

Which makes sense because there are two executors running already and that's the max:

image

@wimjongman 

We could disable the old builds I've underlined here and eventually delete them:

image

The birt-master is especially annoying because it kicks in automatically at the same time as "build" and by using an executor blocks other jobs, e.g., PR builds.   I.e., that is superseded by which does the publishing to the new locations:

https://ci.eclipse.org/birt/job/build/job/master/

merks:
I've disabled all those builds and I've abort the birt-master build.  The https://ci.eclipse.org/birt/job/birt-master/306/console job says it's aborted:
```
Build was aborted
Aborted by [Ed Merks](https://ci.eclipse.org/birt/user/ed.merks@gmail.com)
Skipped archiving because build is not successful
Recording test results
Finished: ABORTED
```
But it's not cleared off of the main page yet:

image

Where the build -> master is still showing in the queue.

Unfortunately the ci instance is not always as well behaved as it ought to be.  I'll check the status again in while...

merks:
I merged my PR because the GitHub action completed and verified it, so that job is killed and now master is building:

https://ci.eclipse.org/birt/job/build/job/master/13/console

which will include both the commit for this PR and the commit for my PR...

speckyspooky:
Ok, I have seen the cancelled jobs from your side and also your new job of your PR.
The original build of my PR was successful but due to the Jenkins-jobs which caused the current situation it wasn't finished at all.

So I cross my fingures that your new jobs works well :o)

merks:
It finished and published a new build:

speckyspooky:
Many thanks to your side!

Issue #1329, opened 2023-06-11
Prepare the index.jsp for 4.14 and renew the styling


speckyspooky:
The current version of index.jsp-side of the BIRT-viewer should be configured for BIRT 4.14 (support for #1286).
In addition to it the style of the old page will be a little bit modernized.

The new index.jsp will be:

![index jsp](https://github.com/eclipse-birt/birt/assets/41593722/7d1580b7-6efd-4666-a0f1-dd8361042179)

The new example (Hello world reports) will be:

Example report 01:

![report01](https://github.com/eclipse-birt/birt/assets/41593722/a6176ac0-c4d4-49ae-8164-ed8429c19ec8)

Example report 02:

![report02](https://github.com/eclipse-birt/birt/assets/41593722/ef37e7d5-654f-4867-9251-3622c8a5dbed)

wimjongman:
Wunderbar!

PR #1330, opened 2023-06-11
Prepare the index.jsp for 4.14 and renew the styling #1329

speckyspooky:
The PR includes the change of the version number for BIRT update 4.14.0,
the reorganization/renewing of the index.jsp page and also the 2 new example reports.


speckyspooky:
@wimjongman, @merks
I have an issue with the OSGI-version which couldn't be finished.
But the log files means only that it is failed. I doesn't see any valid reason of it in the log files.

Do you have an idea was going on here
and what my mistake is?

wimjongman:
Cheers! I am in Prague for a conference, so I am not very responsive but I try to keep up.

PR #1341, opened 2023-06-23
Include org.eclipse.birt.chart.osgi.runtime in birt-runtime-osgi

merks:
This ensures, for one thing, that javax.xml.soap is available.

https://github.com/eclipse-birt/birt/issues/1340


merks:
Oops, I forgot to merge the target platform update.

Issue #1343, opened 2023-06-24
BIRT WebViewer shows some hints at the browser console


speckyspooky:
The current HTML-usage of the web-viewer shows some HTML-hints to imporve the htlm page.

![check-console-message](https://github.com/eclipse-birt/birt/assets/41593722/5c77df5e-a00a-4930-8d96-1d5e817c1e67)

speckyspooky:
Improvement is part of PR #1329

speckyspooky:
Done with built of birt-runtime-4.14.0-202306261418.

PR #1344, opened 2023-06-24
Improve viewer html #1343

speckyspooky:
The blank HTML-page got a body and the processing.jsp got a standard HTML-root-tag


speckyspooky:
Temporary Jenkins issue, the changed is included in #1345

PR #1345, opened 2023-06-24
Added new images and fixed typo #1329

speckyspooky:
Added new images and fixed typo #1329


speckyspooky:
The request includes the fixes for:

- #1346  Renew web viewer parameter dialog & optional flag of textboxes
- #1343  Fix of BIRT WebViewer browser hints
- #1329 improved images of index.jsp

speckyspooky:
@merks 
If you have questions to my changes let me know.

merks:
The target is generated so whatever you change manually will be overwritten the next time the setup resolves the target platform. It’s also minimized to remove implicit transitive requirements.  Let’s not go back to manually maintaining it until we hit a brick wall again. 

What is the rationale for the target changes?

speckyspooky:
I'm fine with the automated re-writting of the target.

The change was done through eclipse automatically and additional with it there appears some errors this afternoon on Jenkins so I thought the new target will fix the issue. And the PR was buildable.

But with your note I understand that the both topics have no dependency and I confirm the auto-generation is better.

speckyspooky:
@merks 
Could you please let me knwo during the day if I can go ahead with the merge. Thanks!

merks:
Notice the "Update branch" drop-down says cannot rebase due to conflicts.  I think that's because you modified the *.target which I already fixed in a separate PR.  So my sense is that something will be overwritten by this...

I suppose I can regenerate it and update it again if necessary...

speckyspooky:
Thanks for your check, so I will later merge it and test the built at all.

Issue #1340, opened 2023-06-22
BIRT 4.14.0 OSGI version seems to be broken


speckyspooky:
Based on the discussion #1331 I tested the latest BIRT 4.14.0 builts designer and the OSGI-version.

And it seems to me that the latest built are brooken because the birt-service couldn't be loaded on Tomcat.

Testing - Windows System:

- Tomcat 9.0.52, Java 17.0.5+8 Eclipse Adoptium
- Tomcat 9.0.76, Java 17.0.5+8 Eclipse Adoptium

Both installation was tested with the following birt versions:
- birt-runtime-osgi-4.13.0-20230302: service could be started and viewer example is available
- birt-runtime-osgi-4.14.0-20230604: service could be started and viewer example is available
- birt-runtime-osgi-4.14.0-20230610: service start crashed
- birt-runtime-osgi-4.14.0-20230612: service start crashed

Attached the both Tomcat error logs of 9.0.52 (same error on 9.0.76)
[tomcat-9.0.52-stderr.birt-osgi-4.14.0-20230610.zip](https://github.com/eclipse-birt/birt/files/11839931/tomcat-9.0.52-stderr.birt-osgi-4.14.0-20230610.zip)
[tomcat-9.0.52-stderr.birt-osgi-4.14.0-20230612.zip](https://github.com/eclipse-birt/birt/files/11839932/tomcat-9.0.52-stderr.birt-osgi-4.14.0-20230612.zip)

What I not understand on the eror log is the origin, is it an Tomcat issue or may be a missing dependency topic of BIRT 4.14.0 (because 4.13.0 is running)
Main error:
	Caused by: java.lang.NoClassDefFoundError: javax/xml/soap/SOAPException
	Caused by: java.lang.ClassNotFoundException: javax.xml.soap.SOAPException

![no-class-def-found-error](https://github.com/eclipse-birt/birt/assets/41593722/678933a4-369e-4fbf-9060-e08ff5447ba0)

On my research I found the following article which is very old and at the end, have we here a task for the new built process to add missing dependencies or is it realy a Tomcat configuration issue.

https://stackoverflow.com/questions/48626824/java-lang-noclassdeffounderror-javax-xml-soap-soapexception

speckyspooky:
@merks 
As an specialist of the built process & dependencies, could you please take a look here.
Perhaps you have an idea to solve the issue with the built process optimization (#1286) when it is a dependency topic.

merks:
[...]

PR #1348, opened 2023-06-24
Required dependencies of osgi-runtime (report generation) #1340

speckyspooky:
Added missing dependencies of osgi-runtime


merks:
I have the following working locally and comparing the two lib folders:

image

The soap mystery is more clear now because formerly that came from the saaj.jar.  Also javax.xm.bind is missing and should be replaced by jakarta.xml.bind, which export javax packages, unless it was never relevant, but best not to make assumptions.

I really don't want to pull things from the designer product.  Better that we pull everything from the target folder of the project itself.

I have a pull request coming and I thing it's better we abandon this one...

merks:
I think this fix is better and most probably more complete:

https://github.com/eclipse-birt/birt/pull/1349

This fix pulls everything from the target folder of the project itself rather than pulling things from other products in other projects.

speckyspooky:
Ok, I'm with you.

My PR will finished in some minutes and I will not push them to the master.
I will wait your improved version so that I can do my retest with the latest version tomorrow.

speckyspooky:
I close the PR because the #1349 "Copy more content into the lib folder of birt-runtime-osgi"
implements the optimized and better version for the dependencies.

PR #1349, opened 2023-06-24
Copy more content into the lib folder of birt-runtime-osgi

merks:
I.e., jakarta.xml.bind, org.apache.commons.logging, and org.apache.commons.discovery/*.jar

https://github.com/eclipse-birt/birt/issues/1340

### DIFF
--- a/build/birt-packages/birt-runtime-osgi/build.xml
+++ b/build/birt-packages/birt-runtime-osgi/build.xml
@@ -71,7 +71,10 @@
     </copy>
     <copy todir="${VIEWER_DIR}/WEB-INF/lib">
       <fileset dir="${ENGINE_DIR}/platform/plugins">
+        <include name="jakarta.xml.bind_*.jar"/>
         <include name="javax.wsdl_*.jar"/>
+        <include name="org.apache.commons.logging_*.jar"/>
+        <include name="org.apache.commons.discovery_*/lib/*.jar"/>
       </fileset>
       <fileset dir="${REPOSITORY_DIR}/plugins">
         <include name="javax.xml.soap_*.jar"/>


merks:
This should produce a lib folder with the same/analogous contents as the 4.13 lib folder:

image

It assembles content only from the product and repository folders in the project's target itself...

speckyspooky:
@merks 
Looks good, so if it is working I would cancel my current PR of the dependencies.

merks:
I have a good feeling when all the jars line up that we will have it right this time.  The ant script still feels fragile and adhoc though but at least we understand the potential source of problems and how to manually assess the content by comparing to a previous version.  I wonder if any of the jars are unnecessary…

speckyspooky:
I tested the new osgi built and the service is running and also the report examples could be generated.
So currently I would say we have all dependencies.

merks:
Woo hoo!  🎉 

I'm also working locally on running the tests for the runtimes, including for the OSGi runtime, though the later feels like a hack because the tester does not really launch it as an OSGi runtime...

wimjongman:
Well done guys. I am following all your work and it is just wonderful. I envy your skills. I will take the runtimes for a spin as well when the build is done.

wimjongman:
> because the tester does not really launch it as an OSGi runtime...

I think it does. If you follow Platform.startup:

https://github.com/eclipse-birt/birt/blob/b496c6b0489e3ae37fa39d792e97cadd441e7f07/core/org.eclipse.birt.core/src/org/eclipse/birt/core/framework/Platform.java#L68

merks:
Ah that’s good reference! The point is that the
 runtime test launches both runtimes the same way using just a URL class loader. For the OSGi runtime it should use what you showed to test it as it’s really used.

wimjongman:
Ok, the birt.war files in both runtimes do not run correctly.

* The birt.war in the osgi runtime can be deployed but the test applications don't work
* The birt.war in the normal runtime can not be deployed
*

wimjongman:
The problem was that java 17 is required. When running tomcat with Java 17 both wars were deployed but the example reports are not running correctly, I think. This is all I see:

image

merks:
I guess that's yet another way of running compared to launch from the IDE where it works:

image

wimjongman:
The only thing that is logged for the failing report in tomcat is:

25-Jun-2023 13:55:56.039 WARNING [http-nio-8181-exec-8] org.apache.axis.utils.JavaUtils.isAttachmentSupported Unable to find required classes (javax.activation.DataHandler and javax.mail.internet.MimeMultipart). Attachment support is disabled.

merks:
javax.activation is in the lib folder.  I don't see javax.mail in the older lib though.

speckyspooky:
I tested the birt.war of osgi-runtime from my side Tomcat 9.0.76 & Java 17.0.5+8 with the latest built of yesterday evening.
(It is the same version of my test this morning with the WebViewerExample.)

And @wimjongman from my side all is running:
- deployed: ok
- service running: ok
- demo reports: ok

I can confirm that the error-log of tomcat shows 1 warning:
org.apache.axis.utils.JavaUtils.isAttachmentSupported
=> unable to find required classes: javax.activation.DataHandler & javax.mail.internet.MimeMultipart
=> result: Attachment support is disabled.


### My Screen of my testing:


birt-war-ok



### Screen of error-log:

birt-war-warnings

speckyspooky:
The warning can be solved with the added dependency of:
- Fix - javax.mail.internet.MimeMultipart => dependency: javax.mail-1.4.6.jar
  -  This will be fix:  javax.mail.internet.MimeMultipart 

- Fix - javax.activation.DataHandler => unnecessary because already given through: javax.activation_1.1.0.v201211130549.jar
  - The dependency is already given in the lib-folder.

My test avoid the warning with adding the "javax.mail-1.4.6.jar" to the lib-folder.
(But I have necer seen mail functionality on "Birt" - big question mark on my brain side.)

speckyspooky:
Now I tested the latest built from yesterday evening "birt-runtime-4.14.0-202306241949" standard (None-OSGI).

Test steps:
- Test 01: WebViewerExample to "birt"-service
- Test 02: "birt.war"-deployment

Result:
- Test 01: Successful, service started, example report created
- Test 02: Successful, service deployed, service started, example report created

The only thing was the the standard-error-log shows the same warning like OSGI but the rest only notes for information (no critical entries!)


### My result screens

birt-runtime



### BIRT-OSGI-Runtime and BIRT-Standard-Runtime tested with FireFox & Chrome and looks good on my side

wimjongman:
I am running my Tomcat on Ubuntu 22. I can't get the reports to popup in any browser. Maybe it is my config, I will try again with the latest builds.

Issue #1356, opened 2023-07-04
WebViewer: Add calender picker for date report parameter


speckyspooky:
According to an older request #939 to add a "dade picker" I plan an implementation based on the HTML5
to add the option of the types "date", "datetime-local" and "time" to the parameter-dialog.

The plan is to use the standard behavior of the browser based on the input-tag.
Therefore the display of the datepicker is dependent on the browser-implementation. The browser Chrome & Edge (Chromium based) will have the best support by standard (without additional libraries). The browser FirFox will sho only the date-picker without support of datetime- and time-picker.

But it will be a good improvement and we will also be able to add a better date-picker-library later in a next step.

The plan is not to change the JSP-fragment by default. The better way is that we will have new options on eclipse-dialog of the parameters based on the data type to select an additional kind of display "Display As".

### Plan of eclipse-dialog

![Folie1](https://github.com/eclipse-birt/birt/assets/41593722/b132fdfc-a31f-4ec6-bcf1-f6a8e1c35b43)

### Date-Picker, pilot implementation displayed on Chrome

![Folie2](https://github.com/eclipse-birt/birt/assets/41593722/1eef1c33-add6-489b-8660-75a3042d0e60)

claesrosell:
@speckyspooky I have an "almost done" implementation of this that I have not pushed. What is left is support for jsp-fragments, some exception regarding context initialization from Jetty.
Would you be interested in this?

speckyspooky:
@claesrosell
Yes, I'm interested on it, so I can compare it with my version.
I have done my implementation through the whole parameter line and do currently some advanced tests.

claesrosell:
Here is a link to my branch : https://github.com/claesrosell/birt/tree/926-calender-picker-for-date-report-parameter
It is not rebased against master.

speckyspooky:
@claesrosell
I reviewed your changes and there is a main difference to my version.
You have defined a new display type for the parameter "DATE_CTRL".
My version instead is based on new "Display As"-style type and the display type keeps a textbox.
With the display-style I handle all version of date, datetime, time and the according options of only date w/ or w/o time picker and the version of short or long time picker.

I would prefer the "Display As"-version. Do you thought about this way?

claesrosell:
No, I did not consider that. I do not have a preferred way of solving this. I just tried to mimic how other controls were implemented.

speckyspooky:
The enhancement was merged with #1359

viniciusalvess:
[...]

PR #1359, opened 2023-07-07
WebViewer, enhance parameter display options with date picker (#1356)

speckyspooky:
The change includes the enhancement according to #1356 to add the date, time and datetime picker.

Addition on it it fix 3 additional topics:
1. Fix typo of DATETIEM_FORMAT_TYPE_* to "DATETIME_FORMAT_TYPE_*
2. Add WebPreview get the BIRT-icon as favicon
3. Optimize the HTML-declaration of FramesetFragment.jsp

### DIFF
--- a/UI/org.eclipse.birt.report.designer.core/src/org/eclipse/birt/report/designer/util/FormatDateTimePattern.java
+++ b/UI/org.eclipse.birt.report.designer.core/src/org/eclipse/birt/report/designer/util/FormatDateTimePattern.java
@@ -98,7 +98,7 @@ public class FormatDateTimePattern {
 	}
 
 	/**
-	 * Retrieves format pattern from arrays given format type categorys.
+	 * Retrieves format pattern from arrays given format type categories.
 	 *
 	 * @param category Given format type category.
 	 * @return The corresponding format pattern string.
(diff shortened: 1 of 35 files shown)


speckyspooky:
@merks I got an Jenkins error which means that the "program.dat" was missing.
This seams to be an build file for Jenkins.

Is it possible to restart the Jenkins process again and how can I do this?

jenkins-error

merks:
If you login using your eclipse account you can invoke Rebuild:

image

Another way is to force push some small change, e.g, to the commit message.

The new build seems to be progressing properly...

claesrosell:
I am currently on vacation and do not have my computer with me, hence I am not able to test this PR. From my mobile phone, it looks tidy though.
There are a lot of changed files, but this is mostly due to some clean-ups and correction of spelling errors.

One thing that draw my attention was the new patterns in FormatDateTimePattern.getPatternForCategory(). Are they used solely for parsing the data time? If it is used for presentation as well, they should probably be localized patterns instead.

Otherwise, from my perspective, this PR is okay.

speckyspooky:
Many thanks for your fast replay and your verification during your vacation(!)

You are right the most changes are coming from the typo fix of "DATETIEM_FORMAT_TYPE_* to "DATETIME_FORMAT_TYPE_*".

And comment to "FormatDateTimePattern.getPatternForCategory()", the date time format is explicit set to non locale format because the value will be used as parsing-value for the date picker. At the beginning I used the local date time but the date will be used to prefill the default value of the date picker and not all date formats will be accepted.

The date format which will work every case is the technical format yyyy-mm-dd (which is also used on the default-value label).
Therefor I decided to use the technical format which works in every case instead to implements an additional handling.

Issue #1365, opened 2023-07-13
Issue on prepared statement if the parameter is null (e.g. optional parameter, research result of #579)


speckyspooky:
I analysed the old PR #579.
It didn't understand what the target in full it seems to me to improve the exception message to get it more readable.

But I was interested on the main reason of the issue, so I veryfied the testing report at the communication and started a deep research.
And the reason of the exception is I think an "old"-topic/"known error". The reason is that the query cannot be executed if the BIRT-parameter of a query-parameter is null.
The most common way is here to use an optional parameter and set him to "null value" (then we will get the exception).

The technical cause of the exception is the implementation of the jdbc-method on the prepared statement to set the "null" value with "isNull()". The problem there is that the parameter type should be fetched to set the value but the current implementation won't be supported through Oracle with the ojdbc-driver:

relevant implementation line at class "Statement", method "setNull":
`this.preStat.setNull(parameterId, pm.getParameterType(parameterId));`

My suggestion to fix it would be to use the standard way which supports oracle in combination with the current version.

![statement-setnull](https://github.com/eclipse-birt/birt/assets/41593722/7311e722-10d1-484f-a8ff-559a4af1a828)

I have done the tests in both versions of statements which was descripted on #579 and I tested it with Oracle V19 and MS SQL Server 2019. With the small change the exception will be avoid. Here my result screens:

![statement fixed](https://github.com/eclipse-birt/birt/assets/41593722/2e8dc7da-64af-4c43-9610-3748d530c9f9)

I would prefer to create here the according PR to fix it.
Any hint to improve the change or avoid the "doubled try-catch" is welcome.

hvbtup:
I'm too busy to test it, but to me it seems reasonable.
I remember when I debugged this back then, it seemed to me that BIRT wasn't able to detect the type correctly.
Which confused me, because it should be straightforward to determine the parameter type without magic type because the type is defined in the rptdesign. But then again, it's Javascript...

The idea to test it with an optional report parameter is good, however, please test it with a sub-report, too:

Say with we have a table PERSONS containing colums ID (as primary key), NAME, MOTHER_ID, FATHER_ID
(First test: The ID columns are VARCHAR2(10), second test: The ID columnd are NUMBER(10).
For most persons, MOTHER_ID and FATHER_ID are set.
For an orphan, MOTHER_ID and FATHER_ID are NULL.
For other persons, maybe only the FATHER_ID is NULL.

Now we want to create a report like this:

[...]

PR #1371, opened 2023-07-15
Fix prepared statement setNull() #1365

speckyspooky:
Fix prepared statement setNull() #1365

### DIFF
--- a/data/org.eclipse.birt.report.data.oda.jdbc/src/org/eclipse/birt/report/data/oda/jdbc/Statement.java
+++ b/data/org.eclipse.birt.report.data.oda.jdbc/src/org/eclipse/birt/report/data/oda/jdbc/Statement.java
@@ -750,10 +750,20 @@ public class Statement implements IQuery {
 		try {
 			java.sql.ParameterMetaData pm = this.preStat.getParameterMetaData();
 			if (pm == null) {
-				this.preStat.setNull(parameterId, java.sql.Types.OTHER);
+				try {
+					this.preStat.setNull(parameterId, java.sql.Types.OTHER);
+				} catch (SQLException e) {
+					// fallback, set the null value with SQL default null type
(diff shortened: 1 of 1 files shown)


speckyspooky:
@merks 
Hi there today I have a problem with the build process on jenkins side.
You fixed the a mylin issue with #1367 to correct the build process on Jenkin side I get a familiar issue.
I'm wondering because your last bult was running and also my production merge.

The jenkins error message are:

jenkins-error2

jenkins-error

Could you please assist here again. Thanks in advance.

merks:
The error suggests you are using an older version of the target platform in the build, i.e., comparing before and after the fix I did:

image

I'm going to rebase your branch:

image

That should start a new build with the latest version of the target platform.

merks:
Note that any other PRs you started before the target platform was fixed will need to be rebased if you need to build them again...

speckyspooky:
Sorry, I'm didn't understand the issue because I checked in the "Statement.class" and before I pushed all changes fro birt-master-branche into my folk.
Yes, I see in your screen that my version is an older and not corresponding to your fix. But I'm confused because realy I thought with the merge from master to my folk to my branch I have the latest version.

speckyspooky:
But you are right I created the new local branch before your fix was bushed and after your push I refreshed.
If I understand you then it isn't enough to start only a change refresh from the master-branch in this case.

merks:
I'm not such a great Git expert.    I personally try to ensure  I have master checked out and pulled.   Then I create my branch, commit, and push to my fork as the remote.  I avoid working on more than one PR at a time.  I'm not sure how to manually rebase after master has diverged.   My poor man's approach is to rename the branch.  Check out master and pull.  Create a new branch with the same name as I used before.  Cherry pick the change from the renamed branch on the newly-created branch, and force push to my fork.  I'm sure there's a better way. 😱

Issue #1372, opened 2023-07-15
BIRT 4.14, elements "label" & "dynamic text" with display "inline" throw NullPointerException


speckyspooky:
I created a report with labels (element: label) in a grid and set the display to "inline".
Currently this constellation can cause an NulPointerException.
The issue is also given in case of "table" usage and inline "label" or inline "text".

![textlabel-inline-error](https://github.com/eclipse-birt/birt/assets/41593722/81ec7fca-9a41-4e12-88bc-cc218853fff1)

The issue is located in the LineArea class in method "align()".
I will create a PR for that issue.

speckyspooky:
This bug is fixed with PR #1373

PR #1373, opened 2023-07-15
Fix align issue of "inline" elements (#1372)

speckyspooky:
Fix align issue of "inline" elements (#1372)


hvbtup:
While I appreciate the code changes, this PR contains unrelated commits.
I just want to mention this (I've made this mistake myself as well).
Perhaps you should wait for a review before merging next time.

speckyspooky:
Yes, I saw it. Normally I wouldn't merge such combination of 2 different topics.
I don't know why Git recognized both changes. Because I created competely new branch and before I fetched all changes from the master. The second change is from my PR of "prepared statement"-fix.
And the merge to the master shows that there was only 1 file changed which is the same result on my local staging result was. 
Sometimes it confused me, because it is hard if the local git-staging comes to another result like the web-git.

So yes, before I have done it I tribble checked the changes from ym side. But you are right it isn't perfect(!)

Issue #1378, opened 2023-07-24
BIRT exception dialog, add new option to copy the stack trace directly to clipboard


speckyspooky:
During the report development of my team we have many to do with the exception dialog.
The exceptions and the according error stack trace is correct. The topic is that we share the error trace to fix our issues.
But in every case we have to mark the trace into the small dialog copy it to an editor to share it.

Therefore I plan a very small enhancement to offer a direct copy function to the clipboard
and increase the exception dialog in a smart way to get the trace a little bit more readable.

### Draft screen of the copy function add the exception dialog

![exception-dialog](https://github.com/eclipse-birt/birt/assets/41593722/268cf803-6f9f-4044-9558-e86f6d007d57)

speckyspooky:
I will create a PR for that change.

speckyspooky:
Solved with PR #1381

PR #1381, opened 2023-07-25
Improve the exception dialog of the birt viewer (#1378)

speckyspooky:
BIRT exception dialog, add new option to copy the stack trace directly to clipboard #1378


hvbtup (approved):
Nice!

Issue #1401, opened 2023-08-30
Images with size unit percentage should be sized on the original image size


speckyspooky:
Based on the discussion #1388 I started a small investigation on the BIRT-behavior.
The result is that the eclipse designer set the height on unit percentage based on the used font size.
The width will be set on percentage unit based on the container arround.

But the original version of the BIRT definition is that the image ratio of an image element if the unit percentage is used should be based on the original image size.

![resize-image](https://github.com/eclipse-birt/birt/assets/41593722/48ada57c-9d17-406a-b3bf-28f520915024)

Therefore I would investigate and try to create a small enhancement to change the behavior on the folowwing elements:
- BIRT-designer
- Preview / HTML emitter
- PDF emitter
- DOCX emitter
- PPTX emitter

The plan is to have the original description (ration based on the original images) supported for raster images.

### The current presentation on the BIRT-designer
=> 1st row percentage unit & 2nd row according px dimensions:

![percentage-designer-error](https://github.com/eclipse-birt/birt/assets/41593722/16f7f298-9761-4c68-9ed5-972826da9730)

### The current presentation on the preview:
=> 1st row percentage unit & 2nd row according px dimensions:

![percentage-preview-error](https://github.com/eclipse-birt/birt/assets/41593722/5cc90283-0eb3-4313-b9f8-5c6cea279e70)

hvbtup:
I think that part of the text excerpt of  that book (I've got one, too) is misleading.

Regarding the text "resizing an image ... typically results in a degradation of its appearance":

I think a better wording would be "may result" instead of "typically results".
This heavily depends on the output format, the type of the image and the environment where the report is read.
For example, in our application, "image" means "photo" in ~ 90% of the cases, "company logo" in 9% of the cases, and "icon" in very rare cases; and the output format is PDF in 98% of the cases.
The photos are stored in a database and it makes no sense to preprocess them to fit to particular pixel dimensions.
So the PDF report contains them with their native dimensions, but with a restricted physical size (preserving aspect ratio).
The reading environment may be a printed report or a PDF file. In the latter case, a viewer may zoom in, so it would be bad practice to store the image with only 72 or 96 dpi at a maximum width/height of eg. 16cm.

However, one reason to resize an image to lower pixel dimensions (before using it in a report) is the crazy pixel dimension that modern cameras and smart phones use (e.g. 30 Megapixels). This only blows up the image files and thus the PDF files.

Regarding "width and height as a percentage of the original size":
[...]

PR #1406, opened 2023-09-01
Images with dimension unit percentage should be sized on the original  (#1401)

speckyspooky:
Images with dimension unit percentage should be sized on the original image size (raster images: px of width and height) (#1401)

### DIFF
--- a/UI/org.eclipse.birt.report.designer.core/src/org/eclipse/birt/report/designer/core/model/schematic/ImageHandleAdapter.java
+++ b/UI/org.eclipse.birt.report.designer.core/src/org/eclipse/birt/report/designer/core/model/schematic/ImageHandleAdapter.java
@@ -129,7 +129,7 @@ public class ImageHandleAdapter extends ReportItemtHandleAdapter {
 
 	/**
 	 * Evaluate the size of the image based on px
-	 * 
+	 *
 	 * @param getRawSize return the real raw size
 	 * @return Return the requested size of the image (unit px)
 	 */
@@ -137,14 +137,13 @@ public class ImageHandleAdapter extends ReportItemtHandleAdapter {
 		int px = 0;
 		int py = 0;
 
-		DimensionHandle widthHandle = getImageHandle().getWidth();
+		DimensionHandle widthHandle = this.getImageHandle().getWidth();
 		if (this.imageFigureSize != null && DesignChoiceConstants.UNITS_PERCENTAGE.equals(widthHandle.getUnits())) {
 			px = (int) DEUtil.convertToPixel(widthHandle, this.imageFigureSize.width, DesignChoiceConstants.UNITS_PX);
 		} else {
 			px = (int) DEUtil.convertoToPixel(widthHandle);
 		}
-
-		DimensionHandle heightHandle = getImageHandle().getHeight();
+		DimensionHandle heightHandle = this.getImageHandle().getHeight();
 		if (this.imageFigureSize != null && DesignChoiceConstants.UNITS_PERCENTAGE.equals(heightHandle.getUnits())) {
 			py = (int) DEUtil.convertToPixel(heightHandle, this.imageFigureSize.height, DesignChoiceConstants.UNITS_PX);
 		} else {
@@ -158,6 +157,11 @@ public class ImageHandleAdapter extends ReportItemtHandleAdapter {
 			if (py == 0 && heightHandle.isSet()) {
 				py = 1;
 			}
+		}
+
+		// proportional scale of the image size
+		if (this.getImageHandle().isProportionalScale()) {
+			py = px;
 		}
 
 		// return the real raw size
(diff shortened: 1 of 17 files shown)


engine/org.eclipse.birt.report.engine.emitter.wpml/src/org/eclipse/birt/report/engine/emitter/wpml/AbstractEmitterImpl.java
```diff
@@ -1055,6 +1057,28 @@ public void startImage(IImageContent image) {
 				float scale = ((float) imageInfo.getHeight()) / ((float) imageInfo.getWidth());
 				height = width * scale;
 			}
+System.out.println("==== standard ==== ");
```
wimjongman, line 1060: Debugging code?

wimjongman (changes requested):

hvbtup (commented):
As you are working with the code, you understand the units of the size values (at least I hope so :-). When you add or modify a method, can you please document the units of size values where appropriate? Eg. is it pixels, is it points, etc? 

And in ImageHandleAdapther.java you created a new public, undocumented method `get2Size()`. I think the naming and documentation could be improved here.

I only looked at the source code, did not test anything.

engine/org.eclipse.birt.report.engine.emitter.wpml/src/org/eclipse/birt/report/engine/emitter/wpml/AbstractEmitterImpl.java
speckyspooky: Confirmed, I will remove the output lines.

speckyspooky:
@hvbtup Thanks for checking the code. I have changed some comments and with the latest version "get2Size()" will be removed completely.

speckyspooky:
@wimjongman & @hvbtup

Thanks to your both of verification both findings are changed and checked in.
Additional on it I attached my test report with embedded images and the calculationexample based in comapre 1st row "percentage units" and 2nd row "px units".

Report & output examples (PDF & DOCX):
[Test examples of percentage.zip](https://github.com/eclipse-birt/birt/files/12541776/Test.examples.of.percentage.zip)

Or here you can download the PF directly:
[Images-embedded-percentage-original-size.pdf](https://github.com/eclipse-birt/birt/files/12541808/Images-embedded-percentage-original-size.pdf)

hvbtup:
Can you please test this report, too?
[test-proportional.zip](https://github.com/eclipse-birt/birt/files/12545793/test-proportional.zip)
The PDF output should look like this:
grafik

speckyspooky:
@hvbtup I have tested your report and before I never worked with your image property.

The result of the PDF is fine. Attached you get a screen and my created image of the current version.:

2023-09-07 20_38_22-PowerPoint-Bildschirmpräsentation  -  Präsentation1

[test-image-4.14.pdf](https://github.com/eclipse-birt/birt/files/12552462/test-image-4.14.pdf)

speckyspooky:
But thi property was never supported on designer level because the property is set
but the designer shows not the scaled image it shows the unit property without your scale-property.
I tested BIRT 4.6, 4.10, 4.12 and 4.14 and the designer result is that:
2023-09-07 20_46_23-C__Users_gutma_Downloads_2023-09-07 20_38_22-PowerPoint-Bildschirmpräsentation


I will try to check it but this hasn't to do with this change and if I got no result I would finish this pull request.
I will let you know.

speckyspooky:
The designer is now checking the proportional flag which is included in the ImageHandleAdapter.java on method evaluateSize()

UI/org.eclipse.birt.report.designer.core/src/org/eclipse/birt/report/designer/core/model/schematic/ImageHandleAdapter.java
```diff
@@ -145,7 +159,21 @@ public Dimension getRawSize() {
 			}
 		}
 
-		return new Dimension(Math.max(px, 0), Math.max(py, 0));
+		// proportional scale of the image size
+		if (this.getImageHandle().isProportionalScale()) {
+			py = px;
```
hvbtup, line 164: I don't understand the code, but my gut feeling is that this might be wrong, even though my test report looks right.

Can you please also test reports where the image itself is not a square?

speckyspooky:
Hello Both, 
could you please give me some notes of it.
If you have no further requests of change I would plan to merge the changes to the master this week on Thursday evening.

hvbtup:
[test_image.zip](https://github.com/eclipse-birt/birt/files/12594982/test_image.zip)

@speckyspooky Thomas, I tested PDF output for typical use cases *without* percent values and the PDF output looks ok.

However, in the designer, the photo is shown way to large:
grafik

I don't know if this is actually related to your changes or not, but this didn't happen in my (patched) 4.9.0 designer.

It seems as if the butterfly image is rendered with a height of 21,33 inch = approx. 542mm (based on the pointless 72x72 dpi metadata from the image file), although the height in BIRT is restricted to 100mm.

Can you please take a look at it?

speckyspooky:
@hvbtup
Hello Henning, thanks for your good test cases and the testing from your side too.
The rendering topic is that the "width" will be used and the open task is that the case not check if the image-width is set.

Only the proportional-flag is checked, so I will do an further check if the image-width really set
and only then the proportional-size will be used.

speckyspooky:
I have added the additional condition and below is the same screen with the new version
based on your report-screen of the designer.

I checked if the size-unit is set because the size-value is set also on empty values to 0 but the size-unit only if the designer has an valid image entry. So you can enter also 0mm, make no sense but you can do it.

hvbtup (approved):
Based on the test output which is now correct in my opinion: OK from my side and a great "Thank You" for your effort.

Issue #1418, opened 2023-09-17
Remove Cassandra Scripted Data Source and Hector-PlugIn


speckyspooky:
The special scripted data source of Cassandra should be removed.
The Cassandra data source is end of live and also the use connection client based on hector-plugin is inactive since 2015.

Based on the current implementation we have some integration issues with SLF4J (#1332), guava-classes (#1376)
and we have also 2 "Dependabot alerts" (level low & moderate). 

The hector plugin is inactive since 2015 - https://github.com/hector-client/hector
Therefor no further changes will be done on th current plugin.

Reference extraction is documented here:
https://github.com/triestram-partner/birt/commit/93f18da8eb8275ba08c1608a5f39eae916898d3b?diff=split

I will create a special PR to remove the "Cassandra Scripted Data Source" with the hector plugin.

(The change was requested the first time with #951)

speckyspooky:
The Cassandra Scripted Data Source is removed with PR #1419

PR #1419, opened 2023-09-17
Remove Cassandra Scripted Data Source & the Hector PlugIn (#1418)

speckyspooky:
According to the issue-ticket (#1418) the "Cassandra Scripted Data Source" and the Hecotr PlugIn will be removed.


speckyspooky:
Hello Wim & Henning, I have you both added to review.

- Henning I used your "data source extraction" verified the details of Cassandra and removed the Cassandra-(Hector)-parts. So I would think you will remember on the most of the removed parts.

- Wim it would be nice too if you can take a look into it especially in your role as project leader before the option will be removed.

wimjongman (approved):
+1 for the removal.

hvbtup (approved):
Your changes seem to correspond to what I did (for other DBs as well in my case).
Assuming that you also did a case-insensitive full text scan for CASSANDRA and HECTOR in the birt directory (both should return 0 results), +1 from me

Issue #1281, opened 2023-05-12
IPlatform:org.eclipse.birt.core.framework.Platform.platform is not initializing after OSGi PlatformLauncher successful startup


94satya:
For environment information please check the below details:
Java- Java11
Birt: 4.9.0
BIRT_HOME: ReportEngine\platform
OSGi: org.eclipse.osgi_3.17.200.v20220215-2237.jar
Compatibilty state: org.eclipse.osgi.compatibility.state-1.2.600.jar

Link to get the Birt OSGi runtime: https://archive.eclipse.org/birt/downloads/drops/R-R1-4.9.0-202203161719/

Any runtime configurations/ any dependency is missing related to OSGi?

speckyspooky:
Have you tried it with the latest version 4.13?
The 4.9.0 an older version and currently we work only with the latest versions of BIRT.

94satya:
As per suggestion, Tried with 4.13. Now platform is initializing but facing EmitterExtension loading issue for below
1- org.eclipse.birt.report.engine.emitter.config.postscript
2- org.eclipse.birt.report.engine.emitter.config.pptx
3-  org.eclipse.birt.report.engine.emitter.docx 
4-  org.eclipse.birt.report.engine.emitter.pdf 
and few more emitters.

Part of stacktrace (getting inside configuration directory of the OSGi report engine)

!ENTRY org.eclipse.birt.report.engine.emitter.pdf 4 0 2023-05-16 23:14:32.225
!MESSAGE FrameworkEvent ERROR
!STACK 0
org.osgi.framework.BundleException: Could not resolve module: org.eclipse.birt.report.engine.emitter.pdf [94]
  Bundle was not resolved because of a uses constraint violation.
[...]

PR #1423, opened 2023-09-19
IPlatform:org.eclipse.birt.core.framework.Platform.platform is not

94satya:
IPlatform:org.eclipse.birt.core.framework.Platform.platform is not initializing after OSGi Platform Launcher successful startup #1281

Updated fixed version number with version range for openpdf

### DIFF
--- a/engine/org.eclipse.birt.report.engine.emitter.odp/META-INF/MANIFEST.MF
+++ b/engine/org.eclipse.birt.report.engine.emitter.odp/META-INF/MANIFEST.MF
@@ -6,7 +6,7 @@ Bundle-Version: 4.14.0.qualifier
 Bundle-ClassPath: .
 Bundle-Vendor: %Bundle-Vendor
 Require-Bundle: org.eclipse.birt.report.engine;bundle-version="[2.1.0,5.0.0)",
- com.github.librepdf.openpdf;bundle-version="1.3.26",
+ com.github.librepdf.openpdf;bundle-version="[1.3.0,3.0.0)",
  org.eclipse.birt.report.engine.odf;bundle-version="4.13.0"
 Bundle-ActivationPolicy: lazy
 Bundle-Activator: org.eclipse.birt.report.engine.emitter.odp.plugin.OdpEmitterPlugin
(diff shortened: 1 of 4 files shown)


merks:
Which version of  com.github.librepdf.openpdf is present?  This is in the target platform:

image

So mostly your change is to exclude any version 3.x or higher, though also allowing some lower versions though lower than what was built against and tested.  These are all the versions:

https://repo1.maven.org/maven2/com/github/librepdf/openpdf/

So I don't really understand the point of this change, but I also don't see that it causes any harm...

94satya:
> Which version of com.github.librepdf.openpdf is present? This is in the target platform:
> 
> image
> 
> So mostly your change is to exclude any version 3.x or higher, though also allowing some lower versions though lower than what was built against and tested. These are all the versions:
> 
> https://repo1.maven.org/maven2/com/github/librepdf/openpdf/
> 
> So I don't really understand the point of this change, but I also don't see that it causes any harm...

In the Target 1.3.30 is present.
Earlier fixed version was used for the openpdf dependency, As per discussion with Thomas Gutmann (https://github.com/speckyspooky) given range is used. PFA screenshot for more details.

image

94satya:
> @94satya
> 
> You will need to sign an ECA as described here:
> 
> https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#setting-up-your-eclipse-and-github-account

Thanks @merks Signed ECA, If anything else needed please let me know. Thanks

merks:
I tried revalidating but it still fails.  Is the email you used for the commit the same one associated with your Eclipse account?  Maybe it just takes a while for the database to update and I have to try again tomorrow.

94satya:
> I tried revalidating but it still fails. Is the email you used for the commit the same one associated with your Eclipse account? Maybe it just takes a while for the database to update and I have to try again tomorrow.

I can see different id(contains few parts of my registered mail id). I followed eclipse-launcher process and automatically this id got used. Not sure how this can be resolved. @merks  any suggestions would help?

merks:
@94satya 

When I look at your commit, there isn't even a link to your account:

image

For other similar PR comments I can see a link to their account:

image

For the validation I see anonymous:

image

For similar PRs I see this:

image

So I don't think the author information in your commit is complete such that it is properly associated with your GitHub account.  I think you will need to amend your comment with proper author/committer details:

image

merks:
@94satya 

This looks improved:

image

Unfortunately there now there are two commits and the one comment is the "bad" one.  You need to squash this into a single commit with the right author/committer and force push the PR...  If that's too tricky, just create a new PR with a single correct commit...

94satya:
> @94satya
> 
> This looks improved:
> 
> image
> 
> Unfortunately there now there are two commits and the one comment is the "bad" one. You need to squash this into a single commit with the right author/committer and force push the PR... If that's too tricky, just create a new PR with a single correct commit...

I tried amend the author, but yeah as you mentioned it has two commits includes bad one.

speckyspooky:
I would suggest close this PR and create e new one because so you would avoid any conflicts with the old one.

94satya:
Thanks @speckyspooky  @merks @wimjongman closing this PR. Will raise a new PR now.

94satya:
New PR raised with correct author
PR: https://github.com/eclipse-birt/birt/pull/1429

Issue #998, opened 2022-06-21
Exception "invalid cell range" when using Spudsoft Excel emitter with AutoFilter ON


rneumerkel:
After upgrading from BIRT 4.3 to BIRT 4.9 we experience the following behaviour:

When trying to export a report consisting of a text element and a table element using the Spudsoft Excel emitter with `AutoFilter` enabled, we sometimes get the following exception from within Apache POI (CellRangeAddress.java):

`java.lang.IllegalArgumentException: Invalid cell range, having lastRow < firstRow || lastCol < firstCol`

This always happens when the dataset returns no results, leaving just the header row of the table along with the text field preceeding it. In this case, when trying to apply the auto filter, the starting row equals 1 while the last row equals 0, hence the exception from Apache POI.

Note 1: this ONLY happens when there is at least one text-element or label preceeding the empty table.
Note 2: if the table DOES have values, then the report is generated without errors

I have attached a sample report using the built-in datasource ClassicModel to demonstrate this behaviour.

How to reproduce:

1. Open the attached report in Eclipse BIRT designer 4.9
2. Run "View Report as XLS_Spudsoft" or "XLSX"

How to fix:

In `uk.co.spudsoft.birt.emitters.excel.handlers.TopLevelTableHandler.endTable(..)` (and possibly other locations), the code should check if startrow <= endrow and startCol <= endCol before attempting to call `new CellRangeAddress(..)`

[bug_report.zip](https://github.com/eclipse/birt/files/8947734/bug_report.zip)

wimjongman:
Thanks for reporting. Can you supply a patch?

https://github.com/eclipse/birt/blob/master/CONTRIBUTING.md

wimjongman:
Thank you. It is important that you:

* Create an Eclipse account
* Set your Github id in the Eclipse account
* Sign the Eclipse Contributor Agreement
* Use the  e-mail address in your Eclipse account to commit

speckyspooky:
The issue will be fixed with PR #1431

PR #1431, opened 2023-09-22
Fix issue with spudsoft auto filter to avoid cell range issue (#998)

speckyspooky:
Fix issue with spudsoft auto filter to avoid cell range issue (#998)


speckyspooky:
The fix avoid the cell range error and set the auto-filter to the correct line.

### Screen of fixed version based on the demo report (bug_report.rptdesign #998 )
2023-09-23 16_17_53-bug_report xlsx  -  Geschützte Ansicht - Excel

speckyspooky:
@wimjongman
Hello Wim, my plan would be to merge that PR for the upcoming release of BIRT 4.14.
It would be nice if you had time to review the change.

wimjongman (approved):
Thanks, Thomas.

PR #1439, opened 2023-09-28
Remove ECF from products

merks:
THe products are over-specified and include directly things that are also indirectly induced by other things.  In this case, the Platform's evolving use of ECF is causing a problem.

https://github.com/eclipse-birt/birt/pull/1433

### DIFF
--- a/UI/org.eclipse.birt.report.designer.ui.rcp/BIRT-headless.product
+++ b/UI/org.eclipse.birt.report.designer.ui.rcp/BIRT-headless.product
@@ -358,12 +358,6 @@ United States, other countries, or both.
       <plugin id="org.eclipse.e4.ui.workbench.renderers.swt"/>
       <plugin id="org.eclipse.e4.ui.workbench.swt"/>
       <plugin id="org.eclipse.e4.ui.workbench3"/>
-      <plugin id="org.eclipse.ecf"/>
-      <plugin id="org.eclipse.ecf.filetransfer"/>
-      <plugin id="org.eclipse.ecf.identity"/>
-      <plugin id="org.eclipse.ecf.provider.filetransfer"/>
-      <plugin id="org.eclipse.ecf.provider.filetransfer.ssl" fragment="true"/>
-      <plugin id="org.eclipse.ecf.ssl" fragment="true"/>
       <plugin id="org.eclipse.emf.common"/>
       <plugin id="org.eclipse.emf.ecore"/>
       <plugin id="org.eclipse.emf.ecore.change"/>
--- a/UI/org.eclipse.birt.report.designer.ui.rcp/BIRT.product
+++ b/UI/org.eclipse.birt.report.designer.ui.rcp/BIRT.product
@@ -361,12 +361,6 @@ United States, other countries, or both.
       <plugin id="org.eclipse.e4.ui.workbench.renderers.swt"/>
       <plugin id="org.eclipse.e4.ui.workbench.swt"/>
       <plugin id="org.eclipse.e4.ui.workbench3"/>
-      <plugin id="org.eclipse.ecf"/>
-      <plugin id="org.eclipse.ecf.filetransfer"/>
-      <plugin id="org.eclipse.ecf.identity"/>
-      <plugin id="org.eclipse.ecf.provider.filetransfer"/>
-      <plugin id="org.eclipse.ecf.provider.filetransfer.ssl" fragment="true"/>
-      <plugin id="org.eclipse.ecf.ssl" fragment="true"/>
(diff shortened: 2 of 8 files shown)


merks:
@speckyspooky 

FYI, these are  much bigger set of changes.  At this point there are two choices:

1. Build against Eclipse Platform 4.29 with older stable versions.
2. Build against Eclipse Platform 4.30 with newer versions of **many** things.

There are new versions of things such as batik, and soon also new versions of other BIRT dependencies, e.g., poi will be refreshed.  Eventually one will be forced to take the second route, and, long term, to eliminate CVE-ridden libraries, one must follow this route.  But, because of the rather fragile nature of how the runtimes are created, there is a non-zero chance that we will yet again have problems with things missing from the runtime....

speckyspooky:
Thanks for your change.
Which eclipse plattform is currently added in our built 4.29 (older version)?

merks:
https://github.com/eclipse-birt/birt/blob/bf6b8cf8ff7015cd17417183d72a5a65580cef48/build/org.eclipse.birt.target/org.eclipse.birt.target.target#L320

merks:
@speckyspooky 

The build worked with these changes.  So I think we are good to move forward. Shall I merge it?

I would just suggest testing/inspecting the runtimes to make sure they work properly and nothing important has gone missing.

speckyspooky:
Yes, I agree on it to merge the change.
I will update my fork afterwards directly and will test this change with the fixing of the current background issue.

speckyspooky:
@merks 
I used the new 4.30 versions and I can see know a package issue wich is upcoming on "openpdf", there is "sun.misc" from the module "jdk.unsupported" is missing.
**Do you know if we have an option to add this module with the package?**

In addition to it such a familiar issue is given with the OSGI-BIRT plattform. My suppluier use WildFly 27 with BIRT 4.13 (OSGI), JDK 17 and get the message java.lang.ClassNotFoundException: sun.misc.Unsafe from [Module "deployment.servbird.ear" from Service Module Loader]. This is an reflection issue of the equinox problem.
Perhaps we could solve it with the integration of the module and the package of "sun.misc".

### eclipse warning on designer execution

2023-09-28 21_05_39-Validation


### OSGi-Framework, MultiplexingFactory.java - misc.sun-missing:

Exception-location: https://github.com/eclipse-equinox/equinox/blob/77fda35b312c6da5c6ce1530829804f7512bae7c/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/url/MultiplexingFactory.java#L61C13-L61C24

Equinox discussion: https://github.com/eclipse-equinox/equinox/issues/94

birt413-sun misc unsafe

Let me know if you would like this topic in a seperate ticket, I would create it fro you.

merks:
This is not a problem that I can fix, and I don't think it's actually broken.   

The openpdf bundle, as it comes directly from Maven,

https://repo1.maven.org/maven2/com/github/librepdf/openpdf/1.3.30/

has a non-optional requirement on that package as selected here:

image

The jdk.unsupport module is generally in a JDK and that module provides the sun.misc package:

image

The validation problem is exactly like what's described here:

https://github.com/eclipse-pde/eclipse.pde/issues/429

I.e., it is a bug in PDE that it complains about a missing package that is actually provided by the JRE/JDK.

As for your stack trace, it appears you are using a JRE without the jdk.unsupported module.  I.e., that stack trace is not caused directly or indirectly by the openpdf bundle.   Moreover, this stack trace does not look like a new problem in Equinox.  The print has been there for quite some time, i.e., since this change:

https://git.eclipse.org/r/c/equinox/rt.equinox.framework/+/182046

So the validation dialog is a bug in PDE that should be fixed in the latest release and the stack trace is an Equinox problem that happens when your JRE is missing something that Equinox generally needs and expects.

speckyspooky:
Thanks Ed for your research and explanation then we can go ahead on our side.

merks:
You're welcome.  😄 

I'm currently working through all the Orbit bundles.  One thing I just noticed is there is POI version 5.x available but BIRT is using POI 4.x, in particular 4.1.1.  There is a 4.1.2 available, but I don't believe there will be future 4.x versions but rather only more 5.x versions:

https://repo1.maven.org/maven2/org/apache/poi/poi/

So I think it will be best that I pull in the latest 5.x version and then we'll need to update BIRT's version ranges to use this newer version.  Hopefully that version will just work...

speckyspooky:
Well, the POI update to 5.x would be greate
because with that we can reduce the count of some CVE-findings
and also we would get some new functions for excel 😄

(Ok, somebody has to implements the new functions.)

Issue #1436, opened 2023-09-27
BIRT 4.14 null pointer exception with HTML output concerning background image


reiner-killet:
Hi,
I'm currently testing 4.14 with the RCP designer. All templates were previously running with 4.8. 
WIth the 4.14 HTML emitter I get a null pointer exception:

 java.lang.NullPointerException: Cannot invoke "org.eclipse.birt.report.engine.nLayout.area.style.BackgroundImageInfo.setUri(String)" because "backgroundImage" is null

The same templates are running error-free with DOCX or PDF output.
Can a BIRT developer please view that problem?
I could send rptDesigm & XML files.

Thanks, Reiner

[Full stack trace.txt](https://github.com/eclipse-birt/birt/files/12737868/Full.stack.trace.txt)

speckyspooky:
Hello @reiner-killet,
thanks for your noet and also for the StackTrace. At first yes, there are some changes on the background images.
Based on your StackTrace I figured out the situation that the main object "BackgroundImageInfo" could be null
and therefore the NullPointerException appears.

Could you please provide an example according to your description because all of my reports with backgroundimages on Tabel, Grid, Row, Cell are running.
I tested different options embedded and URL-reference but I'm not able to get your error.

At which element is the background image located?

reiner-killet:
Hello Thomas,

 

We currently do not use any background pictures, only Image tags.

I don’t know where BIRT expects this property.

This what I tried to find out: 

*	Deleted all styles (we use only common styles).
*	Added background pictures to concerned common styles (label, data, text, report, dynamic text).
*	Added background pictures to all tags.

 

I always got the same exception which is not pointing on a certain element.

The element ID would probably be helpful in the message text.

 

Please find enclosed template and sample data.

The workspace path for the pictures should be c:\temp\birt\pisa. 

 

Thanks, Reiner

 

Von: Thomas Gutmann ***@***.***> 
Gesendet: Mittwoch, 27. September 2023 20:22
An: eclipse-birt/birt ***@***.***>
Cc: Reiner Killet ***@***.***>; Mention ***@***.***>
Betreff: Re: [eclipse-birt/birt] BIRT 4.14 null pointer exception with HTML output concerning background image (Issue #1436)

 

Hello @reiner-killet <https://github.com/reiner-killet> ,
thanks for your noet and also for the StackTrace. At first yes, there are some changes on the background images.
Based on your StackTrace I figured out the situation that the main object "BackgroundImageInfo" could be null
[...]

PR #1440, opened 2023-09-28
Fix of images handling of HTML embedded images which cause NullPointerException (#1436)

speckyspooky:
Fix of images which are embedded like URL on an html document which is embedded into a birt report element (e.g. dynamic text)


speckyspooky:
@wimjongman
Hello Wim, this is a fix which should be included into BIRT4.14 to avoid the according exception.

By the way, I plan no further checkin for 4.14 only if we should get some bug notifications.
Other changes will be moved after the release of 4.14 (e.g. change of the font-mapping).

reiner-killet:
Is fix #1440 contained in the nightly build [](https://download.eclipse.org/birt/updates/nightly/latest/index.html)?
I downloaded the RCP designer for Windows from there and the same error message still occurs.

speckyspooky:
The fix is currently not included to the nightly built because I wait of the approval of @wimjongman 
If he has done the quality check I will consider his response and with the approval I can merge it to the master and afterwards it will go into the nightly built.

wimjongman (approved):
Looks good to me, Thomas.

Issue #1458, opened 2023-10-27
Update toolchain to use Tycho 4.0.x and Maven 3.9.x


claesrosell:
It looks like we need to update to Tycho 4.0.x and hence Maven to 3.9.x due to problems encountered when building against newer versions of the platform.

PR #1459, opened 2023-10-27
Update toolchain to use Tycho 4.0.x and Maven 3.9.x #1458

claesrosell:


claesrosell:
@speckyspooky What do you think about the removal of the design dependency? It looks like it was introduced accidentally.
@wimjongman Does the Changes to the build.look okey?

wimjongman (approved):
Looks good to me, Claes. @merks also OK?

merks:
If the build works and the tests ran/passed it must be okay.

claesrosell:
I see that we use some deprecated actions in our CI workflow. We should probably update those as well. I will open a separate issue for that.

speckyspooky:
Normally I would agree Ed, but we have some changes which will definitely not tested with the default tests.
So it good be that in a very specific case it was used but then we have to validate it and optimize the code there.

merks:
@speckyspooky 

There are three changes here.   The ci.yml changes affects only the github actions.  The MANIFEST.MF change affects only dependencies so I believe that if the build runs, then it's not harmful.  The pom.xml change affects which version of Maven/Tycho are used; using newer ones will be necessary sooner or later and it appears that sooner already arrived with the build problems that @claesrosell encountered that drove that change.  

Is it one of these three specific changes that concerns you?  I would have expected that it's the lack of tests for the runtimes and the lack of tests for the viewers that would be our primary cause for concern, not the changes in this PR...

speckyspooky:
Yes, I'm fine with the Tycho update!
The only thing is realy the MANIFEST.MF change. My brain means that I saw such kind of change in the past but I cannot say in which constellation and if the modification & dependency used at the end.
I think the build will work and BIRT wil be runnable. And if I was able to execute my tests and my reports are running then we are fine with all.

Issue #1451, opened 2023-10-13
java.lang.NoClassDefFoundError: javax/xml/rpc/encoding/TypeMappingRegistry when starting WebViewer 


claesrosell:
There are currently problems with the WebViewer in Master. As seen here: https://github.com/eclipse-birt/birt/issues/1447#issuecomment-1760236651

claesrosell:
We had problems with the javax.xml.rpc-api packages when moving to JavaSE-11 a couple of years ago, which ultimately introduced our own version of that dependency. If I recall correctly it had something to do with the OSGI-headers of the Maven version of javax.xml.rpc .

We recently moved away from our own, modified, version of xml-rpc. We need to investigate if this problem is related to those changes.

merks:
Could someone remind me how to "start the web viewer"?

I see "powered by jetty 10.0.15" in the picture...

merks:
It seems you guys see more than I see.  This is all I see with my partial non-functional migration to jetty 12:

![image](https://github.com/eclipse-birt/birt/assets/208716/64f197e5-8cf2-4095-a987-5050de5148df)

This is with this in my workspace:

https://github.com/eclipse-birt/birt/compare/master...merks:birt:issue-1447

The "missing" class is in my workspace

![image](https://github.com/eclipse-birt/birt/assets/208716/ebd6c69d-5c62-46e2-b8cb-7e70625fa5d1)

And both versions are in the launched IDE:

![image](https://github.com/eclipse-birt/birt/assets/208716/c7b1790b-62fb-4316-8d74-7c79911fee95)

speckyspooky:
Have yu an suggestion how I can support you.
I'm not the expert of the final depencencies but if I can help you please let me know.

I checked the latest changes and I don't know if it has an effect but with PR #1451 the javax.xml.rpc version (seems to be a duplicate was deleted).
I can only see a small difference to the current rpc-version and the deleted version which is the bundle version - deleted version was 1.1.2 and the current orbit version is a 1.1.0.

merks:
Note that the old BIRT version packaged some jars (goodness knows where they came from) into a bundle and that bundle put those jars on the classpath.

It's javax.xml-rpc-api bundle that should be used now, not the very very very old java.xml.rpc bundle.  The rpc-api is at version 1.1.4.  This bundle is provided, as the Plug-in Name suggested, by the Jakarata project and is the most recent version.

[...]

PR #1462, opened 2023-10-27
Issue #1447 without osgi layer

claesrosell:

PR #1464, opened 2023-10-28
java.lang.NoClassDefFoundError when starting WebViewer #1451

claesrosell:
Added missing Import-Package: headers to "our" version of Axis 1.4.1

I do not know the state of the Axis in Orbit. It has the correct import statements but I do not know if it contains the fixes for the vulnerabilities mentioned in #662, which made us build our own version of Axis.

### DIFF
--- a/common/org.apache.axis/META-INF/MANIFEST.MF
+++ b/common/org.apache.axis/META-INF/MANIFEST.MF
@@ -76,6 +76,12 @@ Import-Package: javax.activation;resolution:=optional,
  javax.servlet;version="2.4.0";resolution:=optional,
  javax.servlet.http;version="2.4.0";resolution:=optional,
  javax.xml.rpc;version="1.1.0",
+ javax.xml.rpc.encoding;version="1.1.0",
+ javax.xml.rpc.handler;version="1.1.0",
+ javax.xml.rpc.handler.soap;version="1.1.0",
+ javax.xml.rpc.holders;version="1.1.0",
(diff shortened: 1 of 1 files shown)


merks:
That sounds related to https://github.com/eclipse-orbit/orbit-simrel/issues/16 where the discussion died.   

Where did birt’s version come from?  Was there an ip review associated it?  What exactly was fixed and where is what was done tracked?

claesrosell:
When I am reading through #662 I get the feeling that it was built from the sources of the Apache github repository https://github.com/apache/axis-axis1-java , which have had a lot of commits since the last release ( 2006-04-22). 

Why I want to address this now is that these missing Import-Package statements prevents the WebViewer from working. With them in place my local build of #1462 is working.

claesrosell:
... with that said: I rather use Axis from Orbit

merks:
Yes, I would love to have/get/find a newer version with issues CVE addressed, but from where to get it?   I don't see anything published on maven central based on commits in here:

https://github.com/apache/axis-axis1-java

That links to 

https://axis.apache.org/

And there one sees only axis2 stuff suggesting that it would be far better to use axis2...

merks (approved):
It looks like someone just dropped some new jar in there:

image

But where it was built, and who built it is a mystery.  A mystery that needs to be solved.

In the meantime, yes, there are package requirements missing that are present in the orbit version, so best we add those.

Issue #1455, opened 2023-10-23
An error occurs when creating an XLS/XLSX file


hypn0chka:
An error occurs when generating XLS/XLSX files:
 "java.lang.IllegalStateException: Cannot add merged region A2:E2 to sheet because it overlaps with an existing merged region (A2:E2).".

There was no such error in versions up to 4.8. It looks like there have been changes to the Spudsoft emitter.
Can you fix this error? I can't do this myself because I don't know Java.

[xls_xlsx_error.zip](https://github.com/eclipse-birt/birt/files/13068112/xls_xlsx_error.zip)

speckyspooky:
I have checked your test-report and the issue with the merged cells occure with version BIRT 4.6 XLSX-Spudsoft.
And with all later versions to, so it seems not a new bug this must be an older one.

(I tested BIRT 4.6, BIRT 4.8, BIRT 4.14 - all versions have the same behavior.)

speckyspooky:
The is a workaround (may be) because I checked your test-report and added a text to your empty-text (type: auto) and afterwards I was able to create your report:

![2023-10-23 20_25_21-workspace_rep - C__Users_gutma_Downloads_xls_xlsx_error_xls_xlsx_error rptdesign](https://github.com/eclipse-birt/birt/assets/41593722/23762bf4-bd14-45c4-8b51-cd0d92b9cef6)

The effect is only given on a row with 1 overspan of 2 cells and no text element or label is available with a value.

ivkina:
We are also having the same issue/error in some reports (others work fine). 

Could you please provide a little bit more details on possible workarounds? Is the issue fixed in the latest version of BIRT? We are using verison 4-10 (org.eclipse.birt.runtime_4.10.0-20220721.jar).

Here is the error:
Cannot add merged region B2:B3 to sheet because it overlaps with an existing merged region (A2:K2).
StackTrace:
org.apache.poi.xssf.usermodel.XSSFSheet.validateMergedRegions(XSSFSheet.java:486)
org.apache.poi.xssf.usermodel.XSSFSheet.addMergedRegion(XSSFSheet.java:413)
org.apache.poi.xssf.usermodel.XSSFSheet.addMergedRegion(XSSFSheet.java:367)
uk.co.spudsoft.birt.emitters.excel.StyleManagerUtils.extendRows(StyleManagerUtils.java:1082)
[...]

PR #1467, opened 2023-10-30
Fix excel-issue of removed row then remove also regsitered merge regions from sheet (#1455)

speckyspooky:
Fix the excel-issue if an blank row with merged cells will be removed then remove the registered merge area from the excel-sheet and also an additional double check is added to avoid overlapping merge registrations on the excel-sheet (#1455)


wimjongman (approved):
Looks good. Thank you.

Issue #1470, opened 2023-10-30
org.eclipse.birt.promote fails on master builds after move to Tycho 4.0.3


claesrosell:
https://ci.eclipse.org/birt/job/build/job/master/52/
I guess this was missed since promote is not done on non master builds?
Since we switched to Tycho 4.0.3 the default file extension has changed to tar.gz for Linux and Mac product archives (https://github.com/eclipse-tycho/tycho/issues/913)

merks:
You're awesome!  🥇

PR #1471, opened 2023-10-30
org.eclipse.birt.promote fails on master builds #1470

claesrosell:
Product archives for Linux and Mac have changed from zip files to tar.gz files. Currently org.eclipse.birt.promote fails since it still uses .zip for these files.


merks (approved):
Perfect!

PR #1476, opened 2023-11-01
Export packages with versions from the org.apache.axis bundle

merks:
This allows WTP's package requirements on org.apache.axis packages to resolve to the packages exported by BIRT's 1.4.1 version of that bundle.

https://github.com/eclipse-birt/birt/issues/1424


merks:
@claesrosell @speckyspooky 

This change eliminates some validation warnings when launching and allows WTP's package requirements (that have version constraints) to resolve to BIRT's 1.4.1 version of org.apache.axis.

image

I hope that for M3 WTP fully migrates to Jetty 12 so that those warnings will go away as well.

I also hope that we can eventually provide a 1.4.1 version in Orbit so that all of this can be removed...

speckyspooky:
@merks Sounds like a plan to have only one central axis version!

Issue #840, opened 2022-02-27
TextCombo used in JSEdtior is leaking graphic contexts (gc)


claesrosell:
Guilty code, the GC created is never disposed. Easy enough.
However, from the look of this code I think that we should reconsider how the bold font is retrieved and what the protected member "itemHeight" is used for.

org.eclipse.birt.core.ui.swt.custom.TextCombo:
```
	public TextCombo(Composite parent, int style) {
		super(parent, style);

		GC gc = new GC(this);
		itemHeight = gc.getFontMetrics().getHeight() + 2;

		if (gc.getFont().getFontData() == null || gc.getFont().getFontData().length == 0) {
			fontBold = new Font(Display.getCurrent(), "arial", //$NON-NLS-1$
					9, SWT.BOLD);
		} else {
			FontData fd = gc.getFont().getFontData()[0];
			fontBold = new Font(gc.getDevice(), fd.getName(), fd.getHeight(), fd.getStyle() | SWT.BOLD);
		}
		addDisposeListener(new DisposeListener() {

			public void widgetDisposed(DisposeEvent e) {
				fontBold.dispose();
				choiceMarkerMap.clear();
			}
		});
	}
```

claesrosell:
Closed by #1479

PR #1479, opened 2023-11-03
Fixes some UI strangeness in Birt Designer

claesrosell:
Long since, when I just wanted to do a small fix for #840 I ended up doing some substantial changes to some custom UI controls, since they typically didn't work well on Linux and probably not on any system where the font differs from Windows default, incl UI scaling.

This PR addresses them as well as some other resource leaks.

**TextCombo layout used on "Script tab"**
Before
image
After
image

**Chart style selectors**
Before
image
After
image

**Font selector**
Before
image
After
image


merks:
It definitely looks more usable! 👍

speckyspooky (approved):
Looks good to me and good point to enhance the UI of our other platforms.

PR #1481, opened 2023-11-03
Use batik instead of flute for CSS parsing

merks:
https://github.com/eclipse-birt/birt/issues/1480


merks:
@speckyspooky 

I assume there are no concerns with this simplification...

speckyspooky:
Yes, we can use the batik style parser. I check the flute-version on github and the latest changes of it are very old, so it make really sense to switch.

merks:
Exactly!   I was looking for an update and that screamed, use something else!  😱  It was a bit tricky to get it working, but in the end, just a few simple changes were necessary, so I'm personally happy with that...

speckyspooky:
Yes, I know what you mean. At the beginning so much research & try to get working and at the and the sum of changes looks sometime so small. A very good idea to move to the existing dependency of batik!

Issue #645, opened 2021-04-24
Disable multi-sheet generation in xlsx emitter


SteveSchafer-Innovent:
I have a client that wants to disable multi-sheet generation in the xlsx emitter.

Coincidentally, this bug was just created yesterday: https://bugs.eclipse.org/bugs/show_bug.cgi?id=573117

The spudsoft emitter recognizes a render option "SINGLE_SHEET", which looks like it will do this but I can't find any provision in the viewer to set that option.  It would be ideal if this option could be set globally in viewer.properties and also passed as a URL parameter such as __single_sheet=true/false or __xlsx_options=single_sheet/multi_sheet.

Please let me know if this capability already exists and I missed it, or if there's any reason not to implement it as I suggested.

hvonbargen:
I don't know, but if it doesn't work with BIRT now, one should have a look at the newer sources of the emitter in my fork hvbtup/BIRT.
And regarding the question in the bug, the 40 (rows) is the default page break for lists and/or tables, so it is correct if the emitter creates a new sheet every 40 records. Unfortunately the maximum value was 9999, maybe this is still the case.
The emitter options can usually be set as user properties, but probably also in other easy.  Unfortunately yaytay's bitbucket repository, which also contained documentation, is no longer available.
I'm pretty sure that *any* emitter option can be set in the URL query, but I don't use it myself, so I don't know exactly this works. Anyway, did you try the exact spelling SINGLE_SHEET=true?

speckyspooky:
Hello SteveSchafer, there is an option to set this behavior on report level.

If you add on the root element the user property "ExcelEmitter.SingleSheet" (Boolean) and set the value to "true" then you avoid multi-page sheets.
But you have to handle the repeat of header if it is active due to usage on PDF-site.

The spudsoft-Emitter has further additional options which can be set on report root tag based on user properties like "ExcelEmitter.RemoveBlankRows" or "ExcelEmitter.DisableGrouping".

SteveSchafer-Innovent:
I confirmed that speckyspooky's suggestion works.  Unfortunately, having to modify all their reports isn't a very good option for my client.  I'd still like to implement something at the viewer level.  I suspect one sheet per page is undesirable for most people, especially ones that tend to have large reports.  Being able to set the default value in viewer.properties would be great.

speckyspooky:
I'm back with new details. By default there isn't a way to configure the xlsx-Emitter-behavior based on the viewer.properties.
[...]

PR #1483, opened 2023-11-03
Enhance of the export viewer dialog to add the excel single sheet option (#645)

speckyspooky:
This is an enhancement to add the option to create a single sheet output directly with the viewer export based on the Spudsoft option  (#645)

The option will be displayed if the Spudsoft XLS or XLSX is selected.

### XLS
XLS

### XLSX
XLSX


speckyspooky:
@claesrosell
It would be nice if you could take a look into the small change of the dialog and the new single sheet output option. Thanks in advance.

Issue #1089, opened 2022-10-19
Decimal data type controls are not printing correctly - the data is getting cut off and rounded.


ivkina:
Create a simple data control of Decimal type in a report. Set control value to "1111111111111112.12". Print/preview report in Eclipse/View Report/Web Browser. See simple report design and report output attached. 
Note that the data is getting cut and rounded. If I remove 1 digit and make precision = 17 instead of 18, then it prints correctly. Is there as limit on BigDecimal precision data type in BIRT?
[decimal_not_printing.zip](https://github.com/eclipse/birt/files/9822528/decimal_not_printing.zip)
![decimal_not_printing_viewer](https://user-images.githubusercontent.com/17771801/196753132-77655160-32f0-4fdc-8094-68bf3fa13f21.png)
The issue happens in old BIRT 4.9 (Java 8) and new BIRT 4.9 (Java 11).

If you run/print the same number in Java 8 or 11 it prints correctly:
BigDecimal d = new BigDecimal("1111111111111112.12");
System.out.println(d);

hypn0chka:
Write me. I added an email to my profile.

claesrosell:
After debugging your example I draw this conclusion:
Javascript numeric literals is translated to double:s in Rhino (the javascript engine that Birt use). That double is later converted to a BigDecimal by Birt. It is the String to double convertion that looses the precision in this example.

ivkina:
Thanks for looking into that. In real life example we ran across this issue with total (aggregate) controls. And the totals (both table bindings and controls) are also defined as Decimal data type. Probably, the internal scripting is also doing double conversion in  between and then BIRT classes convert the end result into BigDecimal.
Is there any way to fix that or is that part of rhino scripting which is beyond the scope?

claesrosell:
After "reading up" on numbers in javascript they seem to be double precision floats, which means that the Rhino implementation is correct.

hvbtup:
If you are calculating invoices, it is a mistake to use `float` or `double` data types.
Instead, all calculations should be done with 10-based numbers in this case.

That's the reason why most programming languages (including ancient Cobol) also support data types like `java.lang.BigDecimal` which are able to store decimal values as such and to perform calculation. A conversion from a binary floating point number to a 10-based number and vice versa incurs a conversion error in many cases.

For example, take 31/10 = 3.1 (exactly).
[...]

PR #1484, opened 2023-11-04
Add DENSE_RANK aggregate function

claesrosell:
Implementation and unit tests of DENSE_RANK. I could not decide if I should change the file format version or not, I ended up not doing it.


data/org.eclipse.birt.data.aggregation/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -70,7 +70,7 @@ org.eclipse.jdt.core.compiler.problem.unusedParameterWhenOverridingConcrete=disa
 org.eclipse.jdt.core.compiler.problem.unusedPrivateMember=warning
 org.eclipse.jdt.core.compiler.problem.varargsArgumentNeedCast=warning
 org.eclipse.jdt.core.compiler.release=enabled
-org.eclipse.jdt.core.compiler.source=11
+org.eclipse.jdt.core.compiler.source=16
```
merks, line 73: It seems odd to set it to 16.

claesrosell: Yes. It was done by Eclipse quick-fix. I switched to 17 manually afterwards and missed one line.

merks:
For future reference, you can modify the Bundle-RequiredExecutionEnvironment in the MANIFEST.MF and save.  Then on the context menu of the plug-in project use Plug-in Tools -> Update Classpath... which will update the .classpath and the org.eclipse.jdt.core.prefs to match that value.

claesrosell:
@hvbtup I suspect that you have a lot on your plate. Still, I would like you to test this out since you suggested this new aggregation function and have a lot more knowledge about report design than me.

hvbtup (changes requested):
It works well if the expression is NOT NULL.
But the report fails with

java.lang.ClassCastException: class org.eclipse.birt.data.aggregation.impl.rank.NullObject cannot be cast to class java.lang.Comparable

if the expression can be null.

For example, the report crashes if I use the following query

```
select 
   PRODUCTCODE
 , PRODUCTNAME
 , PRODUCTLINE
 , PRODUCTSCALE
 , PRODUCTVENDOR
 , PRODUCTDESCRIPTION
 , QUANTITYINSTOCK
 , BUYPRICE
 , MSRP
 , case when productname like '%f%' then null else buyprice end as x
from products

```

and then use RANK or DENSERANK based on the column X.

A minor issue (but my German picky eyes stumble over it...) is the non-consistent spelling.
There are 
* DENSERANK (the visible form, which is OK, because it is in line with eg. RUNNINGCOUNT)
* TOTAL_DENSE__RANK_FUNC (with two underscores before RANK)
* DENSE_RANK

claesrosell:
With this latest commit everything should be addressed. I did some more clean-up to the RankObjComparator, added generic types to it.

hvbtup (approved):
It's looking and working much better now!

I think it's worth noting in the docs (if we only had good docs) how the BIRT aggregate functions RANK and DENSERANK treat NULL values:

They are considered to be less than all NON NULL values.

Thus with ascending order, NULLs come first, and with descending order, NULLs come last.

This behavior is the same as with the RANK aggregate in previous releases, that's good.

People with a strong DB-specific background (like me) may be surprised, though.
In Oracle, ORDER BY expr [ASC|DESC] NULLS LAST will put NULLs at the end *independent* of ascending or descending ordering. This SQL syntax is standard, but the default is vendor-specific. In Oracle, NULLS LAST is the default; to get the nulls first one has to specify NULLs FIRST explicitly.

However, this is vendor-dependent (see https://modern-sql.com/concept/null).

We might add explicit control over NULL values later.
A workaround (in SQL) is shown on the modern-sql page.
That principle can be used with BIRT as well by adding a boolean expression that tells us if the column is null or not; then we can first sort by this column and then by the column value itself.

claesrosell:
Today there are no way to pass parameters down to Aggregator functions / Accumulators. If we were to make that possible it would benefit #1089 as well. I might look into that one as my next thing

Issue #1385, opened 2023-08-02
Font aliasing not working for HTML text


hvbtup:
This isn't a new issue, but it's a serious blocker for migrating from running BIRT on Windows to running BIRT on Linux:

The font aliasing specified in fontConfig.xml does not work for HTML font attributes.

I'll try to explain this in detail:

The attached report output should show the text in the same style twice, but the second output using the builtin Times-Roman font instead of a subsetted Arial TrueType font as expected.

[20230802-3134388.pdf](https://github.com/eclipse-birt/birt/files/12239658/20230802-3134388.pdf)

This output was created from a report where I specify font name, size etc. as report parameters and then use different ways to dynamically generate formatted text:

![grafik](https://github.com/eclipse-birt/birt/assets/7644776/15942895-33bc-45ad-a244-28f5717c0c1b)

onCreate event of the first item (row["TEXT"):

    this.getStyle().fontFamily = row["SCHRIFTART"];
    if (row["FETT"] == "J") this.getStyle().fontWeight = "bold";
    if (row["KURSIV"] == "J") this.getStyle().fontStyle = "italic";
    this.getStyle().fontSize = row["GROESSE"];

HTML source of the second item:

    "<span style='font-family:" + row["SCHRIFTART"] + (row["FETT"]=="J"? "; font-weight:bold" : "") + (row["KURSIV"]=="J"? "; font-style:italic" : "") + "; font-size:" + row["GROESSE"] + "'>" + row["TEXT"] + "</span>"

BIRT is running on Linux (in a Podman container, to be precise), with no Arial TrueType font available, but we have Liberation Sans available:
![grafik](https://github.com/eclipse-birt/birt/assets/7644776/e6abbbae-1b01-40df-a699-27e3c3da9daf)
And we have defined font aliasing in the `fontConfig.xml` file:

    ...
	<font-aliases>
		<mapping name="serif" font-family="Times-Roman" />
		<mapping name="sans-serif" font-family="Helvetica" />
		<mapping name="monospace" font-family="Courier" />
		<mapping name="cursive" font-family="Times-Roman" />
		<mapping name="fantasy" font-family="Times-Roman" />
		<mapping name="Arial Narrow" font-family="Liberation Sans Narrow" />
		<mapping name="Arial Black" font-family="Roboto Black" />
		<mapping name="Arial" font-family="Liberation Sans" />
		<mapping name="Times New Roman" font-family="Liberation Serif" />
[...]

PR #1485, opened 2023-11-04
Fix the font alias name to lower case handling (#1385)

speckyspooky:
Fix the font alias name to lower case handling (#1385)

### DIFF
--- a/engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/layout/pdf/font/FontMappingConfig.java
+++ b/engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/layout/pdf/font/FontMappingConfig.java
@@ -21,41 +21,65 @@ import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
 
+/**
+ * Font mapping configuration class
+ *
+ * @since 3.3
+ *
+ */
 public class FontMappingConfig {
 
-	protected Set fontPaths = new HashSet();
+	protected Set<String> fontPaths = new HashSet<String>();
 
 	/** The font-family replacement */
-	protected HashMap fontAliases = new HashMap();
+	protected HashMap<String, String> fontAliases = new HashMap<String, String>();
 
 	/** The encoding for the fonts */
-	protected HashMap fontEncodings = new HashMap();
+	protected HashMap<String, String> fontEncodings = new HashMap<String, String>();
 
 	/** the global sequences defined for composite fonts */
-	protected HashMap searchSequences = new HashMap();
+	protected HashMap<String, String[]> searchSequences = new HashMap<String, String[]>();
 
 	/**
 	 * composite fonts is constructed by multiple physical fonts which may cover
 	 * large amount of glyph
 	 */
-	protected HashMap compositeFonts = new HashMap();
+	protected HashMap<String, CompositeFontConfig> compositeFonts = new HashMap<String, CompositeFontConfig>();
 
+	/**
+	 * Constructor
+	 */
 	public FontMappingConfig() {
 	}
 
+	/**
+	 * merge the font configuration to the existing font setup
+	 *
+	 * @param config font mapping configuration
+	 */
 	public void merge(FontMappingConfig config) {
 		fontPaths.addAll(config.fontPaths);
-		fontAliases.putAll(config.fontAliases);
+
+		// fontAliases.putAll(config.fontAliases);
+
+		// merge alias fonts, special handling in addFontAlias()
+		Iterator<?> iterAlias = config.fontAliases.entrySet().iterator();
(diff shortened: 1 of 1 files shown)


engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/layout/pdf/font/FontMappingConfig.java
```diff
@@ -21,41 +21,65 @@
...
+	/**
+	 * merge the font configuration to the existing font setup
+	 *
+	 * @param config font mapping configuration
+	 */
 	public void merge(FontMappingConfig config) {
 		fontPaths.addAll(config.fontPaths);
-		fontAliases.putAll(config.fontAliases);
+
+		// fontAliases.putAll(config.fontAliases);
```
claesrosell, line 63: Remove this, we have the old behavior in the GIT history.

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/layout/pdf/font/FontMappingConfig.java
```diff
@@ -21,41 +21,65 @@
...
+
+		// merge alias fonts, special handling in addFontAlias()
+		Iterator<?> iterAlias = config.fontAliases.entrySet().iterator();
```
claesrosell, line 66: While I haven't anything against while loops over iterators, I would prefer something like this instead. Since the generic types you introduced in this PR are available.
```
	for( Entry<String, String> fontAliasEntry : config.fontAliases.entrySet()) {
		String fontAlias = fontAliasEntry.getKey();
		String fontName = fontAliasEntry.getValue();
		this.addFontAlias(fontAlias, fontName);
	}
```
Or a short, short version. Which probably would need a small comment
```
	for( Entry<String, String> fontAliasEntry : config.fontAliases.entrySet()) {
		this.addFontAlias(fontAliasEntry.getKey(), fontAliasEntry.getValue());
	}
```

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/layout/pdf/font/FontMappingConfig.java
```diff
@@ -21,41 +21,65 @@
...
+			this.addFontAlias(fontAlias, fontName);
+		}
 		fontEncodings.putAll(config.fontEncodings);
 		searchSequences.putAll(config.searchSequences);
 
 		// merge the composite fonts
-		Iterator iter = config.compositeFonts.entrySet().iterator();
-		while (iter.hasNext()) {
-			Map.Entry entry = (Map.Entry) iter.next();
+		Iterator<?> iterComposite = config.compositeFonts.entrySet().iterator();
```
claesrosell, line 77: Same here

claesrosell (commented):
I just have some small comments. I read up on the issue itself and it seems like you guys are on top of the problem and the solution to it.

I think that it is hard to give code comments in a project like Birt. It is a lot of old code in Birt, and it is always a struggle to decide what one should change. Typically one wants to keep the changes small, but still wants to improve/modernize the code. If think that you have found a good balance here. Still, here are my suggestions for what I think would be an improvement.

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/layout/pdf/font/FontMappingConfig.java
speckyspooky: I like this and will change it.

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/layout/pdf/font/FontMappingConfig.java
speckyspooky: I will change it to.

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/layout/pdf/font/FontMappingConfig.java
speckyspooky: Nor problem, I will remove it.

speckyspooky:
@claesrosell
Thanks for your comments, I have done the changes. Of course sometimes it is really hard to change the old stuff. I like your suggestion of the looping to avoid the iterators.
At the end the change include only to change the alias names to lower case according to the font-alias-handling on the text areas (according to the original issue).

claesrosell:
Excellent!  Go ahead and commit.

hvbtup:
Thanks, Thomas!

PR #1490, opened 2023-11-07
Replace BIRT's axis copy with the new one from Orbit

merks:


merks:
@speckyspooky  @claesrosell 

Orbit now produces a 1.4.1 version of org.apache.axis that is directly built from the latest GitHub source.  That's bundle is being used in this PR and I've deleted BIRT's copy of it.  All looks fine in terms of the build and the tests, so I assume we're good to merge this change, right?

speckyspooky:
I have seen your changes and it sounds good and looks good too.
So from my side it is good to merge.

merks:
I hope that WTP makes good progress for M3 so that we can further simplify and update the target platform.

It will probably be a good idea to test the runtimes to make sure they are still complete...

PR #1493, opened 2023-11-10
Regenerate the target platform after DataTools changes

merks:
This also fixes the strange java.xml.namespace.QName compile error though why that's the case isn't clear.

https://github.com/eclipse-birt/birt/issues/1492

### DIFF
--- a/build/org.eclipse.birt.target/org.eclipse.birt.target.target
+++ b/build/org.eclipse.birt.target/org.eclipse.birt.target.target
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <?pde version="3.8"?>
-<target name="Generated from BIRT" sequenceNumber="25">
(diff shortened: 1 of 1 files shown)


claesrosell:
I searched the web for "net.sourceforge.lpg.lpgjavaruntime". Got a hit on the maven repository, turns out that it is an old Birt artifact:  https://mvnrepository.com/artifact/org.eclipse.birt.runtime/net.sourceforge.lpg.lpgjavaruntime/1.1.0.v201004271650
It has been part of the auto-generated target platform for six months. I wonder how that ended up in the target platform in the first place and why was it removed now?  Strange.

merks:
Well, that part is clear, if only to me.  😜 

It's gone because I got rid of it:

https://github.com/eclipse-datatools/datatools/issues/13

It was some ancient thing the origin for which (CVS-based orbit) has disappeared.  Kind of scary...

PR #1494, opened 2023-11-11
Update to the latest derby

merks:


merks:
The whole derby thing in maven looks to be a huge mess with mistakes and missing sources.  I don't have the resource for cleaning such a mess at this point in time.

Issue #1525, opened 2023-12-18
NPE in SvgFile.isSvg(String)


ilnur-z:
Where is a NullPointerException on method org.eclipse.birt.report.engine.util.SvgFile.isSvg(String) then uri is null.
I think that because some brackets needed:

`
if (uri != null && ( uri.endsWith(".svg") || uri.toLowerCase().contains(URL_IMAGE_TYPE_SVG) ) ) 
`

(just like on method org.eclipse.birt.report.engine.util.SvgFile.isSvg(String, String, String))

caused by https://github.com/eclipse-birt/birt/commit/59c1d03895244d437e1153fd1ecd0e6e3ffb4b47

wimjongman:
Please add the stacktrace

ilnur-z:
here stacktrace

ilnur-z:
ImageManager2 is our little modified ImageManager, which not takes rasterised data for svg-images

speckyspooky:
Yes, I can confirm the NPE is based through the missing brackets.
I added a PR #1526 some minutes ago to fix it.

PR #1526, opened 2023-12-18
Fix the issue with null-reference of uri in SvgFile.isSvg() (#1525)

speckyspooky:
The method isSvg(String) needs additional brackets to avoid a possible NullPointerException.


wimjongman:
The OP mentioned that this bug was located in two files.

wimjongman (approved):
LGTM. Thanks.

PR #1539, opened 2024-01-10
Fix data item leading whitespace preserve #1536

hvbtup:
This fixes #1536. Note that preserving leading whitespace only works for data items, but not for dynamic text items.


speckyspooky (approved):
Review done and the issue will be fixed with it.

Issue #1547, opened 2024-01-19
BIRT 4.14 creates an indent for grids in footer & header


reiner-killet:
Hi team,
I tested the last release of 2024-01-12. PDF is looking pretty good!
But regarding DOCX there is a change in behavior compared to 4.8:
We use grids in header & footer for easier positioning of pictures, page number etc.
In 4.8 the footer fits to the main area. See screenshot `No indentation of header and footer in 4 8 output DOCX.png`.
In 4.14 the footer moves to right. See screenshot `Indentation of header and footer in 4 14 output DOCX.png`.
I examined the footer and found out that BIRT creates a nested grid.
The inner grid results from my template.
The outer grid is added by BIRT.
The difference is that in 4.8 the outer grid hat a negative left indent and the inner grid had a positive indent of same amount.
In 4.14 the outer grid has no indent and the inner grid has a positive indent.
Next, I found a workaround:
The `Simple Master Page` has 2 properties in the `Advanced` slider:
- `Columns` which is set to `1 : inherited` as default.  
- `Column spacing` which is set to `12pt : inherited` as default.
When I change one of these values (`Columns` to `3`, or `Column spacing` to `0pt`) 4.14 behaves like 4.8.
Thereto, see screenshots `Indentation of header and footer in 4 14 - Workaround effect in XML.png` and `Indentation of header and footer in 4 14 - Workaround in designer`.
I do not understand the meaning of `Columns`. Set to `3` BIRT does not create 3 columns. 
Can you please check if this is a bug or maybe a feature somewhere else?
In my opinion it is a bug because a template with only default values should not lead to different indents in header, footer, and main area.
Thanks, Reiner
  
![Indentation of header and footer in 4 14 - Workaround effect in XML](https://github.com/eclipse-birt/birt/assets/95427074/214f2332-dc75-46c4-8313-d19008859299)
![Indentation of header and footer in 4 14 - Workaround in designer](https://github.com/eclipse-birt/birt/assets/95427074/b0d5483e-1f0d-4378-b2b8-64c9a750bb10)
![Indentation of header and footer in 4 14 output DOCX](https://github.com/eclipse-birt/birt/assets/95427074/f5889ec9-50ae-437f-9191-f3b32e9871e9)
![No indentation of header and footer in 4 8 output DOCX](https://github.com/eclipse-birt/birt/assets/95427074/e0ae5577-251e-4199-9234-52c24b290160)

reiner-killet:
[...]

PR #1552, opened 2024-01-24
Fix the docx issue with the unset margin of the header-footer-container (#1547)

speckyspooky:


speckyspooky:
I created a new method "writeTableCellMarginZero()" which create only the zero-margin-tag of docx.
The method will be called from "startTable()" (like before as single instructions) and new is the call of the new method through "startHeaderFooterContainer()".

### Header w/o margin, footer with margin
example-header-wo-margin-footer-w-margin

wimjongman (approved):
Thanks, Thomas!

PR #1561, opened 2024-02-04
Fix osgi service viewer issue, added copy actions on build.xml (only OSGi-Version)

speckyspooky:


releng/maps/orbit_bundles.map
```diff
@@ -5,7 +5,7 @@ plugin@javax.xml,1.3.4=GET,http://download.eclipse.org/tools/orbit/downloads/dro
 
 plugin@javax.xml.rpc,1.1.0=GET,http://download.eclipse.org/tools/orbit/downloads/drops/R20130517111416/repository/plugins/javax.xml.rpc_1.1.0.v201209140446.jar,unpack=true,usetimestamp=true
 
-plugin@javax.xml.soap,1.2.0=GET,http://download.eclipse.org/tools/orbit/downloads/drops/R20130517111416/repository/plugins/javax.xml.soap_1.2.0.v201005080501.jar,unpack=true,usetimestamp=true
+plugin@javax.xml.soap,1.3.0=GET,http://download.eclipse.org/tools/orbit/downloads/drops/R20130517111416/repository/plugins/javax.xml.soap_1.3.0.v201105210645.jar,unpack=true,usetimestamp=true
```
merks, line 8: FYI, I can find no reference to this file.   I'm doubtful this is used in the build.  I can find these in the workspace (in target folders produced by the Tycho build):

image

That's also what's in the target platform:

image

So it's not clear if this change has any impact at all. I doubt it.

speckyspooky: Of course I think I will haven't an effect. The problem is that we need the version 1.3.0 in the lib-folder or the saaj.jar.
But the first try with saaj.jar crashed, the next try crashed at this moment to.

How can we get 1.3.0 or saaj.jar to the lib-folder?

Habe you an idea?

merks: I see it here:

image

It looks like it's already in the lib folder...

speckyspooky: It should be but in the latest build it isn't there for the OSGi-version.
(The standard-runtime has all libraries in the lib-folder and that is why it is running.)

speckyspooky: Have you and idea why the run was cancelled there is no reason listed at the log file.

merks:
I know other people love their GitHub actions, but I'm not one of those people.  My internet here in Spain right now is not the most stable and not so fast, so scrolling to the end of one of these pages is painful.

Locally my build fails like this so maybe there is something wrong with the current I-Build.

> 
> [WARNING]  Parameter 'qualifier' (user property 'buildQualifier') is read-only, must not be used in configuration
> [ERROR] An error occurred while transferring artifact canonical: binary,org.eclipse.equinox.executable_root.cocoa.macosx.aarch64,3.8.2400.v20240129-1338 from repository https://download.eclipse.org/eclipse/updates/4.31-I-builds/I20240202-1800:
> [ERROR]    download from https://download.eclipse.org/eclipse/updates/4.31-I-builds/I20240202-1800/binary/org.eclipse.equinox.executable_root.cocoa.macosx.aarch64_3.8.2400.v20240129-1338 failed

Have you tried the new launcher so that you can make progress more quickly locally and can see what file exist where locally.  That seems to me will make things progress 10 times faster.

In any case, it's Sunday and I'm calling it a day, so probably I can't help much more today. Sorry.

speckyspooky:
I fetched your changes but the new Build-option won't be displayed.
So I stuck completely, but that's life.

merks:
I'm about to go out now, but I hate to leave you suck...

So if you open this:

image

Is there not even an m2e (Maven Build) category?

image

I assume m2e is installed because the setup installs it and once launch is in your workspace, it should appear in this group...

image

speckyspooky:
Yes, I confirm, there is no entry at the menu-side and also into the "Run configuration" is no entry listed.

merks:
Can you paste a capture of the full contents of the left viewer in the run configurations dialog?

speckyspooky:
The "BIRT Tycho Build 2" is my own try to copy the details manually from your screen.

tycho

merks:
The category is there. You’re sure the launch file is really in the workspace? Ctrl-shift-R should show it.

speckyspooky:
I thought when I fetch your changes to my branch and I saw your PR locally as I fetched the changes then it would be directly available.

By the way, your screen of "saaj.jar" shows different target locations.
Do you know where it is defined because I didn't find a configuration reference of saaj.jar?

And where is the instruction of "javax.xml.soap_1.2.0 located that it will be unpacked.
The problem of my first try was that the unpacked-folder doesnt exists but on the "normal" runtime it will be placed directly in the WebViewerExample\WEB-INF\lib - Where is it configured to get an idea for modification?

At this point, many thanks for your support!

merks:
Back from dinner...

I expect this would be in your workspace:

image

And that would make the launch appear in the group.

It appears that the 1.2 version of soap has a lib folder with that jar (in my ~/.p2/pool):
```
$ls *soap*
javax.xml.soap_1.2.0.v201005080501:
META-INF/  about.html  about_files/  lib/  plugin.properties

javax.xml.soap_1.3.0.v201105210645:
META-INF/  about.html  javax/  plugin.properties

$ls *soap*/lib
saaj.jar
```

A bundle can have an instruction to say the shape is "dir" and then it will be unzipped rather than remaining a jar after it is downloaded by p2.  That appears to be what happens for soap.

So when/where all the jars and folders are copied to the lib folder, one might try also **/lib/*.jar to copy any nested jars from lib folders of unzipped jars...

speckyspooky:
Your last sentence got me the right hint and I added it to the build.xml of the OSGi-version and it works!
I checked the artifact of this PR and the OSGi-version is now working again :o)

I agree with your Tycho-target, the new option should be given but currently not.
I will double check it.

PR #1567, opened 2024-02-20
Use org.mozilla.rhino instead of org.mozilla.javascript

merks:
https://github.com/eclipse-birt/birt/issues/1566

### DIFF
--- a/UI/org.eclipse.birt.report.debug.core/.classpath
+++ b/UI/org.eclipse.birt.report.debug.core/.classpath
@@ -1,7 +1,15 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
-	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-11"/>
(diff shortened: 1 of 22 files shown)


UI/org.eclipse.birt.report.debug.core/.classpath
```diff
@@ -1,7 +1,15 @@
...
-	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
-	<classpathentry kind="src" path="src/"/>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-11">
+		<attributes>
+			<attribute name="module" value="true"/>
+		</attributes>
+	</classpathentry>
+	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins">
+		<accessrules>
+			<accessrule kind="accessible" pattern="org/mozilla/javascript/debug/**"/>
```
merks, line 10: It looks like Tycho is not respecting this information that makes package accessible.

merks: I guessed correctly that Tycho 4.0.6 fixes this problem, because I discussed this "access rule" issue with Christoph not that long ago.

Issue #1576, opened 2024-02-29
Font aliasing case-insensivity still not quite working


hvbtup:
Thomas,

once I wrote the font names in lower-case in my fontsFonfig.xml, font aliasing did not work as expected despite PR #1485.

In my fork, I had to add a tiny change, please see 
https://github.com/triestram-partner/birt/commit/5383d9c545c10435d94012ea95723431443ce19a

hvbtup:
I'm not absolutely sure if this actually an issue in BIRT, but I think it my changes are correct, to really get case-insensivity.

I stumbled about the issue not in BIRT itself, but in my PDFPostProcessor, which I use to add page numbers etc. *after* BIRT generated the PDF, and which reuses the font configuration of BIRT to some extent.

speckyspooky:
PR #1578 created to add this option finally

speckyspooky:
The change is pushed to the master.

PR #1592, opened 2024-03-12
Enhance font alias mapping #1576

speckyspooky:


speckyspooky:
This PR won't be published - github effect.

Issue #1588, opened 2024-03-11
BIRT 4.15 milestone: eclipse-designer (all-in-one) shows error on 1st report and the report will be hidden after restart (eclipse-restart-option)


speckyspooky:
I tested the all-in-one-designer BIRT 4.15 milestone (Windows). Directly on the creation moment of a new empty report the eclipse-error-log shows an internal error on designer level because an object is null and the null should be converted to an integer value.

If I use the eclipse-restart (main-menu)-option then the new report is hidden not only blank, the report isn't drawn
with the same error message.

The problem is located at EditorRulerLayout at line number 57. The main problem is that I cannot reproduce the issue on my dev-environment.
I have the issue only with the built 4.15 milestone (tested on 2 different PCs) and after sync with master also at my fork but again nit on my dec-env.

### Class: EditorRulerLayout.java, Line 57
https://github.com/eclipse-birt/birt/blob/master/UI/org.eclipse.birt.report.designer.ui/src/org/eclipse/birt/report/designer/internal/ui/editors/rulers/EditorRulerLayout.java
![EditorRulerLayout](https://github.com/eclipse-birt/birt/assets/41593722/4368d155-373a-4bd8-987d-b2018cb1547e)

### Error 01, report creation
![error-01-create-report](https://github.com/eclipse-birt/birt/assets/41593722/98b847ca-e436-448b-b830-3dee6a4db913)

### Error 02, restarted eclipse-designer
![error-02-restart](https://github.com/eclipse-birt/birt/assets/41593722/519876f6-acc0-4b22-940a-b9101ff7c2fa)

speckyspooky:
The issue is solved with PR #1593

PR #1593, opened 2024-03-12
Fix issue with NullPointerException due to Integer-cast (#1588)

speckyspooky:


merks (approved):
It seems like the only thing to do.

Issue #1595, opened 2024-03-13
OSGi-version: BackgroundImageInformaton, could produce "NoSuchElementException" due to unchecked iterator-element


speckyspooky:
With testing of the BIRT-OSGi-version 4.14 was figured out an issue with the method "BackgroundImageInformation.getImageResolution()".

The image-fetch based on an iterator isn't checked with the "hasNext()"-method and therefore an exception will be thrown if no elemnt is listed - see screen below. This case should be fetched accordingly. 

![osgi-background](https://github.com/eclipse-birt/birt/assets/41593722/7669f5b3-6528-45aa-9613-2d0ac79c1e57)

speckyspooky:
Fix suggestion provided under PR #1596

speckyspooky:
Fix is merged to the master branch.

PR #1596, opened 2024-03-13
Fix issue with unchecked iterator-element (#1595)

speckyspooky:


speckyspooky:
@merks
Good morning Ed, yesterday evening I found an "issue" with image handling. Normally the problem isn't a real problem because the exception of my ticket will be catched.

The topic of my change is that I try to avoid the "exception" because the "e.printStackTrace()" procude a very huge delay in summary on my side. 1 exception-trace 0.2s and my report hase lot of images > 50images. So I will get a delay of 10s (terrible).

Question 1:
Make the "e.printStaceTrace" really sense?

Question 2:
Can I directly deactivate the output before we have the change and BIRT 4.15?

Issue #1604, opened 2024-03-24
BIRT 4.15: SVG images won't be displayed on HTML-preview which are embedded on TEXT-element (type HTML)


speckyspooky:
On my regression tests I found an issue with SVG-images which are embedded directly into a TEXT-element with type HTML.
The images won't be displayed longer on the HTML-preview.
The issue is given with BIRT 4.15 "nightly build" (and further versions of 4.15 2024).

BIRT 4.14 is correct and also the first BIRT 4.15 of December 2023 are fine.
I don't know which change cause the behavior. The problem is that till BIRT 4.14 the embedded SVG-images was directly embedded into the HTML-editor and the preview works fine.
The current 4.15 versions externalize the svg as file and the sources will be displayed like broken image.

As cross check, the output of PDF & DOCX works correct.

### BIRT 4.15 nightly build
![images415](https://github.com/eclipse-birt/birt/assets/41593722/0645df49-fc05-4eaa-9b53-ebe79d4ce9ed)

### BIRT 4.14 
![images414](https://github.com/eclipse-birt/birt/assets/41593722/b12b7d63-6305-4767-b2f9-a9bff71dbf60)

[image_issue_img_src.zip](https://github.com/eclipse-birt/birt/files/14735969/image_issue_img_src.zip)

merks:
If you do Open Image in New Tab it shows the content:

![image](https://github.com/eclipse-birt/birt/assets/208716/827c5c19-7a2a-4b98-8cdd-5d8c4081c984)

If I copy the content into a file, including adding the XML header, it still renders like this:

![image](https://github.com/eclipse-birt/birt/assets/208716/a4baaad0-ecfb-43c2-9494-aae3c3d668d9)

Only if the file extension is .svg does it render correctly:

![image](https://github.com/eclipse-birt/birt/assets/208716/afcf9aa9-00ea-4036-904c-c09ae9d9a413)

speckyspooky:
Yes, of course with that extension it works. I have done a small change at 2 classes which checks "external" sources
and added the "data:"-URL-protocol key and the first try shows the old behavior where the svg-data-URL will be embedded.

It seems to me that the behavior is changed due to the changed tidy-parser behavior and therefore we got some effects.

I will take a look into the file extension way because there we would need a special role to get the extension from the URL.

speckyspooky:
@merks 
I need your mind because I have 2 ways to solved it but which one is the better way (?)

[...]

PR #1605, opened 2024-03-24
fix issue with embedded svg images added like data-URL (#1604)

speckyspooky:
fix issue with embedded svg images added like data-URL (#1604)


merks (approved):
Looks good. Nice to see cleaner code and helpful comments too.

PR #1606, opened 2024-03-25
Remove the RCP product

merks:
Remove other unused build infrastructure:

- birt-integration-test
- birt-publish


merks:
@speckyspooky 

The RCP product does not work well and I can see no good reason to maintain it.  It's only slightly smaller than the all-in-one package:

image

Given @wimjongman  already expressed is doubt about it's value and utility, removing it seems like a good idea...

wimjongman (approved):
LGTM!

speckyspooky (approved):
I'm fine with it.

hvbtup:
+1 for removing.
I never used the RCP designer and always teached using the All-in-One designer.

There's only one aspect for which RCP designer is superior compared to the All-in-One: Simplicity.
The All-In-One IDE intimidates newcomers, because it is overloaded with a lot of things that BIRT report designers actually do not need. But most of that comes in the form of of the icon bar and the menus, so after a few hours/days when newcomers desparately look for the whatsitsname menu, they learn where it is and to ignore all the superfluous menu entries.

Most of the superfluous menu entries etc are probably there because the original developers thought that BIRT developers are Java developers at the same time (which is seldom true from my experience), and so they thought that the BIRT IDE should be a Java IDE at the same time.
It would be great if there was an optional IDE "setup" (or whatever its called) that removes all this Java related stuff from the UI, leaving only what's necessary for BIRT and Javascript.

Issue #1620, opened 2024-04-03
DOCX emitter doesn't set language for correction


doortokaos:
MS Word has functionalities to help the user writing correctly in a chosen language.
Since I use a German Version of MS Word, I don't know exactly the name in the English UI, but I think it might be called something like "Language for correction assistance". Here a screen of the menu in my word:
![grafik](https://github.com/eclipse-birt/birt/assets/22446825/ae4bd580-3ad3-4c0b-a7b2-88cd9a86a8fa)

The language is always set to "English (US)" no matter in which language I create the report:
![grafik](https://github.com/eclipse-birt/birt/assets/22446825/1995ea51-76d5-49bf-b498-2b7f906c0adf)

The report I used for the example is this (remove the .txt 😉):
[korrekturhilfe.rptdesign.txt](https://github.com/eclipse-birt/birt/files/14853989/korrekturhilfe.rptdesign.txt)

I generated the DOCX with the German locale active in the designer and no further settings on the report or eclipse.
![grafik](https://github.com/eclipse-birt/birt/assets/22446825/563eda08-0518-4237-8f1d-7e30b7e0e1a3)

The designer is 4.15 Release all in one eclipse.

It would be nice, if the emitter could set the "Sprache für die Korrekturhilfe" to the language in which the report was created. So the user of the DOCX report won't have to set it every time manually after creating the report.

doortokaos:
@hvbtup thanks for the extensive reply.
It all makes sense and is more complicated as it seems to be.
I already thought that it can't be that easy because it would have been done already but I couldn't find information why it wasn't done already.

I wouldn't use the locale of the OS running BIRT as well, but I think using the locale in which the report is generated is a bit better than using nothing or defaulting to en_US. Since the user or report creator already set a locale for the complete report.

Since the perfect solution with a language for each text in the report seems rather complex, using a "better" locale for the complete document would be a step in the right direction in my opinion.

In my experience and user environment most documents contain only one language, so this could be an improvement, while not being perfect.

What do you think about this?

speckyspooky:
Yes, the tickets will be read :o)

The solution to verify if we could set the property for the whole documents sounds good to me.
The special thing is more a technical thing because the implementation of the docx-version is a mixture of a central library and own written source (from the original developers).
Therefore we cannot use the library-api directly and so we need a research to figure out is there a cetral property on document level.
(The latest MS Word versions support the language on document level and on paragraphs & tables.)

[...]

PR #1627, opened 2024-04-05
Enhancement to add the document language to DOCX based on the report locale (#1620)

speckyspooky:
Enhancement to add the document language to DOCX based on the report locale (#1620)

### DIFF
--- a/engine/org.eclipse.birt.report.engine.emitter.docx/src/org/eclipse/birt/report/engine/emitter/docx/writer/Document.java
+++ b/engine/org.eclipse.birt.report.engine.emitter.docx/src/org/eclipse/birt/report/engine/emitter/docx/writer/Document.java
@@ -51,9 +51,10 @@ public class Document extends BasicComponent {
 	private int mhtId = 1;
 
 	private int wordVersion;
+	private String documentLanguage = "";
 
 	Document(IPart part, String backgroundColor, String backgroundImageUrl, String backgroundHeight,
-			String backgroundWidth, boolean rtl, int wordVersion) throws IOException {
+			String backgroundWidth, boolean rtl, int wordVersion, String documentLanguage) throws IOException {
 		super(part);
 		this.backgroundColor = backgroundColor;
 		this.backgroundImageImgUrl = backgroundImageUrl;
@@ -61,6 +62,7 @@ public class Document extends BasicComponent {
 		this.backgroundWidth = backgroundWidth;
 		this.wordVersion = wordVersion;
 		this.rtl = rtl;
+		this.documentLanguage = documentLanguage;
 		writeStylesPart();
 		writeSettingsPart();
 	}
@@ -95,7 +97,7 @@ public class Document extends BasicComponent {
 			stylesPartWriter.attribute("w:cs", "Times New Roman");
 			stylesPartWriter.closeTag("w:rFonts");
 			stylesPartWriter.openTag("w:lang");
-			stylesPartWriter.attribute("w:val", "en-US");
+			stylesPartWriter.attribute("w:val", this.documentLanguage);
 			stylesPartWriter.attribute("w:eastAsia", "zh-CN");
 			stylesPartWriter.attribute("w:bidi", "ar-SA");
 			stylesPartWriter.closeTag("w:lang");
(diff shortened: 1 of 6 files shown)


engine/org.eclipse.birt.report.engine.emitter.docx/src/org/eclipse/birt/report/engine/emitter/docx/writer/Document.java
hvbtup, line 1: I don't understand the two dozens of deleted lines here. They seem unrelated to the language change.

hvbtup, line 1: And I'd prefer (only if MS Word accepts it) if it was possible to *not* set a language, e.g. skip the whole `w:lang` element, instead of using the old `en-US` default (which is a minor bug IMO).

engine/org.eclipse.birt.report.engine.emitter.wpml/src/org/eclipse/birt/report/engine/emitter/wpml/AbstractEmitterImpl.java
hvbtup, line 1: I assume that the `locale` property inside the rptdesign is already used to initialize `reportContext.getLocale()` ?

hvbtup (commented):
The PR seems to contain an change in one file (see my comments). Can you explain or remove this deletion block?

Personally I think the default en-US should be changed to if possible, but that's a matter of taste, so since you wrote the code, you may decide yourself.

And one question regarding using `reportContext.getLocale()` as the source. It seems to work, but at first glance I couldn't find where the rptdesign's locale property is used to initialize the reportContext's locale.

engine/org.eclipse.birt.report.engine.emitter.docx/src/org/eclipse/birt/report/engine/emitter/docx/writer/Document.java
speckyspooky: The deleted code is clean-up of old commented code.

speckyspooky: I will not change the structure over all because we can get effects if we remove the "w:lang" over all.
In my understanding this tag is a default tag of the styles of Word. Therefore I will not remove them.

hvbtup: OK.

engine/org.eclipse.birt.report.engine.emitter.wpml/src/org/eclipse/birt/report/engine/emitter/wpml/AbstractEmitterImpl.java
speckyspooky: The "locale"-property is a new one and will be initialized in the "public void initialize()"-method.
This was a local variable before and I use this one for the language. So I avoid special handling if the local is null and use the same behavior.

hvbtup: Yes, I understood that already. My question is from the perspective that we said we want to use the locale property *from the rptdesign file* as the source to set the language inside the DOCX file. 
And in the PR I see that we use the locale property from `reportContext` as the source instead.
So there's kind of a missing link from "rptdesign.report.locale" to "reportContext.locale", which I assume is existing somewhere in the existing source (outside of the PR). but I couldn't find it. I only want to be sure.

speckyspooky: Ohh, well, this is a good question. On of my test cases was to enter a local on report side "test-locale" & "auto"
and I got "test-locale" and "auto" internally which will be switched to "en" because these are no valid locale-values.
I tested also the option of "blank" locale and the result was the fallback "en" again.

My test environment is MS Windows 11 "de". So the system-locale wasn't used on my side.
But I cannot say directly where the transfer will be done on code level.

hvbtup: OK.

speckyspooky:
Is your "OK" the approval for that change then I would merge the change.

hvbtup:
Yes, pls merge.

PR #1628, opened 2024-04-07
Suppress Maven javadoc generation in the IDE

merks:
https://github.com/eclipse-birt/birt/issues/1618


merks:
@SteveSchafer-Innovent @speckyspooky 

I learned today that one can suppress maven executions that don’t work for the ide.  For me this eliminates the maven errors in the ide.

speckyspooky:
Sounds good!

Issue #1633, opened 2024-04-17
Performance Bottleneck


giangianoulas:
Hello

After running a huge report (pdf) on a customer witha very large dataset, using a custom plain file driver, we get OutOfMemmory.
While trying to figure out this one i got a java flight recording that showed a performance issue
![image](https://github.com/eclipse-birt/birt/assets/21333067/8bbddd3a-a067-49de-abcc-ebe00373b8a1)

looking at 
```org.eclipse.birt.data.engine.impl.CancelManager```

```java
	public CancelManager() {
		cancellableList = new ArrayList<>();
	}

	/**
	 *
	 * @param cancellable
	 */
	public void register(ICancellable cancellable) {
		synchronized (cancellableList) {
			cancellableList.add(cancellable);
		}
	}

	/**
	 *
	 * @param cancellable
	 */
	public void deregister(ICancellable cancellable) {
		synchronized (cancellableList) {
			cancellableList.remove(cancellable);
		}
	}
```

It apperas to be a candidate for LinkedList as it is not accessed too regularly
but elements are getting added and removed quite often.

hvbtup:
Hmm, I don't think changing the type to a `LinkedList` instead of `ArrayList` would help. In both cases removing elements takes _O(n)_.
Shouldn't it be a `HashMap` or something like that instead?

If you are profiling anyway, it would also be interesting to see the average element count of the structure.

Maybe the size is more of a problem. Looking at the code, it seems to me that registering/deregistering only takes place in the methods of `DataSourceQuery` class.

Maybe there is some kind of _memory leak_ like forgetting to call `deregister` somehow or delaying the call too long? That would cause the structure to grow and grow, if a big report uses several thousands or millions of queries. (Well, with a classical remote RDBMS, such a report would be a candidate for refactoring anyway, because it'd run forever and slow the DB down)

Because otherwise I cannot understand why a few (or a few dozens) queries which are open at the same time can cause the structure to grow to a significant size _n_.

speckyspooky:
Yes, I confirm the idea of Henning. I have also some reports which use many queries in combination with loops and with large row count (1Mio.) but I have never seen this problem.

Do you got the problem with the latest BIRT version 4.15?

giangianoulas:
There was indeed a memory leak. In ```DataSourceQuery#execute``` elements were added but never removed.
Also there was a unnecessary big syncronized block.

I agree that this report needs a redesign, but it is not that easy to be done.
[...]

PR #1635, opened 2024-04-18
Fix memory leak & and use Map instead of list to avoid itterating an Array constantly inside a synchronized block

giangianoulas:
Switch from List to Map on CancelManager to improve performance, 
Fix memory leak & reduce scope of synchronized block Issue #1633

### DIFF
--- a/data/org.eclipse.birt.data/src/org/eclipse/birt/data/engine/impl/CancelManager.java
+++ b/data/org.eclipse.birt.data/src/org/eclipse/birt/data/engine/impl/CancelManager.java
@@ -14,9 +14,9 @@
  *******************************************************************************/
 package org.eclipse.birt.data.engine.impl;
 
-import java.util.ArrayList;
-import java.util.List;
+import java.util.Set;
 import java.util.TimerTask;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  *
@@ -24,13 +24,12 @@ import java.util.TimerTask;
 
 public class CancelManager extends TimerTask {
 	//
-	private List<ICancellable> cancellableList;
+	private final Set<ICancellable> cancellables = ConcurrentHashMap.newKeySet();
 
 	/**
 	 * Constructor
 	 */
 	public CancelManager() {
-		cancellableList = new ArrayList<>();
 	}
 
 	/**
@@ -38,9 +37,7 @@ public class CancelManager extends TimerTask {
 	 * @param cancellable
 	 */
 	public void register(ICancellable cancellable) {
-		synchronized (cancellableList) {
-			cancellableList.add(cancellable);
-		}
+		cancellables.add(cancellable);
 	}
 
 	/**
@@ -48,9 +45,7 @@ public class CancelManager extends TimerTask {
 	 * @param cancellable
 	 */
 	public void deregister(ICancellable cancellable) {
-		synchronized (cancellableList) {
-			cancellableList.remove(cancellable);
-		}
+		cancellables.remove(cancellable);
 	}
 
 	/*
@@ -64,12 +59,9 @@ public class CancelManager extends TimerTask {
 	}
 
 	public void doCancel() {
-		synchronized (cancellableList) {
-			List<ICancellable> cancellableLists = new ArrayList<>(cancellableList);
-			for (ICancellable cancellable : cancellableLists) {
-				if (cancellable.doCancel()) {
-					cancellable.cancel();
-				}
+		for (ICancellable cancellable : cancellables.toArray(ICancellable[]::new)) {
+			if (cancellable.doCancel()) {
+				cancellable.cancel();
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


hvbtup (approved):
Looks good to me, thank you.

data/org.eclipse.birt.data/src/org/eclipse/birt/data/engine/impl/CancelManager.java
```diff
@@ -24,22 +26,21 @@
 
 public class CancelManager extends TimerTask {
 	//
-	private List<ICancellable> cancellableList;
+	private Map<String, ICancellable> cancellableMap = new HashMap<>();
```
merks, line 29: Why don't we just use a set?   If one uses a concurrent collection, would could avoid synchronize calls.  E.g, wouldn't this be better?
```
/*******************************************************************************
 * Copyright (c) 2004, 2009 Actuate Corporation.
 *
 * This program and the accompanying materials are made available under the
 * terms of the Eclipse Public License 2.0 which is available at
 * https://www.eclipse.org/legal/epl-2.0/.
 *
 * SPDX-License-Identifier: EPL-2.0
 *
 *
 * Contributors:
 *  Actuate Corporation  - initial API and implementation
 *******************************************************************************/
package org.eclipse.birt.data.engine.impl;

import java.util.Collections;
import java.util.Set;
import java.util.TimerTask;
import java.util.concurrent.ConcurrentHashMap;

/**
 *
 */
public class CancelManager extends TimerTask {
	//
	private final Set<ICancellable> cancellables = Collections
			.newSetFromMap(new ConcurrentHashMap<ICancellable, Boolean>());

	/**
	 *
	 * @param cancellable
	 */
	public void register(ICancellable cancellable) {
		cancellables.add(cancellable);
	}

	/**
	 *
	 * @param cancellable
	 */
	public void deregister(ICancellable cancellable) {
		cancellables.remove(cancellable);
	}

	/*
	 * (non-Javadoc)
	 *
	 * @see java.lang.Runnable#run()
	 */
	@Override
	public void run() {
		doCancel();
	}

	public void doCancel() {
		for (ICancellable cancellable : cancellables.toArray(ICancellable[]::new)) {
			if (cancellable.doCancel()) {
				cancellable.cancel();
			}
		}
	}
}
```

merks (changes requested):
I think generating keys in order to use a map is overkill and is certainly not optimal when a set will do and using a concurrent implementation can avoid synchronization entirely.

data/org.eclipse.birt.data/src/org/eclipse/birt/data/engine/impl/CancelManager.java
giangianoulas: While you write is true that a concurrent map can be used, it is definitely not bulletproof proof in concurrent environment. With that said i am not sure how this thing works and I opted to use a Map with synchronized. Also a Set uses a Map internally so i am not sure what the benefit will be.

As for the key generation, again i am not sure how this code is supposed to work and i wanted to be sure that it actually works.

merks: Adding methods for an ID and generating a UUID just so that you can create what is  effectively a set makes no sense to me and is not minimally invasive either.  A concurrent map is documented that "all operations are thread-safe" so I'm not sure how that is not bullet proof.  Now I see the class has a factory method for creating a set, so this simplification is more direct:
```
/*******************************************************************************
 * Copyright (c) 2004, 2009 Actuate Corporation.
 *
 * This program and the accompanying materials are made available under the
 * terms of the Eclipse Public License 2.0 which is available at
 * https://www.eclipse.org/legal/epl-2.0/.
 *
 * SPDX-License-Identifier: EPL-2.0
 *
 *
 * Contributors:
 *  Actuate Corporation  - initial API and implementation
 *******************************************************************************/
package org.eclipse.birt.data.engine.impl;

import java.util.Set;
import java.util.TimerTask;
import java.util.concurrent.ConcurrentHashMap;

/**
 *
 */
public class CancelManager extends TimerTask {
	//
	private final Set<ICancellable> cancellables = ConcurrentHashMap.newKeySet();

	/**
	 *
	 * @param cancellable
	 */
	public void register(ICancellable cancellable) {
		cancellables.add(cancellable);
	}

	/**
	 *
	 * @param cancellable
	 */
	public void deregister(ICancellable cancellable) {
		cancellables.remove(cancellable);
	}

	/*
	 * (non-Javadoc)
	 *
	 * @see java.lang.Runnable#run()
	 */
	@Override
	public void run() {
		doCancel();
	}

	public void doCancel() {
		for (ICancellable cancellable : cancellables.toArray(ICancellable[]::new)) {
			if (cancellable.doCancel()) {
				cancellable.cancel();
			}
		}
	}
}
```
Looking in the Eclipse SDK's sources, I see this specific map implementation is widely used there:

image

giangianoulas: ok, updated it according to suggestion

merks: FYI, it's generally better to amend the commit and force push it so that there is a single commit.

merks:
It would be great if you could squash this into a single commit.  I'm not sure how to actually do that with EGit though.  But if you don't know how to do it either, I can try to help with that.  Thanks for contributing!

giangianoulas:
I have no idea how to actually do it ;)

merks:
FYI, I've been investigating why the build fails with memory exceptions.  I finally got one to pass but now your force push kicked another build:

image

Let's see how that goes.

giangianoulas:
Commits are not squashed

merks (approved):
It looks better now. I just want to see a clean build before proceeding.

merks:
@giangianoulas 

Thanks for contributing and for squashing too.  🏅

Issue #1643, opened 2024-04-23
Enhancement to add excel formula options - original discussion: BIRT report with EXCEL formula #858


speckyspooky:
### Discussed in https://github.com/eclipse-birt/birt/discussions/858

<div type='discussions-op-text'>

<sup>Originally posted by **darryljohn** March 10, 2022</sup>
I have a BIRT report that I would like to export to EXCEL.  I would like to include EXCEL formulas in the report.  There is very little documentation online.  Does anybody have experience with this?  
![screenshot](https://user-images.githubusercontent.com/101336556/157678609-6801937c-e172-42d2-bed2-6a44f6a15c1c.jpg)

</div>

speckyspooky:
The ExcelEmitter options will be implemented to configure the dynamic text according to the original discussion as new options for the BIRT-excel-emitter.

The code-base is the following fork: 
https://github.com/hvbtup/birt/commit/e09556398936fbc02e07ccd27fd3a96bee23044f

The source won't be merged, I have done a specific analysis of the fork and the current code base from the master-branch.
The implementation will add the following two options:
- ExcelEmitter.ValueAsFormula
- ExcelEmitter.Formula

In addition to it some problems of the original code will be changed for example issues with the styling if excel formula will be used.

speckyspooky:
PR #1644 added to implements the enhancement.

I have created according test reports to test the functionality in different versions.
Attached I added screens and also one of my test reports with a static grid which include different excel-formula for numeric-calculation and string-concatination and also a small data table based on the classic cars with a column which include an excel-formula.

### Screen of the report and the preview
![report-design](https://github.com/eclipse-birt/birt/assets/41593722/4d5e9b59-bd0c-4c95-a4c1-c7e8dd7d21e6)

### Screen of the excel document: opening and result after execution
![excel-opening](https://github.com/eclipse-birt/birt/assets/41593722/65e34345-3ef5-4288-a37c-cbb7b78d5a16)

### Testing report and the data result like excel-file
[excel_formula.zip](https://github.com/eclipse-birt/birt/files/15083493/excel_formula.zip)

speckyspooky:
The PR #1644 with the enhancement is merged to the master branch

PR #1644, opened 2024-04-23
Enhancement to add excel formula options (#1643)

speckyspooky:
Enhancement to add excel formula options (#1643)

### DIFF
--- a/engine/uk.co.spudsoft.birt.emitters.excel/src/uk/co/spudsoft/birt/emitters/excel/StyleManagerUtils.java
+++ b/engine/uk.co.spudsoft.birt.emitters.excel/src/uk/co/spudsoft/birt/emitters/excel/StyleManagerUtils.java
@@ -787,7 +787,9 @@ public abstract class StyleManagerUtils {
 	 * @param defaultFont  The font to be used prior to the first RichTextRun.
 	 * @param widthMM      The width of the output.
 	 * @param richTextRuns The list of RichTextRuns to be applied to the string
-	 * @param wrap         The text use wrapped style
+	 * @param wrap         If the excel cell marked as wrapped text then use the
+	 *                     wrapped line break LineBreakMeasurer else use raw
+	 *                     TextMeasurer
 	 * @return The height, in points, of a box big enough to contain the formatted
 	 *         sourceText.
 	 */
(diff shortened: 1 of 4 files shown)


speckyspooky:
I researched the original changes of JT and compared it with the latest version of our master-branch.
The change is no copy or direct merge of the change. It is focused on the formula-scope.
So it is a new version at all but it can be compared with the original changes of JT.

engine/uk.co.spudsoft.birt.emitters.excel/src/uk/co/spudsoft/birt/emitters/excel/StyleManagerUtils.java
```diff
@@ -786,11 +787,12 @@ protected int getRichTextRunIndexForStart(List<RichTextRun> richTextRuns, int st
 	 * @param defaultFont  The font to be used prior to the first RichTextRun.
 	 * @param widthMM      The width of the output.
 	 * @param richTextRuns The list of RichTextRuns to be applied to the string
-	 * @return The heigh, in points, of a box big enough to contain the formatted
+	 * @param wrap         The text use wrapped style
```
hvbtup, line 790: The comment for wrap is not clear to me (and grammar errors)

hvbtup (commented):
There's too much code changed to understand it.
Is all of this actually related to formulas?

I mean, you added a whole line-breaking algorithm. 
From what I did decades ago when I developed the wordaxe hyphenation library including using it in a new a paragraph line-breaking algorithm for the Python ReportLab library, I know that those line-breaking algorithms are very fragile kind of program, where things can easily go wrong.

Is it possible for you to split the formula support and the layout enhancements you mentioned (which exactly?) into two separate PRs?

engine/uk.co.spudsoft.birt.emitters.excel/src/uk/co/spudsoft/birt/emitters/excel/StyleManagerUtils.java
speckyspooky: I will change it.

speckyspooky:
Of course there are some changes because the biggest change is to add the "formula"-options and give the formula-values the option for the correct styling on excel side.

You talking about the "new a paragraph line-breaking algorithm" - which section do you mean exactly?
Then I will check if it make sense to split it from my side or may be we can use an additional user-property for the "new line algroithm" (default: false).

hvbtup:
I mean the height calculation in `calculateTextHeightPoints`.

hvbtup:
Thomas, if you say this is needed then I believe you :-).
It's just that I don't understand what's  the height calculation got to do with formulas...
My naive feeling is that when a cell contains a formula, then it's not possible to automatically compute a meaningful height anyway, because we don't know the value (only the formula).

I'm +0.5 for merging. Even if I don't understand all the changes, probably they will have an effect only for cells with formulas, which are a new feature. Thus I think that the risk of breaking existing reports is near zero.

PR #1646, opened 2024-04-24
Switch to actions @v4

wimjongman:
projects hosted on GitHub and utilizing actions should take note that GitHub has issued a deprecation notice for some commonly used actions.

* actions/upload-artifact
* actions/download-artifact

The following versions are deprecated and will stop working in the not so near future:

* v1 and v2: from 30/06/2024
* v3: from 30/11/2024

It is highly encouraged to update to v4 of these actions asap to avoid any disruption.

See more details at:

https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/ https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/

Best regards,
Thomas Neidhart


wimjongman:
Maybe we can drop the uploads altogether?

merks:
The actions are just for verification.  Only Jenkins produces signed artifacts that are valuable to keep/update.

merks (approved):
Out with the old, in with the new.  😀

Issue #1651, opened 2024-04-27
ExcelEmitter, add the option to force the recalculation of formulas on excel side


speckyspooky:
The excel-emitter should be improved to support the recalculation flag of excel that the excel file will be start a formula recualculation when the file will be opend, Apache POI API: "setForceFormulaRecalculation".

The option should be available like user property: ExcelEmitter.ForceRecalculation

speckyspooky:
Change is merged with PR #1652 to the master branch

PR #1652, opened 2024-04-27
Enhance the excel emitter with force recalculation option (#1651)

speckyspooky:
enhance the excel emitter with force recalculation option (#1651)

### DIFF
--- a/engine/uk.co.spudsoft.birt.emitters.excel/src/uk/co/spudsoft/birt/emitters/excel/ExcelEmitter.java
+++ b/engine/uk.co.spudsoft.birt.emitters.excel/src/uk/co/spudsoft/birt/emitters/excel/ExcelEmitter.java
@@ -154,6 +154,9 @@ public abstract class ExcelEmitter implements IContentEmitter {
 	/** property: ExcelEmitter.StreamingXlsx */
 	public static final String STREAMING_XLSX = "ExcelEmitter.StreamingXlsx";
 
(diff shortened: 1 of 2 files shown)


merks (approved):
I'm not actually qualified to review this type of thing.  But of looks fine in terms of following the existing design pattern.

merks:
@speckyspooky 

BTW, the master builds are failing quite regularly because of memory restriction violations.  I will try to get this ironed out, but I need the help of the Foundation staff to come up with a reliable solution.  Mostly right now I just kick the build again and hope for the best:

https://ci.eclipse.org/birt/job/build/job/master/

Also, if  the following branch is stale, could you please delete it?

https://github.com/eclipse-birt/birt/tree/issue-1089

speckyspooky:
I deleted the branch of issue-1089

hvbtup:
Idea for an enhancement: In the absence of the `ExcelEmitter.ForceRecalculation` option, wouldn't it make sense to force the recalculation iff BIRT generated any formula? This way, BIRT would do the right thing by default.

speckyspooky:
The default handling of excel is that the recalculation will be done if necessary.
The forced recalculation will trigger excel to recalculate every value of an opened file and this could need lot of time.
So I see this option more as a point for the report developer as a kind of optimization level and can be used if it is really necessary.
That is the cause why I set the configuration not as active by default.

I understand the argument but I don't know which time consuming effect the changed default could cause.

PR #1653, opened 2024-04-27
documentation for the supported user properties of the PdfEmitter

speckyspooky:


hvbtup (commented):
In line 5, there is "optimize the excel output". You certainly mean PDF.

Issue #1656, opened 2024-04-28
Excel, add corresponding options of "display grindlines" and "display row-column header" for print out


speckyspooky:
Currently the excel emitter support the display of grindlines and the display of the row-column headings.
The apache poi library supports the corresponding options to control the print options of these.
Additional the options to set the "page fit" for printing and the "zoom"-option of excel should be added as user property.

The following user properties will be added:
- ExcelEmitter.PrintGridlines
- ExcelEmitter.PrintRowColHeadings
- ExcelEmitter.PrintFitToPage
- ExcelEmitter.DisplaySheetZoom

The options will be added on the same elements like e.g. "ExcelEmitter.DisplayGrindlines":
- page
- table
- list

speckyspooky:
The enhancement is given at PR #1657

speckyspooky:
The PR is merged to the master.

PR #1657, opened 2024-04-28
Enhanced print out options of the excel emitter (#1656)

speckyspooky:


speckyspooky:
Hello Henning, I have done a small enhancement of additional excel emitter properties which I need for one of my projects.
The change is corresponding to existing user properties. I would say that the changes are not so critical. My testings were fine.

hvbtup (approved):
A very useful addition!

Issue #1666, opened 2024-05-05
Excel & Indent: implementation of the "indent"-option for excel output


speckyspooky:
The current implementation of the excel-emitter doesn't support the usage of the excel formatting option of text-indent.
In the most cases addition blank columns were added to get the effect of an indent which increase the report structure and isn't really the best way to handle (simulate) the excel-indent.

The Apache POI library support the method "setIndention()" to set the indent for an excel-cell.

The target of the enhancement will be to add the indent-method and the spacing should be calculated based on the margin- & padding-spacing of the content element and also the cell-padding should be used to calculate the indent in full.

This would be a bigger change of the behavior for the excel output. Therefore I will implements an UserProperty to deactivate the indent setting with the name "ExcelEmitter.DisplayTextIndent" (default: true).

I will create a PR for this enhancement.

speckyspooky:
The PR #1667 containse the according changes to get the new option.

I created a special reference report to test the label-, text-, dynamic text- and data elements.
Attached the following details will be shown:
1. Screen of the report which will produce 3 pages for the 3 master test cases
2. Screen of the 3 generated cases
3. Attachments
3.1 The original report
3.2 The excel-output file with text-indent
3.3 The pdf-output file for comparison

### 1. Screen of the report
![excel-indent](https://github.com/eclipse-birt/birt/assets/41593722/191b762f-1a19-491a-a4d1-ca07b08b4769)

### 2. Screens of the excel-result
**2.1 Labels**
![Test-Case-Labels](https://github.com/eclipse-birt/birt/assets/41593722/b88646c2-63ef-4656-91ec-e8a30d28a66f)

**2.2 Text & Dynamic Text**
![Test-Case-Text-Elements](https://github.com/eclipse-birt/birt/assets/41593722/37dfa077-5d64-434e-9763-b0c58587b005)

**2.3 Data Elements**
![Test-Case-Data-Elements](https://github.com/eclipse-birt/birt/assets/41593722/6974fbd2-b0be-4282-a8b4-346fa088b360)

### 3. Attachments
**3.1 The original report**
[excel_indent.zip](https://github.com/eclipse-birt/birt/files/15214572/excel_indent.zip)

**3.2 The excel-output file with text-indent**
[demo-excel-indent.xlsx](https://github.com/eclipse-birt/birt/files/15214571/demo-excel-indent.xlsx)

**3.3 The pdf-output file for comparison**
[...]

PR #1667, opened 2024-05-05
Enhance the excel output to get the formatting option of text-indent (#1666)

speckyspooky:
Enhance the excel output to get the formatting option of text-indent (#1666)

### DIFF
--- a/engine/uk.co.spudsoft.birt.emitters.excel/src/uk/co/spudsoft/birt/emitters/excel/BirtStyle.java
+++ b/engine/uk.co.spudsoft.birt.emitters.excel/src/uk/co/spudsoft/birt/emitters/excel/BirtStyle.java
@@ -66,6 +66,7 @@ public class BirtStyle {
 	private IStyle elemStyle;
 	private CSSValue[] propertyOverride = new CSSValue[BirtStyle.NUMBER_OF_STYLES];
 	private CSSEngine cssEngine;
+	private boolean useTextIndent = true;
 
 	/**
 	 * Constructor 01
@@ -219,6 +220,10 @@ public class BirtStyle {
 
 	private static BitSet PrepareSpecialOverlayProperties() {
 		BitSet result = new BitSet(BirtStyle.NUMBER_OF_STYLES);
+		result.set(StyleConstants.STYLE_MARGIN_LEFT);
+		result.set(StyleConstants.STYLE_MARGIN_RIGHT);
+		result.set(StyleConstants.STYLE_PADDING_LEFT);
+		result.set(StyleConstants.STYLE_PADDING_RIGHT);
 		result.set(StyleConstants.STYLE_BACKGROUND_COLOR);
 		result.set(StyleConstants.STYLE_BORDER_BOTTOM_STYLE);
 		result.set(StyleConstants.STYLE_BORDER_BOTTOM_WIDTH);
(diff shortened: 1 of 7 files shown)


speckyspooky:
I have added the test results, screens and original test reports at the issue-ticket.

The idea is to sum the margin & padding and calculate it to the indent-factor.
The text-indent is a special MS Excel unit and I researched an according translation-factor to get the indent-values. 

To sum the margin & padding the values must be synced to the same value. I decieded to use "mm" as default unit,
because it was used within Spudsoft as default-unit in different situations.

The main changes are given at the CellContentHandler within the new method "overlayBirtCellStyleSpacing()"and at the StyleManager "createStyle()".
The method "overlayBirtCellStyleSpacing()" handle the whole topic of the correct style creation of the birtCellStyle
and the "createStyle()" includes the new calculation of the text-indent for excel.

**Hint:**
One of the main topics is that "overlayBirtCellStyleSpacing()" solve some topics with the style properties.
Because the text element is handled in a different way for excel because it is not like "Word" or "PDF" an own element with a "p"(aragraph), it is a combination of cell- and text-style.
Therefore the method handle the overlay to set the correct margin and padding for the excel-cell.

hvbtup (commented):
I see the same typo at least two times in README.md: "teh" instead of "the".

hvbtup (changes requested):
I'm not sure if the default value of True for `ExcelEmitter.DisplayTextIndent` is a good idea.
It will probably change (aka break) the layout of existing reports.
IMHO the default should be to _not_ use indentation.

Furthermore, I'm not sure if using the margin of the element inside a cell is a good idea.
It is hard to predict for a developer what should happen if there are more than one elements inside a cell.
Instead, only the padding of the cell should be used IMHO.

speckyspooky:
I will remove the typo and I will double check the behavior of multiple content.

But from my side it would be good to have the indention active because I would expect the indent by default.
It is for me the same like PDF or Word. There the indent is working directly without any activation. Yes, I know this is a change of the behavior therefore the user property to switch of the change.

I will check the padding of the cell but I will have also an indent if the cell has "no padding" but the labels / text-elements would have a margin & padding. Then I wants to get an indent.
But for that I will add an additional user-property for "ExcelEmitter.DisplayIndentMode" that we can configure the calculation with the options "SpacingCell", "SpacingElement" and "SpacingAll".

hvbtup:
This may be a silly question, but:
Can the user-properties be defined at the top-level (i.e. the report itself) and will they inherit from the corresponding container?
If yes, I'm fine with "indentation is the default".

What should happen if the layout is more complex, e.g. a Grid Item or Table Item inside a Cell, possibly at 3 or 4 levels?
Such reports do exist (called sub-reports in some other tools).
In fact in our case, they are quite common, however we don't use them with Excel output.

Personally, I don't think it is possible to correctly calculate the spacing in such more general cases.
But I expect users to expect this :-).
Thus I think it should be clarified to what extent the indentation can be expected to work correctly.

speckyspooky:
The indent will be calculated only on label / text-element level and will take a look to the parent cell.
The nested tables are handled with the default behavior and so the structures will be restructured from your mentioned case to a possible excel table/grid.
This situation of multiple table-in-table-in-table-in-grid will be handled before and the change will work afterwards if the cell contains a label/text-element.

And in such special cases the ExcelEmitter has a default behavior because he doesn't allow to nest to many tables in a cell.
Then you will get an error by default.

speckyspooky:
question 1: Can the user-property defined at top-level / report-level?
answer:
- Yes, you can configure the user-property on report level and no indent will be set.
- But you can use also mixed usage if you have different tables & grids then you can set it on the tables which souldn't have an indent.
- The lowest level is the label/text-lement-level him self.

question 2: What will be happend if 2 (more) elements will be added in a cell and both element have a different margin & padding?
answer:
- The spacing values of the first cell element will used for all text-elements on a cell. It is not possible to have different indents at a cell.

question 3: What will be happend if a cell contains further inner tables/grids?
answer:
- The excel emitter resolves the internal grids directly and the last action will be the cell-content formating. This will be done based on the first cell-conetnt-element. The indent of the first element will be used.


**Added user property for indent mode**
I have also added the according user property to control the indent-calculation with the options "SpacingCell", "SpacingElement" and "SpacingAll".

hvbtup (approved):
The text in ..../uk/co/spudsoft/birt/emitters/excel/README.md could use some polish, but this can be done later, and should preferably be done by a native English speaker.

wimjongman:
Thanks, Thomas!

Issue #1670, opened 2024-05-07
Fixing of eclipse warnings noted at change of #1667


speckyspooky:
I figured out a lot of eclipse warnings during the change of #1667.
I will create additional changes to add missing documentation and solve eclipse hints.

speckyspooky:
Changes are merged into the master branch.

PR #1671, opened 2024-05-07
Fix eclipse warnings and add java documentation (#1670)

speckyspooky:
Fix eclipse warnings and add java documentation (#1670)


speckyspooky:
Hi Wim, I added some java docs to solve eclipse warnings and changed some code warnings.
I changed also the "PropertyParser". This seems to me very special/old stuff. Please double check it to avoid mistakes from my side.
Thanks in advance!

wimjongman (approved):
Very nice, Thomas. Thank you.

PR #1674, opened 2024-05-08
Fix eclipse warnings and add java documentation, part 2 (#1670)

speckyspooky:


engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/content/IReportContent.java
```diff
@@ -61,17 +61,25 @@ public interface IReportContent {
...
 	 * @return the TOC Tree
 	 */
 	ITOCTree getTOCTree(String format, ULocale locale);
 
 	/**
-	 * return root content
+	 * Get the root content
 	 *
-	 * @return
+	 * @return Return the root content
```
wimjongman, line 74: For the future it is not needed to add Return to @return

wimjongman (approved):
LGTM. Thank you!

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/content/IReportContent.java
speckyspooky: I will think on it!

Issue #1665, opened 2024-05-05
BIRT 4.16 202405030536, Designer-Crashed: topic depended on JRE 17 vs. 21


speckyspooky:
The All-In-One-designer is crashed with the currently "nightly build".

Tested designer versions from my side 4.16:

» N202405030536: 4.16 designer crashed
» N202404301804: 4.16 ok
» N202404291807: 4.16 ok
» N202404271127: 4.16 ok
» N202404261927: 4.16 ok

Could it be a topic with the latest eclipse changes an the required version of JRE 21 for some libraries or some central changes at eclipse-framwork side?
I have no problems with my dev-environment locally.

**Log:**  [log.zip](https://github.com/eclipse-birt/birt/files/15214098/log.zip)

### Designer
![416-crash](https://github.com/eclipse-birt/birt/assets/41593722/eee187ff-1056-4991-8b29-b94d1b8e5244)

### Error-Screen 01, 1st message
![details-error-1](https://github.com/eclipse-birt/birt/assets/41593722/18a1da38-1bf9-4c93-911a-dfa05c72d077)

### Error-Screen 02, 2nd message
![details-error-2](https://github.com/eclipse-birt/birt/assets/41593722/e50c90ca-56a6-4f6d-b5c7-6a92dc6a9e18)

merks:
GEF has been making **lots** of changes lately.  It seems to me we already dealt with a problem like this in the recent past. 

I have limited capacity this week to investigate problems.

merks:
@speckyspooky 

If you do  Help -> Perform Setup Tasks... you will update the target platform and see this problem.

The problem is a result of this commit:

https://github.com/eclipse/gef-classic/commit/451bc1fff9a904b7e303c5747d05989b738d1c70

Which is for this issue:

https://github.com/eclipse/gef-classic/issues/418

It happens here:

org.eclipse.birt.report.designer.internal.ui.editors.rulers.EditorGuideEditPart.updateLocationOfFigures(int)

![image](https://github.com/eclipse-birt/birt/assets/208716/e6951391-3415-4687-89f9-063590235d54)

Apparently that should now be a Rectangle, but I have no idea the design intent here in BIRT, i.e., how to convert the position to a rectangle.

@azoitl  @ptziegler

I'm not sure this new exception throwing is okay.  The API being called has no constraint on the constraint:

[...]

PR #1678, opened 2024-05-09
Fix BIRT-designer, replaced XYLayout and XYLayoutEditPolicy (#1665)

speckyspooky:
Fix BIRT-designer, switch from XYLayout to AbstractConstraintLayout and switch from XYLayoutEditPolicy to ConstrainedLayoutEditPolicy (#1665)

### DIFF
--- a/UI/org.eclipse.birt.report.designer.ui/src/org/eclipse/birt/report/designer/internal/ui/editors/rulers/EditorRulerLayout.java
+++ b/UI/org.eclipse.birt.report.designer.ui/src/org/eclipse/birt/report/designer/internal/ui/editors/rulers/EditorRulerLayout.java
@@ -16,8 +16,8 @@ package org.eclipse.birt.report.designer.internal.ui.editors.rulers;
 
 import java.util.List;
 
+import org.eclipse.draw2d.AbstractConstraintLayout;
 import org.eclipse.draw2d.IFigure;
-import org.eclipse.draw2d.XYLayout;
 import org.eclipse.draw2d.geometry.Dimension;
 import org.eclipse.draw2d.geometry.Rectangle;
 
@@ -25,7 +25,7 @@ import org.eclipse.draw2d.geometry.Rectangle;
  * add comment here
  *
  */
-public class EditorRulerLayout extends XYLayout {
+public class EditorRulerLayout extends AbstractConstraintLayout {
 
 	/**
 	 * @see org.eclipse.draw2d.AbstractLayout#calculatePreferredSize(org.eclipse.draw2d.IFigure,
@@ -49,7 +49,7 @@ public class EditorRulerLayout extends XYLayout {
 	 */
 	@Override
 	public void layout(IFigure container) {
-		List children = container.getChildren();
+		List<?> children = container.getChildren();
 		Rectangle rulerSize = container.getClientArea();
 		for (int i = 0; i < children.size(); i++) {
 			IFigure child = (IFigure) children.get(i);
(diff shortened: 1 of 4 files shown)


speckyspooky:
@merks 
Hi Ed,
according to the hints of ptziegler I changed the according classes and replaced the XYLayout & XYLayoutEditPolicy.
I have added the missing methods from the original classes to the child class (= missing methods).
Please take a look and let me know what your opinion is because it is abstract from my side to be so deep on the designer-ui-level.

UI/org.eclipse.birt.report.designer.ui/src/org/eclipse/birt/report/designer/internal/ui/editors/rulers/EditorRulerLayout.java
```diff
@@ -49,7 +49,7 @@ public Object getConstraint(IFigure child) {
 	 */
 	@Override
 	public void layout(IFigure container) {
-		List children = container.getChildren();
```
ptziegler, line 52: I suggest using `List<? extends IFigure>` instead, to avoid the explicit cast in line 55.

UI/org.eclipse.birt.report.designer.ui/src/org/eclipse/birt/report/designer/internal/ui/editors/schematic/editpolicies/TableXYLayoutEditPolicy.java
```diff
@@ -28,16 +31,27 @@
...
+
+	private static final Dimension PREFERRED_SIZE = new Dimension(-1, -1);
 
 	/**
 	 * Constructor
 	 *
 	 * @param layout
 	 */
-	public TableXYLayoutEditPolicy(XYLayout layout) {
+	public TableXYLayoutEditPolicy(AbstractConstraintLayout layout) {
```
ptziegler, line 43: If the layout is not used internally, does it make sense to keep it as a constructor argument?

I also wonder if it makes sense to keep the class name, when it no longer extends the XYLayoutEditPolicy.

UI/org.eclipse.birt.report.designer.ui/src/org/eclipse/birt/report/designer/internal/ui/editors/rulers/EditorRulerLayout.java
speckyspooky: Done.

UI/org.eclipse.birt.report.designer.ui/src/org/eclipse/birt/report/designer/internal/ui/editors/schematic/editpolicies/TableXYLayoutEditPolicy.java
speckyspooky: Good question, my first topic was to get it runable :o)
(Now we can talk about improvements.)

speckyspooky: Oh yea, you are completely right, I removed the parameter.

merks:
@speckyspooky 

I appreciate that you are being proactive about improving the code!

@ptziegler 

I appreciate very much that you as a GEF expert are helping review the changes!

(I'm traveling this week so it's harder to keep up.)

merks (approved):
@speckyspooky 

I find it difficult to review changes when there are a bunch of stacked commits.  Generally it would be better, in my opinion, if you need to modify a PR's commit, that you amend it and force push it so that a PR has a single commit.  I think some of the intermediate states here don't even compile and/or are not functionally complete/correct so it kind of pollutes the history with unless/misleading intermediate states.

But I don't want to hold things up with fussiness because the final changes do look correct and do appear to function well to eliminate the logged warnings.

merks:
It's a bit ironic that a contribution intended purely to make contribution easier (a contribution that has no impact on existing functionality neither in the code nor in the existing Oomph project setup) languishes for a week.  I expect everyone is very busy with other things just like me so I do understand, but, if I were average-Jane user, I would be a little more annoyed with every passing day.

Issue #1686, opened 2024-05-15
Excel, error on file opening when large column-width used at tables-/grid-definition


speckyspooky:
If a report contains a table or grid with a huge column-width and the user create an excel file then an error appears at the moment of file opening. A message will be displayed that there is an issue and excel will start a fxing of it.
But at the end it could be that data will be lost.

Excel message - part 01
![grafik](https://github.com/eclipse-birt/birt/assets/41593722/c8ef2eeb-80ac-461e-9703-155f13063457)

Excel message - part 02
![grafik](https://github.com/eclipse-birt/birt/assets/41593722/750bcea7-29d0-4c7f-a290-eb2306319ff8)

**Reason:**
The problem is caused on the calculation of the column with of excel.
At a specific moment the number type "short" is used and it gives a number overflow (from positiv to negativ, number circle).

I will populate a PR for that.

speckyspooky:
The fix is merged to the BIRT-master.

PR #1687, opened 2024-05-15
Fix of excel column width calculation (#1686)

speckyspooky:


speckyspooky:
The PageHandler has the ame calculation issue at the method "processCellImage".
There is also the data type "short" used to calculate the image width but due to number circle overflow
the calculated with becomes negative which cause uncorrectly scaled images.
Therefore the data type "short" is removed.

PR #1704, opened 2024-05-23
do not autosize columns in case of StreamingXlsx eclipse-birt#1703

aremund:


speckyspooky (changes requested):
Instead to use the check of "XSSFSheet" twice with your condition, it would be better to set it directly as an own "if-block" before the "if (calcWidth > 1.0) {..." starts.

So you can avoid your special inner "} else if (" which will not set in all cases in compare to the old version.

aremund:
I understand what you mean and to start with, I implemented it this way.
But what I still want in the StreamingXlsx-case is setting the column to the calculated width. I would need to repeat this calculation for the StreamingXlsx-case. (the line `state.currentSheet.setColumnWidth(col, (int) calcWidth);`)
(Or am I now the confused one and do see the obvious anymore?)

aremund:
It's official, I am the one not seeing the obvious anymore. I adapted the code accordingly.

speckyspooky (approved):

aremund:
Sorry, I cannot merge due to my permissions. Would be happy, if someone else can merge.

merks:
For the future, please keep a PR to a single commit by amending and force pushing 

https://github.com/eclipse-birt/birt/pull/1701#issuecomment-2123940466

PR #1717, opened 2024-06-06
Lock in specific permanent update sites in the target platform

merks:


merks:
This is a work in progress.  The JustJ EPP repo won't be final until early next week, but it will be a promotion/copy of repo referenced here so there will be actual content changes between now and then.

PR #1720, opened 2024-06-07
Build product without org.eclipse.rcp.configuration.feature.group

merks:


merks:
FYI, I have tested locally that the product builds and launches correctly with this dependency and removing it helps simplify the target platform's dependencies.

PR #1725, opened 2024-06-10
Lock in justj/epp update site in the target platform

merks:


merks:
@speckyspooky @wimjongman 

FYI, this is the final change to support a reproducible build for the 4.16 release.

image


It will be great, one day, to get rid of the legacy thing.  I'd love to get rid of derby.  Updating that has major overhead; overhead that is difficult to justify for the primary uses case as an example...

wimjongman:
Do we need to restest?

merks:
This change does not change the actual content of the target platform, so while I could produce a new milestone build for testing, it would not have any dependencies with different binary content. 

So if folks are happy that the current milestone is of release quality, then we are in good shape to proceed with a release on Wednesday.  It all seems to be going so smoothly this time around!  👍

Issue #1727, opened 2024-06-11
Out of memory error when creating a chart


wimjongman:
Steps to reproduce

* Create a simple table with the customer table
* Drag a chart below the table
* Fill in X and Y

!ENTRY org.eclipse.ui 4 0 2024-06-11 11:35:50.977
!MESSAGE Unhandled event loop exception
!STACK 0
org.eclipse.swt.SWTException: Failed to execute runnable (java.lang.OutOfMemoryError: Java heap space)
	at org.eclipse.swt.SWT.error(SWT.java:4922)
	at org.eclipse.swt.SWT.error(SWT.java:4837)
	at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:135)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:4058)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3676)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$5.run(PartRenderingEngine.java:1151)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:339)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.run(PartRenderingEngine.java:1042)
	at org.eclipse.e4.ui.internal.workbench.E4Workbench.createAndRunUI(E4Workbench.java:152)
	at org.eclipse.ui.internal.Workbench.lambda$3(Workbench.java:639)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:339)
	at org.eclipse.ui.internal.Workbench.createAndRunWorkbench(Workbench.java:546)
	at org.eclipse.ui.PlatformUI.createAndRunWorkbench(PlatformUI.java:173)
	at org.eclipse.ui.internal.ide.application.IDEApplication.start(IDEApplication.java:152)
	at org.eclipse.equinox.internal.app.EclipseAppHandle.run(EclipseAppHandle.java:208)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.runApplication(EclipseAppLauncher.java:143)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.start(EclipseAppLauncher.java:109)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:439)
[...]

PR #1730, opened 2024-06-13
Out of memory error when creating a chart #1727

wimjongman:
1. Sometimes NumberFormat is used to handle the decimal point
and in other cases there is a dependency on ULocale.

A small fix was made to use the default ULocale in the
Numberformat calculations instead of the default NumberFormat

The whole operation is not fully understood so more issues
could arise.

2. To avoid memory cramps, an additional 1GB was added to the
maximum memory.

See the issue for a discussion.

### DIFF
--- a/build/birt-packages/birt-report-all-in-one/BIRT.product
+++ b/build/birt-packages/birt-report-all-in-one/BIRT.product
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?pde version="3.5"?>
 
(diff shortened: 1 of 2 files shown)


speckyspooky (approved):
Looks good!

merks (approved):
Looks simple and consistent so if that fixed the problem you see, it seems likely to be a good thing.

wimjongman:
Cheers!

hvbtup:
Fun fact: While driving to work this morning, I thought that the issue might be caused by the usage of different locales at different places, which could be causes by using different libraries (or by different versions of the same library, not all initialized in the same way, "thanks" to OSGI).

Looking at Wim's fix, it seems that I was on the right track.

I'd bet that the issues I experienced on a German Windows eg. when I try to specify a border width of 0.5 pt (because "small" border width is still not small enough) have similar reasons.

wimjongman:
Hi Henning, yes, you were on the right track. Now we know better the underlying issue we can probably also fix these.

Issue #1736, opened 2024-06-14
Use edge as default browser and use modern help


wimjongman:
I would like to activate the modern help and use edge as the default browser.

```
-Dorg.eclipse.swt.browser.DefaultType=edge
-Dorg.eclipse.help.webapp.experimental.ui=true

```
## L&F of the modern help
![image](https://github.com/eclipse-birt/birt/assets/180969/7063893b-cb75-4175-bd98-6d053448b5a4)

merks:
I guess that edge setting is one that you should only set for Windows we have to be a bit careful how we do that in the product file...

wimjongman:
I think it will not harm if edge is not available. It will go to the default.

wimjongman:
But I made the change anyway.

PR #1737, opened 2024-06-14
Use edge as default browser and use modern help #1736

wimjongman:


build/birt-packages/birt-report-all-in-one/BIRT.product
```diff
@@ -39,6 +39,8 @@ Powered by the Eclipse Platform: https://eclipse.org/eclipse
       </programArgs>
       <vmArgs>-Xms512m
 -Xmx2048m
+-Dorg.eclipse.swt.browser.DefaultType=edge
```
merks, line 42: I don't think we should set this for all operating system.  See `-Dorg.eclipse.swt.internal.carbon.smallFonts` below?

UI/org.eclipse.birt.report.designer.ui.rcp/BIRT.product
```diff
@@ -1,33 +1,64 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?pde version="3.5"?>
 
-<product name="%ApplicationName" id="org.eclipse.birt.report.designer.ui.rcp.BIRT" version="4.17.0.qualifier" application="org.eclipse.birt.report.designer.ui.rcp.DesignerApplication" useFeatures="false" includeLaunchers="true">
+<product name="BIRT Report Designer All-In-One" uid="org.eclipse.birt.report.designer.all" id="org.eclipse.birt.branding.birt_all_in_one" application="org.eclipse.ui.ide.workbench" version="4.17.0.qualifier" type="features" includeLaunchers="true" autoIncludeRequirements="true">
```
merks, line 4: I don't think we even ship this product.   This is the only one we ship:

image

Let's not change stuff that is effectively garbage as far as I can tell.

wimjongman: Sorry my bad, I used github to make the quick change.

merks (changes requested):
I'm really confused about which product is what.  I can't stand all these confusing files. 

In any case, you have a bunch of changes here that don't appear related to the title of the PR...

merks:
If I look at the IDs correctly I think this is the one we ship isn't it?

image

wimjongman:
Yes, this is a mess. I think I got the correct one now.

merks:
Yes, I'm pretty sure:

image

Also in this report I just generated:

https://download.eclipse.org/oomph/archive/reports-extra/birt-4.16/download.eclipse.org/birt/updates/release/4.16.0/index/org.eclipse.birt.designer_4.16.0.v202406131946.html

All these *.product files are way too confusing...

wimjongman:
I updated it yesterday as well and I got the correct changes in the all-in-one

wimjongman:
https://github.com/eclipse-birt/birt/commit/bb1631625f1ecff8a55b2d51dfcfd5b6d6c758ad

merks:
> Yes, this is a mess. I think I got the correct one now.

This suggests that we put one product in the update site but zip up a different one for redistribution: one which is not actually in the update site.  I really could literally just scream...

Also, multiple commits in a PR make me unhappy.  These all end up in the history.  And pushing branches to the main repo instead of too a fork also makes me unhappy:

image

It would be so much better to use a workflow like this one:

https://github.com/orgs/eclipse-simrel/discussions/3

In any case, it's very late which also makes me unhappy.

wimjongman:
> > Yes, this is a mess. I think I got the correct one now.
> 
> This suggests that we put one product in the update site but zip up a different one for redistribution: one which is not actually in the update site. I really could literally just scream...

LOL! Yes, I will work on that for the 4.17 release.

> 
> Also, multiple commits in a PR make me unhappy. These all end up in the history. And pushing branches to the main repo instead of too a fork also makes me unhappy:

image


> It would be so much better to use a workflow like this one:
> 
> https://github.com/orgs/eclipse-simrel/discussions/3

We use exactly that workflow 🎉

> 
> In any case, it's very late which also makes me unhappy.

Late in the day? Or late to get this in 4.16? 4.16 has already sailed.

merks (approved):

merks:
Late in the day. A glass of wine 🍷 improves the mood.  We can make 4.17 better. After all, no one complained about it. 😜

wimjongman:
> Late in the day. A glass of wine 🍷 improves the mood. 

Cheers! You deserved it.

Issue #1750, opened 2024-06-23
Excel, image scaling is wrong when only one image dimension is set


speckyspooky:
The spudsoft excel emitter shows images into the created excel files with a wrong scaling when only 1 dimension is set.
The reason of this behavior is the internal default handling which calculate the missing dimension based on the original physical values
and not based on the scaling in combination of the given value and the physical dimensions (like on the designer or web viewer).

I will provide a fix for that topic.

### Image with given width and without height

![image-excel-no-height](https://github.com/eclipse-birt/birt/assets/41593722/2c7979c2-e187-43f8-a8d1-42bb89dd3745)

### Image with given height and without width

![image-excel-no-width](https://github.com/eclipse-birt/birt/assets/41593722/e3a93ed7-007d-4700-811f-c509c8707f7e)

speckyspooky:
The PR for this fix is #1751

PR #1751, opened 2024-06-23
Fix the image scaling based on the given dimension (#1750)

speckyspooky:
The scaling of the images will be done based on the given dimension if one of dimesnions are missing in relation to the given physical deimensions.


speckyspooky:
I have added the test results at the issue ticket #1750

wimjongman (approved):
96.0 is a magic number. For the next iteration please move to a constant.

speckyspooky:
I know the magic of the number. I didn't changed it because was original code.
I will change the number directly with this PR so we have it directly clean up.

wimjongman:
> I know the magic of the number. I didn't changed it because was original code. I will change the number directly with this PR so we have it directly clean up.



> I know the magic of the number. I didn't changed it because was original code. I will change the number directly with this PR so we have it directly clean up.

You're the best!

Issue #1752, opened 2024-06-24
Excel, small images will be scalled to large images and lose the original size


speckyspooky:
The excel emitter (spudsoft) has an internal scaling to handle images in a special way for the excel output.
The excel which are too large and would overspan a cell will be shrinked, which is ok.
But images which have a smaller size like the cell will be grown.

The normal handling should be to keep the original image size of images which are smaller like the content cell.

### Screen with 3 cases:
- Case 1: standard image size without scalling - Ok
- Case 2: large image which will be shrinked to th cell - Ok
- Case 3: small image will be grown - Error

I will provide a PR for that. As fallback I will provide a user-property to restore the original behavior.

![scaling-standard-spudsoft-excel](https://github.com/eclipse-birt/birt/assets/41593722/b1d911fe-f871-4bf9-b6cf-bfde82c8c165)

speckyspooky:
The change is included at PR #1753 

The new user property is **ExcelEmitter.ImageScalingToCellDimension** (default: false). If it is set to true all images will be scaled to the cell dimension also the small images.

### Screen of new handling
![enhanced-scaling-standard-spudsoft-excel](https://github.com/eclipse-birt/birt/assets/41593722/142aaf63-c334-48d8-94a8-7b50f900c7e6)

speckyspooky:
The change is merged to the master.

PR #1753, opened 2024-06-24
Enhance the image scaling of excel and keep the size of small images (#1752)

speckyspooky:
The image scaling is enhanced to keep the original image size of small images without scaling to cell dimension


speckyspooky:
The demo screen is added to the issue ticket.
I will take a look into the orientation of the image in relation to the cell (alignment).

wimjongman (approved):
🙏

PR #1767, opened 2024-06-29
Remove the org.eclipse.birt.designer product from the update site

merks:
- Include the org.eclipse.birt.report.designer.all product in the update site with all of its transitive dependencies
- Change references to org.eclipse.birt.designer using org.eclipse.birt.branding.birt_all_in_one instead.


merks:
@wimjongman 

I've changed things so that bogus product is no longer in the update site but instead the update site includes the all-in-one product.

I've tested locally that it's possible to generate an Oomph Product (like we do for the EPP packages) from the BIRT update site's product IU:

image

And I tested that one can create a BIRT installation from that.

As a result, it will be possible to update installations from the update sites made available already:

image

I checked that the other zips we produce and redistribute look to be pretty much the same size as before so I think it looks good.

speckyspooky:
@merks
I like your idea for this reorganization.

My questions is to the naming when we do this reorganization and we haven't any mor the rcp-version
why we call it "birt-all-in-one" instead of only "birt-designer"?
Yes, we can keep it, but may be if we reorganize it we can think about the naming.

@wimjongman Have you a preference here?

speckyspooky (approved):

merks:
One consideration is that the unzipped products already use the given id so we probably don’t want to change any ids so that updates from now on will just start to work. But of course we can change any of the user-facing labels  to whatever we want.  That being said, these have probably become familiar by now.

wimjongman:
Yes, this looks very nice. Thank you. 

#### Sidenote
I would like to have a lightweight product at some point. This all-in-one is a bit bloated.

afbeelding

merks:
> #### Sidenote
> 
> I would like to have a lightweight product at some point. This all-in-one is a bit bloated.
> 
> afbeelding

Yes, it looks like everything including the kitchen sink is in there and I'm really not sure how folks decided what should be included.

The EPP product contained way less stuff:

image

Maybe we should consider pruning down the all-in-one. Are PDE and so many EMF features really needed? The above also contains additional stuff not in the all-in-one, i.e., jpt and jst.

It's of course easily maintain a single right-weight product...  Maybe we should open an issue to agree on a best approach....

PR #1769, opened 2024-06-30
Normalize all line separators to linux

merks:
- Convert all line separators that have been committed with \r\n to be just \n.


merks:
@speckyspooky @wimjongman 

This is a commit from hell.   The comments I saw the other day about strange diffs made me wonder to what extend files have been committed to the repository with Windows line endings.  Apparently this is a very common mistake that as accumulated over the years.  This commit normalizes them all.


As an example, if you save this file:

https://raw.githubusercontent.com/eclipse-birt/birt/master/build/birt-packages/birt-epp/category.xml

and view it in an editor that displays the line endings:

image

you should see no Windows line endings, but here we see a mixture of such things.

I've done this type of fix in other projects, e.g., in all the Eclipse TLP's repositories...

I assume you guys are okay with this massive corrective action...

speckyspooky:
Wow, lot of changes over the whole stuff - there are inldued all kind of files, .txt, .msg, .html, .rptdesign, xml, .java and so on.
But we cannot avoid that we will get again windows-line breaks.

merks:
The Oomph setup configures the repository via autocrlf to convert \r\n to \n while committing and \n to \r\n when creating the working tree:

image

Sometimes folks get it into their heads that they should have this set to false, but anything other than setting it to input or true can and will cause problems:

https://stackoverflow.com/questions/3206843/how-line-ending-conversions-work-with-git-core-autocrlf-between-different-operat

So we can and should avoid this by minimally configuring autocrlf  to input.

wimjongman (approved):
Yes, great.

PR #1777, opened 2024-07-01
Remove *.product files that are not actually used to build products

merks:
- Update the all-in-one product configuration to match that of the Platform SDK and the EPP products.


merks:
@wimjongman @speckyspooky 

Here I am deleting the confusing *.product files that are not actually used to build products.  They're actually not used at all.  

This way there are only two product files:

image

The one used to build the all-in-one and the one used to build birt-runtime-osgi...

I assume you're okay with this...

speckyspooky:
Yes, I'm fine with it.

PR #1779, opened 2024-07-02
Simplify and cleanup the build process

merks:
- Garbage collect unused ant and shell scripts.
- Garbage collect unused checked-in jars.
- Avoid creating nested  within jars as much as possible.


hvbtup:
Ed, I can't look at the changes, but I just want to say that I'm glad you tackle this task.

merks:
@hvbtup 

I appreciate the good wishes!  The build is such a nightmare of complex stuff that represent a major maintenance hurdle.  Deleting unused stuff just makes what's left easier to understand and easier to maintain.

One goal in this process here is to figure out why these jars checked in and to replace them with proper dependencies.

image

Of course I'm a bit scared with all the cross ant scripting that something might break, but what I have built locally (via a local Tycho build) all appears to continue to work fine:

image

merks:
@speckyspooky  @wimjongman 

Given the runtimes are tested as part of the build, I assume that all is okay after this change.    But is there some other manual confirmation/testing that you folks do.  If something has gone missing, it's best to find out about it sooner than later!

Issue #1778, opened 2024-07-02
securityDomain should be null if setSecurityController() was never called


len-ro:
No report can be run when deploying birt 4.16 in a tomcat with securityManager on. I am trying to upgrade birt from an older 3 version and have tried with various combinations of java and birt 4 version yet I am having this error:

```
java.lang.IllegalArgumentException: securityDomain should be null if setSecurityController() was never called
   at org.mozilla.javascript.Context.compileImpl(Context.java:2422)
   at org.mozilla.javascript.Context.compileString(Context.java:1404)
   at org.mozilla.javascript.Context.compileString(Context.java:1392)
   at org.eclipse.birt.report.engine.javascript.JavascriptEngine.compile(JavascriptEngine.java:214)
   at org.eclipse.birt.report.engine.javascript.JavascriptEngine.compile(JavascriptEngine.java:1)
   at org.eclipse.birt.core.script.ScriptContext.compile(ScriptContext.java:127)
   at org.eclipse.birt.report.engine.executor.ExecutionContext.compile(ExecutionContext.java:694)
   at org.eclipse.birt.report.engine.executor.ExecutionContext.evaluate(ExecutionContext.java:686)
   at org.eclipse.birt.report.engine.executor.ExecutionContext.evaluate(ExecutionContext.java:671)
   at org.eclipse.birt.report.engine.api.impl.EngineTask.validateAbstractScalarParameter(EngineTask.java:701)
   at
``` 

Seems related (or the same) as: https://bugs.eclipse.org/bugs/show_bug.cgi?id=412170. Yes I know the securityManager is deprecated and will be removed but I cannot drop it at this point. Anyone had a similar issue? 

Thanks.

len-ro:
Just to respond to my own comment, patching birt as in: https://github.com/len-ro/rhino-fork-for-birt/tree/Rhino1_7_15_Release_branch solves this issue for birt 4.16 which is using js.jar 1.7.15. Maybe there is a more elegant solution to do this?

speckyspooky:
I have reserached the location of the security manager on the BIRT project.

But I cannot change something to handle an issue based from another project
when I have to use deprecated classes to make functionality available and later to remove it again.
We go ahead step by step to support higher JDK version so the next version will be LTS JDK 21
and so it is more important to remove the deprecated stuff.

May be your fix solved your problem. I started a draft-PR to remove the security manager.
But this will not solve your problem.

len-ro:
[...]

PR #1780, opened 2024-07-03
Remove the deprecated security manager and add a JVM-birt system property for activation (#1778)

speckyspooky:
Remove the deprecated security manager and add a JVM-birt system property for activation (#1778)

### DIFF
--- a/core/org.eclipse.birt.core/src/org/eclipse/birt/core/script/JavascriptEvalUtil.java
+++ b/core/org.eclipse.birt.core/src/org/eclipse/birt/core/script/JavascriptEvalUtil.java
@@ -1,6 +1,5 @@
-/*
- *************************************************************************
- * Copyright (c) 2005 Actuate Corporation.
+/**************************************************************************
+ * Copyright (c) 2005, 2024 Actuate Corporation and others
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -12,8 +11,7 @@
  * Contributors:
  *  Actuate Corporation  - initial API and implementation
  *
- *************************************************************************
- */
+ **************************************************************************/
 package org.eclipse.birt.core.script;
 
 import java.io.File;
@@ -48,6 +46,12 @@ import org.mozilla.javascript.Wrapper;
  */
 public class JavascriptEvalUtil {
 	private static Logger logger = Logger.getLogger(JavascriptEvalUtil.class.getName());
+
+	/** System property of the JavaScript version */
+	private static final String ECMA_SCRIPT_SECURITY_PROPERTY_KEY = "birt.ecmascript.security"; //$NON-NLS-1$
+
+	/** Valid keys of the system property */
+	private static final String ECMA_SCRIPT_SECURITY_ENABLED = "on"; //$NON-NLS-1$
 
 	/*
 	 * LRU cache for compiled scripts. For performance reasons, scripts are compiled
@@ -372,7 +376,7 @@ public class JavascriptEvalUtil {
 	}
 
 	private static Object getSecurityDomain(final String file) {
-		if ((file == null) || (System.getSecurityManager() == null)) {
+		if ((file == null) || !isECMAScriptSecurityEnabled()) {
 			return null;
 		}
 		try {
@@ -386,4 +390,18 @@ public class JavascriptEvalUtil {
 		}
 	}
 
+	/**
+	 * Evaluate the system property to use the JavaScript security based on
+	 * certificates
+	 */
+	private static boolean isECMAScriptSecurityEnabled() {
+		boolean scriptSecurity = false;
+		/* System property: -Dbirt.ecmascript.security */
+		String configuredEcmaScriptSecurityProperty = System.getProperty(ECMA_SCRIPT_SECURITY_PROPERTY_KEY);
+		if (configuredEcmaScriptSecurityProperty != null
+				&& configuredEcmaScriptSecurityProperty.equalsIgnoreCase(ECMA_SCRIPT_SECURITY_ENABLED)) {
+			scriptSecurity = true;
+		}
+		return scriptSecurity;
+	}
 }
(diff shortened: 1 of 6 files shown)


speckyspooky:
This PR is a suggestion to remove the deprecated "Security Manager" since JDK 17.
Instead of the Security Manager a system property via JVM configuration can be used for activation.

This change is based on ticket #1778. But the PR won't fix the problem of the original ticket.
And currently I see no fix of the original topic which is located on mozilla -JavaScript.

The target here is only to remove the deprecated "Security Manager".

core/org.eclipse.birt.core/src/org/eclipse/birt/core/script/JavascriptEvalUtil.java
```diff
@@ -386,4 +390,18 @@ private static Object getSecurityDomain(final String file) {
...
+	private static boolean isECMAScriptSecurityEnabled() {
+		boolean scriptSecurity = false;
+		/* System property: -Dbirt.ecmascript.security */
+		String configuredEcmaScriptSecurityProperty = System.getProperty(ECMA_SCRIPT_SECURITY_PROPERTY_KEY);
+		if (configuredEcmaScriptSecurityProperty != null
+				&& configuredEcmaScriptSecurityProperty.equalsIgnoreCase(ECMA_SCRIPT_SECURITY_ENABLED)) {
+			scriptSecurity = true;
+		}
+		return scriptSecurity;
+	}
```
claesrosell, line 406: While this is fully functional, I would probably go for "true" / "false" here instead ( and alter the name of the property to something like: "birt.ecmascript.enable-security"). This would potentially reduce the code to:

return Boolean.getBoolean(ECMA_SCRIPT_ENABLE_SECURITY_PROPERTY_KEY); // getBoolean actually gets the value from System.getProperty()

data/org.eclipse.birt.report.engine.script.javascript/src/org/eclipse/birt/report/engine/javascript/ScriptUtil.java
```diff
@@ -22,14 +22,31 @@
...
 
+	/**
+	 * Get the security domain
+	 *
+	 * @param file JavaScript file
+	 * @return the secure JavaScript code
+	 */
 	public static Object getSecurityDomain(final String file) {
-		if ((file == null) || (System.getSecurityManager() == null)) {
+		if ((file == null) || !(new JavascriptVersion()).isECMAScriptSecurityEnabled()) {
```
claesrosell, line 49: I do not know enough about the SecurityManager nor the deprecation of it, but as I understand it, this is a way to give Rhino "what it wants" without the presence of a SecurityManager. Would not that help in cases like #1780 ?

data/org.eclipse.birt.report.engine.script.javascript/src/org/eclipse/birt/report/engine/javascript/JavascriptVersion.java
```diff
@@ -84,6 +93,28 @@ public String getConfiguredECMAScriptVersion() {
...
+	 */
+	private void evaluateEcmaScriptSecurity() {
+		/* System property: -Dbirt.ecmascript.security */
+		String configuredEcmaScriptSecurityProperty = System.getProperty(ECMA_SCRIPT_SECURITY_PROPERTY_KEY);
+		if (configuredEcmaScriptSecurityProperty != null
+				&& configuredEcmaScriptSecurityProperty.equalsIgnoreCase(ECMA_SCRIPT_SECURITY_ENABLED)) {
+			this.configuredEcmaScriptSecurity = true;
+		}
+	}
+
```
claesrosell, line 117: Same as earlier, but I would prefeer these to be static methods / fields, since there are no way to alter them per JavaScriptVersion -instance.

claesrosell (commented):
I have left some comments, you are free to do what you want with them. Overall I think that this change make sense ,since the SecurityManager is soon to be removed.

speckyspooky:
@claesrosell
Thanks for your hints. This sounds very good and I will change the property handling and also the method calls to simplify the stuff.

Yes, we could add some code to set the security manager but then we have again deprecated code which should be reoved.

core/org.eclipse.birt.core/src/org/eclipse/birt/core/script/JavascriptEvalUtil.java
speckyspooky: Is changed.

data/org.eclipse.birt.report.engine.script.javascript/src/org/eclipse/birt/report/engine/javascript/JavascriptVersion.java
speckyspooky: Is changed completely.

data/org.eclipse.birt.report.engine.script.javascript/src/org/eclipse/birt/report/engine/javascript/ScriptUtil.java
speckyspooky: Changed to static access.

data/org.eclipse.birt.report.engine.script.javascript/src/org/eclipse/birt/report/engine/javascript/README.md
```diff
@@ -24,6 +24,19 @@ The configuration will be done as a global starting parameter of the JVM.
...
+**birt.ecmascript.security.enabled**
+
+	Content    	configuration to enable or disable the JavaScript security based on certificates (as replacement of the Security Manager)
+	Parameter  	-Dbirt.ecmascript.security.enabled
+	Location   	JVM
+	Data type  	string
+	Values     	on, the JavaScript code use security mechanism certificate based
+	           	off, JavaScript without special JavaScript mechanism
+	Supported   	on, off
+	Default    	off
```
claesrosell, line 37: true / false

speckyspooky: typo changed

wimjongman:
IBM i?

PR #1793, opened 2024-07-15
Use jgit as the timestamp provider

merks:


merks:
I want to first do a test build where a dirty worktree will cause an error so I can be sure that nothing in the build is currently dirtying the tree.

merks:
Included in https://github.com/eclipse-birt/birt/pull/1794

PR #1794, opened 2024-07-15
Remove references to non-BIRT bundles from features

merks:
- Use jgit timestamp provider for product qualifiers.
- Add org.apache.batik.ext.awt.image.codec.imageio where org.apache.batik.transcoder.image is used.
- Reduce the content of the all-in-one product, mostly to remove redundant includes.
- Ensure the runtime product is complete enough after the features have been simplified such tha the OSGi runtime test still works.

### DIFF
--- a/UI/org.eclipse.birt.report.designer.core/META-INF/MANIFEST.MF
+++ b/UI/org.eclipse.birt.report.designer.core/META-INF/MANIFEST.MF
@@ -25,6 +25,7 @@ Require-Bundle: org.eclipse.ui;bundle-version="[3.2.0,4.0.0)",
  org.eclipse.birt.report.model;bundle-version="[2.1.0,5.0.0)"
 Bundle-ActivationPolicy: lazy
 Bundle-RequiredExecutionEnvironment: JavaSE-11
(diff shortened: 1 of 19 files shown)


merks:
@wimjongman  @speckyspooky 

Guys, this change effectively removes all references to 3rd party bundles as well as references to all bundles and feature from other projects from BIRT's features.  This ensures that BIRT does not lock in any specific versions.   The requires of bundles are sufficient to ensure that the necessary requirements are installed.

Also, with this change, I think using the jgit timestamp provider will be an improvement, i.e., builds will produce a new version of a bundle only when there is a commit to the bundle and for features only if there is a commit to anything included bundle or feature included the feature (which is now only other BIRT bundles and features).  Most projects are using the jgit timestamp provider to avoid producing all new artifacts with each build.

Do you have any concerns?

speckyspooky:
Yes, I confirm your changes, It sounds good and also your test results.
So each shrink of the all-in-one-product ist welcome.

wimjongman:
Thank you, Ed! Brilliant.

Issue #1800, opened 2024-07-17
Issue on Double-Byte-Characters: text-parameters do not display correctly on PDF/XLS/XLSX/DOCX (e.g. CN, GR, DE-Umlaut)


bossyang:
BIRT WebViewer 4.16

When report parameter value contains Chinese character, the dataset does not return data. 

In the following example, let's say parameter value is "积极会员" . 
It will work in HTML format.
![BIRT-20240717-003](https://github.com/user-attachments/assets/03f7cb57-a27c-4a4c-8bf3-f3c510d8d020)

But fail in pdf/xls/docx format.

![BIRT-20240717-002](https://github.com/user-attachments/assets/f9c87c18-88c1-49b0-bcc0-cc145caee40e)

speckyspooky:
I have tested your description with BIRT 4.16 standardinstellation MS Windows 11.
The result looks good on my side for HTML, PDF, DOCX & XLSX.

Attached you will find my test-report and the according results. Can you please execute my report to see the results on your BIRT-version.

Additional on it, can you provide an example from your side with the problem but without external dependencies and external data sources.

### Grafik of my pdf result
![grafik](https://github.com/user-attachments/assets/ae4773ac-5fa3-4350-ac39-451633187761)

### Test results
[language-chinese.zip](https://github.com/user-attachments/files/16273013/language-chinese.zip)

speckyspooky:
Of course, you are right and there is really a problem with all double-byte-characters according your description,
with the run types **"run"**, **"preview"** the UTF-8-characters won't be handled correctly.

If the run type **"frame set"** is used, then all looks fine.

The problem exists in BIRT: 4.14, 4.15, 416
(Currently I'm not able to test older versions and the current build.)


![double-byte-cha
[languages_2byte.zip](https://github.com/user-attachments/files/16310370/languages_2byte.zip)
racters-finding-BIRT-4 14-4 15-4 16](https://github.com/user-attachments/assets/9a32ece7-d310-4795-bede-d1c60f9a6b5a)

### New version with different Double-Byte-Languages
[languages_2byte.zip](https://github.com/user-attachments/files/16310371/languages_2byte.zip)

speckyspooky:
@wimjongman , @hvbtup , @merks 

Hi guys,
here we have a problem (I doesn't like such kind of problems) and I know that this has to do with the double-byte-characters and the UTF-8 handling.
[...]

PR #1805, opened 2024-07-19
Change parameter encoding from 8859-1 to UTF-8 (#1800)

speckyspooky:


merks (approved):
Definitely best to use an encoding that can encode any Unicode character.

PR #1807, opened 2024-07-20
Include website clone in the project setup

merks:
Fixes https://github.com/eclipse-birt/birt/issues/1320


merks:
This has no build impact, so merging immediately.

PR #1814, opened 2024-07-21
Enhance comments, raw types and other small hints to reduce java warnings

speckyspooky:
The PR includes some changes to reduce the java warnings based on optimization of:
- java documentation
- use raw types with type operator
- replace deprecated methods

### DIFF
--- a/engine/uk.co.spudsoft.birt.emitters.excel/src/uk/co/spudsoft/birt/emitters/excel/ExcelEmitter.java
+++ b/engine/uk.co.spudsoft.birt.emitters.excel/src/uk/co/spudsoft/birt/emitters/excel/ExcelEmitter.java
@@ -408,7 +408,11 @@ public abstract class ExcelEmitter implements IContentEmitter {
 				}
 			}
 			if (handlerState.getWb() instanceof SXSSFWorkbook) {
-				((SXSSFWorkbook) handlerState.getWb()).dispose();
+				try {
+					((SXSSFWorkbook) handlerState.getWb()).close();
+				} catch (IOException ex) {
+					log.debug("ex:", ex.toString());
+				}
 			}
 			handlerState = null;
 			reportOutputFilename = null;
--- a/model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/DesignElementHandle.java
+++ b/model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/DesignElementHandle.java
@@ -70,7 +70,7 @@ import org.eclipse.birt.report.model.elements.MultiViews;
 import org.eclipse.birt.report.model.elements.ReportDesign;
 import org.eclipse.birt.report.model.elements.TemplateElement;
 import org.eclipse.birt.report.model.elements.interfaces.IDesignElementModel;
-import org.eclipse.birt.report.model.elements.interfaces.IReportItemModel;
+import org.eclipse.birt.report.model.elements.interfaces.IInternalReportItemModel;
 import org.eclipse.birt.report.model.elements.interfaces.IStyleModel;
 import org.eclipse.birt.report.model.i18n.MessageConstants;
 import org.eclipse.birt.report.model.metadata.ElementDefn;
@@ -444,12 +444,10 @@ public abstract class DesignElementHandle implements IDesignElementModel {
 	/**
 	 * Gets the value of a property as a list.
 	 *
-	 * @param module   the module
 	 * @param propName the name of the property to get
 	 * @return the value as an <code>ArrayList</code>, or null if the property is
 	 *         not set or the value is not a list
 	 */
-
 	public List getListProperty(String propName) {
 		PropertyHandle propHandle = getPropertyHandle(propName);
 		return propHandle == null ? null : propHandle.getListValue();
@@ -520,7 +518,7 @@ public abstract class DesignElementHandle implements IDesignElementModel {
 	 */
 
 	public void setFloatProperty(String propName, double value) throws SemanticException {
-		setProperty(propName, new Double(value));
+		setProperty(propName, Double.valueOf(value));
 	}
 
 	/**
@@ -571,14 +569,13 @@ public abstract class DesignElementHandle implements IDesignElementModel {
 	 */
 
 	public void clearAllProperties() throws SemanticException {
-		List props = getDefn().getProperties();
+		List<IElementPropertyDefn> props = getDefn().getProperties();
 
 		ActivityStack stack = module.getActivityStack();
 		stack.startTrans(CommandLabelFactory.getCommandLabel(MessageConstants.CLEAR_PROPERTIES_MESSAGE));
 
 		try {
-			for (int i = 0; i < props.size(); i++) {
-				PropertyDefn propDefn = (PropertyDefn) props.get(i);
+			for (IElementPropertyDefn propDefn : props) {
 				String propName = propDefn.getName();
 
 				if ((IDesignElementModel.NAME_PROP.equals(propName))) {
@@ -681,13 +678,13 @@ public abstract class DesignElementHandle implements IDesignElementModel {
 	 * @return List contains the methods.
 	 *
 	 */
-	public List getMethods() {
+	public List<IElementPropertyDefn> getMethods() {
 		List<IElementPropertyDefn> methods = getElement().getDefn().getMethods();
 		if (getElement().isInSlot(IModuleModel.PAGE_SLOT)) {
 			// Added for bugzila 276665, filter the method "onPageBreak" if the
 			// element is contained by a master page.
 			for (IElementPropertyDefn method : methods) {
-				if (IReportItemModel.ON_PAGE_BREAK_METHOD.equals(method.getName())) {
+				if (IInternalReportItemModel.ON_PAGE_BREAK_METHOD.equals(method.getName())) {
 					methods.remove(method);
 					break;
 				}
@@ -705,14 +702,11 @@ public abstract class DesignElementHandle implements IDesignElementModel {
 	 *                           value is invalid.
 	 * @see #setProperty(String, Object)
 	 */
-
-	public void setProperties(Map properties) throws SemanticException {
+	public void setProperties(Map<?, ?> properties) throws SemanticException {
 		if (properties == null) {
 			return;
 		}
-
-		for (Iterator iter = properties.entrySet().iterator(); iter.hasNext();) {
-			Entry entry = (Entry) iter.next();
+		for (Entry<?, ?> entry : properties.entrySet()) {
 			setProperty((String) entry.getKey(), entry.getValue());
 		}
 	}
@@ -1054,10 +1048,9 @@ public abstract class DesignElementHandle implements IDesignElementModel {
 			PropertyHandle propHandle = propHandles.get(propName);
 			if (propHandle != null) {
 				return propHandle;
-			} else {
-				cachePropertyHandles();
-				return propHandles.get(propName);
 			}
+			cachePropertyHandles();
+			return propHandles.get(propName);
 		}
 
 		return new PropertyHandle(this, propDefn);
@@ -1311,12 +1304,10 @@ public abstract class DesignElementHandle implements IDesignElementModel {
 	 *
 	 * @return an iterator over the properties. Each call to <code>getNext( )</code>
 	 *         returns an object of type {@link PropertyHandle}.
-	 * @see PropertyIterator
 	 * @see PropertyHandle
 	 * @see UserPropertyDefnHandle
 	 */
-
-	public Iterator getPropertyIterator() {
+	public PropertyIterator getPropertyIterator() {
 		return new PropertyIterator(this);
 	}
 
@@ -1364,7 +1355,6 @@ public abstract class DesignElementHandle implements IDesignElementModel {
 	 *         {@link DesignElementHandle}. Nothing will be iterated over an element
 	 *         that is not <code>ReferenceableElement</code>.
 	 */
-
 	public Iterator clientsIterator() {
 		return new ClientIterator(this);
 	}
@@ -1441,8 +1431,7 @@ public abstract class DesignElementHandle implements IDesignElementModel {
 	 *
 	 * @param list the list to sort
 	 */
-
-	public static void doSort(List list) {
+	public static void doSort(List<DesignElementHandle> list) {
 		Collections.sort(list, new Comparator<DesignElementHandle>() {
 
 			@Override
@@ -1608,10 +1597,9 @@ public abstract class DesignElementHandle implements IDesignElementModel {
 		case IPropertyType.LIST_TYPE:
 
 			assert value instanceof List;
-			List valueList = (List) value;
+			List<?> valueList = (List<?>) value;
 			PropertyHandle propHandle = targetHandle.getPropertyHandle(propName);
-			for (int i = 0; i < valueList.size(); i++) {
-				Object item = valueList.get(i);
+			for (Object item : valueList) {
 				if (propDefn.getSubTypeCode() != IPropertyType.ELEMENT_REF_TYPE) {
 					propHandle.addItem(item);
 				} else {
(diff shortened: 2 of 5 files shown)


merks:
FYI. It’s hard to review from my phone. I’ll look tomorrow.

engine/uk.co.spudsoft.birt.emitters.excel/src/uk/co/spudsoft/birt/emitters/excel/ExcelEmitter.java
```diff
@@ -408,7 +408,11 @@ public void end(IReportContent report) throws BirtException {
 				}
 			}
 			if (handlerState.getWb() instanceof SXSSFWorkbook) {
-				((SXSSFWorkbook) handlerState.getWb()).dispose();
+				try {
+					((SXSSFWorkbook) handlerState.getWb()).close();
+				} catch (IOException ex) {
+					log.debug("ex:", ex.toString());
```
merks, line 414: If there is an exception, I thought it's better to call a log method that shows the stack trace of the exception.

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/DesignElementHandle.java
```diff
@@ -705,14 +703,13 @@ public List getMethods() {
...
-
-	public void setProperties(Map properties) throws SemanticException {
+	public void setProperties(Map<?, ?> properties) throws SemanticException {
 		if (properties == null) {
 			return;
 		}
 
-		for (Iterator iter = properties.entrySet().iterator(); iter.hasNext();) {
-			Entry entry = (Entry) iter.next();
+		for (Iterator<?> iter = properties.entrySet().iterator(); iter.hasNext();) {
```
merks, line 711: A for-each loop would be much less verbose.  Also, the entrySet is already of type `Entry<?,?>` so this use of an iterator of type `<?>`  loses the type and then you need a cast below.

```
for (Map.Entry<?, ?> entry : properties.entrySet()) {
}
```
```

File: model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/DesignElementHandle.java [hunk 3]
```diff
@@ -571,7 +569,7 @@ public void clearProperty(String propName) throws SemanticException {
 	 */
 
 	public void clearAllProperties() throws SemanticException {
-		List props = getDefn().getProperties();
+		List<?> props = getDefn().getProperties();
```
merks, line 572: Here too, type info is being lost.  

Something like this would seem better:
```
	public void clearAllProperties() throws SemanticException {
		List<IElementPropertyDefn> props = getDefn().getProperties();

		ActivityStack stack = module.getActivityStack();
		stack.startTrans(CommandLabelFactory.getCommandLabel(MessageConstants.CLEAR_PROPERTIES_MESSAGE));

		try {
			for (IElementPropertyDefn propDefn : props) {
				String propName = propDefn.getName();
```

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/DesignElementHandle.java
```diff
@@ -1054,10 +1051,9 @@ public PropertyHandle getPropertyHandle(String propName) {
 			PropertyHandle propHandle = propHandles.get(propName);
 			if (propHandle != null) {
 				return propHandle;
-			} else {
-				cachePropertyHandles();
-				return propHandles.get(propName);
 			}
+			cachePropertyHandles();
+			return propHandles.get(propName);
```
merks, line 1056: There the `propHandle` variable already has already gotten the value.

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/DesignElementHandle.java
```diff
@@ -1608,7 +1600,7 @@ public void copyPropertyTo(String propName, DesignElementHandle targetHandle) th
 		case IPropertyType.LIST_TYPE:
 
 			assert value instanceof List;
-			List valueList = (List) value;
+			List<?> valueList = (List<?>) value;
```
merks, line 1603: A for-each loop below would be simpler/clearer.

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/DesignElementHandle.java
```diff
@@ -1639,7 +1631,7 @@ public void copyPropertyTo(String propName, DesignElementHandle targetHandle) th
 	 * @see ErrorDetail
 	 */
 
-	public List semanticCheck() {
+	public List<?> semanticCheck() {
```
merks, line 1634: This method should really be like this:
```
	public List<ErrorDetail> semanticCheck() {
		// Validate this element.

		List<SemanticException> exceptionList = getElement().validate(module);
		List<ErrorDetail> errorDetailList = ErrorDetail.convertExceptionList(exceptionList);

		return errorDetailList;
	}
```

But then the ErrorDetail method should be more accepting on any type of list of exceptions like this;
```
	public static List<ErrorDetail> convertExceptionList(List<? extends Exception> exceptionList) {
		List<ErrorDetail> errorDetailList = new ArrayList<>();

		Iterator<? extends Exception> iterError = exceptionList.iterator();
		while (iterError.hasNext()) {
			Exception e = iterError.next();

			ErrorDetail errorDetail = new ErrorDetail(e);
			errorDetailList.add(errorDetail);
		}

		return errorDetailList;
	}
```

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/DesignElementHandle.java
```diff
@@ -2205,7 +2196,7 @@ public List paste(String propName, DesignElementHandle content) throws SemanticE
 	 * @return a list containing all errors for the pasted element
 	 * @throws SemanticException if the element is not allowed to paste
 	 */
-	public List paste(String propName, IDesignElement content) throws SemanticException {
+	public List<?> paste(String propName, IDesignElement content) throws SemanticException {
```
merks, line 2199: All these paste method return a result like that of `List<ErrorDetail> convertExceptionList` so this return type seems unhelpful.

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/ReportDesignHandleImpl.java
```diff
@@ -265,9 +265,9 @@ public SlotHandle getScratchPad() {
 	 * @return the iterator of included scripts.
 	 */
 
-	public Iterator includeLibraryScriptsIterator() {
-		List libList = module.getAllLibraries();
-		List includeLibScriptList = new ArrayList();
+	public Iterator<Object> includeLibraryScriptsIterator() {
```
merks, line 268: This doesn't compile, but the Javadoc says the element type IncludeScriptHandle so if you're going to parameterize it, better confirm to the expectations set by the Javadoc:
```
	public Iterator<IncludeScriptHandle> includeLibraryScriptsIterator() {
		List<Library> libList = module.getAllLibraries();
		List<Object> includeLibScriptList = new ArrayList<Object>();
		if (libList != null) {
			for (Library lib : libList) {
				PropertyHandle propHandle = lib.getHandle(lib).getPropertyHandle(INCLUDE_SCRIPTS_PROP);

				Iterator scriptIter = propHandle.iterator();
				while (scriptIter.hasNext()) {
					includeLibScriptList.add(scriptIter.next());
				}
			}
		}
		return includeLibScriptList.iterator();
	}
```

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/ReportDesignHandleImpl.java
```diff
@@ -378,19 +378,18 @@ public void setRefreshRate(int rate) {
...
 	 * @return each item is <code>CssStyleSheetHandle</code>
 	 */
-
-	public List getAllCssStyleSheets() {
+	public List<CssStyleSheetHandle> getAllCssStyleSheets() {
 		ReportDesign design = (ReportDesign) getElement();
-		List allStyles = new ArrayList();
-		List csses = design.getCsses();
+		List<CssStyleSheetHandle> allStyles = new ArrayList<CssStyleSheetHandle>();
+		List<?> csses = design.getCsses();
```
merks, line 392: The method is `	public final List<CssStyleSheet> getCsses() {` so you should not need casting and can use a for-each loop.

merks (commented):
There is still quite a bit of room for improvement and some of the return type especially are poor.

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/DesignElementHandle.java
speckyspooky: I didn't understand what your suggestion is. The most suggestions are comming from the eclipse-java-side as solving hint.

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/DesignElementHandle.java
speckyspooky: I solved only the warning to make small changes not to reenigeer the variables.

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/DesignElementHandle.java
speckyspooky: What is the suggestion?

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/DesignElementHandle.java
speckyspooky: I haven't changed it because it would cause more changes according to your description of "convertExceptionList()" therefore I haven't changed all.

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/ReportDesignHandleImpl.java
speckyspooky: I will take a look into it.

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/DesignElementHandle.java
speckyspooky: Typ info is lost: before it was "List", now it is "List<?>" where is the type lost?

I will try your suggestion.

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/DesignElementHandle.java
merks: Here is the full replacement:
```
	public void setProperties(Map<?, ?> properties) throws SemanticException {
		if (properties == null) {
			return;
		}

		for (Entry<?, ?> entry : properties.entrySet()) {
			setProperty((String) entry.getKey(), entry.getValue());
		}
	}
```
It's extremely rare that you should need to use an Iterator anywhere, i.e., only if you need to call remove on it.

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/DesignElementHandle.java
merks: Yes, this was already poor.  When touching it, might as well make it less poor.  But that's your choice.

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/DesignElementHandle.java
speckyspooky: I can try it but if there no success I will full reset the change back to version with warnings.

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/DesignElementHandle.java
merks:
```
			assert value instanceof List;
			List<?> valueList = (List<?>) value;
			PropertyHandle propHandle = targetHandle.getPropertyHandle(propName);
			for (Object item : valueList) {
```

Again, it's your option.

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/DesignElementHandle.java
speckyspooky: I will try it and I understand that you reference to the change of "convertExceptionList()".
But at my testing I got conflicts with the methods which are using the "paste()", so I will see after the change.

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/DesignElementHandle.java
merks: Note that the return type of List<ErrorDetail> errorDetailList = ErrorDetail.convertExceptionList(exceptionList) is already correct and can be used, even if leave exceptionList  as a raw type.

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/DesignElementHandle.java
speckyspooky: I confirm your thinking.

But the "cachePropertyHandles" can change the propHandles and therefore the get-Method will be use again instead of the "propHandle". So for me it's ok, what would the improvement from your side?

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/ReportDesignHandleImpl.java
speckyspooky: I will remove the type from the Iterator
because it causes a converting issue, but I will add your suggestion for the better loop.

grafik

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/DesignElementHandle.java
merks: So it might be a different object than what was in the map before?  In that case it’s surprising but fine.

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/ReportDesignHandleImpl.java
merks: I would do an instance of test in the loop. 

My primary concern is that once you replace the raw type with a parameterized type it becomes very hard to correct it in the future.

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/DesignElementHandle.java
speckyspooky: Loop is changed.

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/DesignElementHandle.java
speckyspooky: Change is done.

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/DesignElementHandle.java
speckyspooky: Change is done.

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/ReportDesignHandleImpl.java
speckyspooky: Loop is changed, raw type without parameter (like original version).

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/ReportDesignHandleImpl.java
speckyspooky: Change is done.

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/api/DesignElementHandle.java
speckyspooky: Change is done.

merks (approved):
Thanks. 🙏

PR #1820, opened 2024-07-23
Ensure that org.eclipse.birt.build.ant.PackTask filters out CODESIGN.*

merks:
https://github.com/eclipse-birt/birt/issues/1819


merks:
I've manually re-run with a modified Jenkinsfile configures the PR build to do signing.

PR #1821, opened 2024-07-24
Reduce org.eclipse.birt.engine.runtime content

merks:
- Remove org.eclipse.rcp to eliminate UI dependencies.
- Add a minimized set of additional plugins.
- Remove UI bundles from the org.eclipse.birt.chart.osgi.runtime feature.
- Remove org.eclipse.core.resources from org.eclipse.birt.report.viewer.

### DIFF
--- a/build/birt-packages/birt-runtime-osgi/reportengine.product
+++ b/build/birt-packages/birt-runtime-osgi/reportengine.product
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?pde version="3.5"?>
 
-<product name="BIRT Engine Runtime" uid="org.eclipse.birt.engine.runtime" version="4.17.0.qualifier" type="features" includeLaunchers="false" autoIncludeRequirements="true">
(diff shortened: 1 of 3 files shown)


merks:
@speckyspooky  @wimjongman  @hvbargen @SteveSchafer-Innovent 

Guys, this significantly reduces the content of the two runtimes; about ~15-20% smaller in bytes.  Both runtimes pass their basic tests.

The /ReportEngine/lib has 82 jars instead of 141.  The /birt.war/WEB-INF/lib/ has 107 jars instead of 166.

Any concerns before I merge?

---  

FYI, I'm still suspicious out the SWT dependencies.  Only the win32 fragment is present so I'm doubtful this can work across different operating system and no one has complained about that.

speckyspooky:
It sounds very good to me to reduce the dependencies and shrink the runtimes.
After the merge we should test the runtimes integrated at application server to verify the working of our standard reports.

wimjongman:
Nice. Please merge.

PR #1822, opened 2024-07-25
Improve org.eclipse.birt.engine.runtime content

merks:
- Remove SWT dependencies because only the win32 fragment has been available so this can't have worked across operating systems.
- Ensure the runtime is based on the latest orbit aggregation dependencies, except for derby.
- Specify org.apache.commons.commons-logging in DeploymentRuntime category.xml and in org.eclipse.birt.report.data.oda.excel.
- Improve the runtime_readme.txt to remove the out-dated dependency descriptions and to use current https URLs.
- Remove org.eclipse.birt.chart.device.swt from the org.eclipse.birt.chart.osgi.runtime feature.
- Remove org.eclipse.pde.ui from org.eclipse.birt.report.designer.ui.ide.


merks:
@speckyspooky  @wimjongman 

This reduces the runtime content to the following where only the derby dependency does not come from the current orbit aggregation.

image

I really think removing the swt dependencies is the right thing to do...

Any concerns?

wimjongman:
I think SWT is used for the carting component.

merks:
Yes, it's used in the IDE to render charts and it's still in the charts feature and still installed in the IDE:

image

image

In the runtime though I expect charts are rendered to SVG and/or PDF and not to SWT widgets.  Moreover, it can't possibly have worked in the runtime in general in the past because there has never been linux or macos fragments included (and two of each for the two support architectures).  Even if it those were included, in the non-OSGi runtime it's a flat classpath, so one cannot possibly include both or many and expect the correct one be chosen and to work correctly.  So, as far as I can tell, it always will be and always has been completely bogus to include it in the runtimes.  Of course the entire workbench was previously included in the runtime which also completely bogus.  I expect no one complains about bogus junk but rather only note a problem if something is missing because something doesn't work.

wimjongman:
I'm fine to merge, but the runtime readme is explicit about it:

https://github.com/eclipse-birt/birt/blob/57330a797708ac686f1fcffcd107b55b39a4c2a3/build/org.eclipse.birt.releng.birtbuilder/extras/runtime_readme.txt

merks:
Looking back the org.eclipse.birt.chart.device.swt bundle was first added in 4.14 and was not present in 4.13 or earlier:

image

The org.eclipse.birt.chart.device.pdf was added at this point as well, which to me makes sense, but the SWT one does not.

wimjongman:
Ok, let's merge and test some reports when the nightly is available.

wimjongman (approved):

speckyspooky:
If you find more and more then we will get a BIRT-slim-version ;o)

merks:
Guys, if I manage to create a new bundlized version of derby, how can I test/validate that it work corrects?  Please assume that I"m really stupid and that nothing which might seem obvious to you will actually be obvious to me...

wimjongman:
In BIRT all-in-one? I have a getting started video here:

https://www.youtube.com/watch?v=WaRNFVlSMo0&t=296s

hvbtup:
You can perform at least a basic test if for example you create a new data source, choose the Classic Models Inc Sample DataSource,
then add a JDBC dataset, type `select * from productlines` and drag the dataset into the layout.
This should give you some columns, two of them aren't rendered correctly which is normal:
The column HTMLDESCRIPTION cannot be rendered as a data item. Delete the data item, and insert a "dynamic text item" instead in the layout. Choose `row["HTMLDESCRIPTION"]` as expression. BIRT should now show the HTML correctly formatted.

For the image, it is slightly more complicated, so I'm not going to explain here. The data is shown as hex values.

But for the sake of testing basic Derby functionality, this should suffice IMHO.
It demonstrates that queries can be executed and that BLOB columns work.

merks:
Thanks for the info guys.  This is super helpful!!

wimjongman:
Drop this in a project and run it. Remove the txt extension.

[new_report.rptdesign.txt](https://github.com/user-attachments/files/16389124/new_report.rptdesign.txt)

merks:
FYI, it looks very promising that this "specially rewrapped" version just works as a drop-in upgrade:

image

speckyspooky:
@merks
The report looks good.

I see another point at your screen where I not know if it is wanted or not.
The register at the lefthand side are without the white background in compare to BIRT 4.16

Do you know if it is "normal" / "wanted" with eclipse 2024-09?

**Compare-Screen**
grafik

merks:
That's the "new improved modern styling" from the platform:

https://github.com/eclipse-platform/eclipse.platform.ui/issues/2114

speckyspooky:
Ahhh, ohhh, it's not a bug its a feature.

PR #1826, opened 2024-08-02
Switch CrosstabXYLayoutEditPolicy to use TableXYLayoutEditPolicy

merks:
https://github.com/eclipse-birt/birt/issues/1824


merks:
Guys, this fixes the class cast problem but I have no clue if the behavior is correct.

speckyspooky (approved):
Replacement is fine. Changed approved.

speckyspooky:
The general topic is if the new policy won't work correctly then we will have problems with the eclipse layout designer
and we wouldn't be able to create any report. So if we can design new reports with the commen items then it is fine.
Internally it will be used to draw the 2d graphics but this is really deep knowledge of the designer implementation.

The original problem was the change of the policy which was more restrictive on class cast side
so I hope that the new policy provide is only an improvement without special changes. We will see.

merks:
I merged so we get a new build the user can test.

PR #1829, opened 2024-08-03
Update the target platform to use the new version of derby

merks:
- Disable tests that fail with the new target platform for later investigation.

https://github.com/eclipse-birt/birt/issues/1828

### DIFF
--- a/build/org.eclipse.birt.releng/BIRT.setup
+++ b/build/org.eclipse.birt.releng/BIRT.setup
@@ -539,8 +539,6 @@
         <repository
             url="https://download.eclipse.org/tools/orbit/simrel/orbit-aggregation/nightly/latest"/>
         <repository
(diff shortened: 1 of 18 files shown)


merks:
@SteveSchafer-Innovent 

Steve, with this change I believe that all the libraries of both the runtimes originate from the latest orbit content:

https://download.eclipse.org/tools/orbit/simrel/orbit-aggregation/table.html

The above table has links to Maven Central.

The derby version is created by Orbit as described in this issue:

https://github.com/eclipse-orbit/orbit-simrel/issues/39

It's equivalent to this version and I believe that jar would work on a flat classpath, i.e., as a pom dependency, but not in an OSGi runtime:

https://repo1.maven.org/maven2/org/apache/derby/derby/10.16.1.1/

Issue #1832, opened 2024-08-04
Enhancement, JDBC driver manager - configuration of the property validation


speckyspooky:
For the enhancement of #1825 the JDBC driver manager got a change to validate the supported driver properties (PR #1831, BIRT 4.17).
Through this change the support of the DuckDB driver is possible.

The target of this change is to give BIRT a fallback configuration and a small individual configuration to handle which driver(s) should be validated an which driver(s) should be excluden from the validation.

The reason of it is that the change of the JDBC driver manager is a central change and change the standard behavior.
The second reason is that it could be that the JDBC driver information of the supported properties are not full defined
and with this additional change it can be handled.

The following options will be given as *global system environment parameters* of the JVM:
- birt.driver.property.validation.enabled
o true, property validation is enabled (default)
o false, property validation is disabled

- birt.driver.property.validation.excluded.drivers
o unset, all drivers will be checked
o is set, the listed drivers will be excluded from the validation
o value, driver class name, for multiple driver listing a list splitted by comma separated class names can be used

** examples of global JVM parameters:**
- -Dbirt.driver.property.validation.enabled
o the validation is disabled

- -Dbirt.driver.property.validation.excluded.drivers=com.microsoft.sqlserver.jdbc.SQLServerDriver
o validation is enabled but without the driver of "SQLServerDriver"

- -Dbirt.driver.property.validation.excluded.drivers=org.duckdb.DuckDBDriver,com.microsoft.sqlserver.jdbc.SQLServerDriver
o validation is enabled but without the both drivers of "DuckDb" & "SQLServerDriver"

speckyspooky:
The PR #1833 is added for the change.

speckyspooky:
The change is merged to the master branch.

PR #1833, opened 2024-08-04
Enhanced JDBC driver manager with configuration based on new JVM parameters (#1832)

speckyspooky:


data/org.eclipse.birt.report.data.oda.jdbc/src/org/eclipse/birt/report/data/oda/jdbc/JDBCDriverManager.java
```diff
@@ -70,8 +71,17 @@ public class JDBCDriverManager {
...
 	private static final String EMPTY_STRING = ""; //$NON-NLS-1$
 
+	private boolean isDriverPropertyValidationEnabled = true;
+
+	private String excludedDriverForValidation;
+
+	private static final String DRIVER_PROPERTY_VALIDATION = "birt.driver.property.validation.enabled"; //$NON-NLS-1$
+
+	private static final String DRIVER_PROPERTY_VALIDATION_EXCLUDED = "birt.driver.property.validation.excluded.drivers"; //$NON-NLS-1$
+
```
merks, line 84: It would be nicer if all the statics came before all the non-statics and would be good the private statics and public statics remain segregated.

data/org.eclipse.birt.report.data.oda.jdbc/src/org/eclipse/birt/report/data/oda/jdbc/JDBCDriverManager.java
```diff
@@ -70,8 +71,17 @@ public class JDBCDriverManager {
 	public static final String JDBC_PASSWORD_PROP_NAME = "password"; //$NON-NLS-1$
 
 	private static final String INVALID_AUTH_SQL_STATE = "28000"; // X/Open SQL State //$NON-NLS-1$
+
 	private static final String EMPTY_STRING = ""; //$NON-NLS-1$
 
+	private boolean isDriverPropertyValidationEnabled = true;
+
+	private String excludedDriverForValidation;
```
merks, line 79: The property name "...excluded.drivers" suggests multiple exclusions are support.  This field suggest a single one is specified.  Later it just does a contains test on the String, so the separator doesn't matter.   If you were to write some documentation somewhere how to use system property to exclude multiple drivers from validation, how would you describe the syntax for it?

speckyspooky: Yes, I will create a documentation of it with a README.MD.
Yes, the delimitor is independend because the class name will be checked with "contain()" but I recommand ","-comma.

data/org.eclipse.birt.report.data.oda.jdbc/src/org/eclipse/birt/report/data/oda/jdbc/JDBCDriverManager.java
speckyspooky: Done.

merks (approved):
Nicer.

PR #1834, opened 2024-08-04
Remove UI dependencies from org.eclipse.birt.osgi.runtime

merks:
- Remove org.eclipse.birt.report.viewer and org.eclipse.birt.data.oda.mongodb.ui from org.eclipse.birt.osgi.runtime
 feature.
- Simplify test dependency configuration.
- The org.junit dependencies aren't optional and non-test bundles should not have such a dependency, even an optional one.

### DIFF
--- a/chart/org.eclipse.birt.chart.tests/pom.xml
+++ b/chart/org.eclipse.birt.chart.tests/pom.xml
@@ -35,22 +35,6 @@
 					<argLine>-Duser.country=us -Duser.language=en</argLine>
 				</configuration>
 			</plugin>
-            <plugin>
-				<groupId>org.eclipse.tycho</groupId>
-				<artifactId>target-platform-configuration</artifactId>
-				<version>${tycho.version}</version>
-				<configuration>
-					<dependency-resolution>
-						<extraRequirements>
-							<requirement>
-								<type>eclipse-feature</type>
-								<id>org.eclipse.birt</id>
-								<versionRange>0.0.0</versionRange>
-							</requirement>
-						</extraRequirements>
-					</dependency-resolution>
-				</configuration>
-			</plugin>
 		</plugins>
 	</build>
 </project>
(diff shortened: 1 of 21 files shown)


merks:
The org.eclipse.birt.osgi.runtime feature has nothing to do with the the two runtime zips that are produced, despite the name suggesting otherwise.  The runtime contents are defined by this product's features:

image

That being said, I assume the UI dependencies mean this only functions in the UI. 

image

The stack trace suggests it trying to load this load this class by name, though goodness knows why it don't just use the class directly:

image

In any case it appears axis is missing. but it's explicitly required:

image

The axis bundle is not present in the runtime, but I believe this from previous cleanups. Is this the first you've tested since the reduced the runtime content?

merks:
These are the only references I find to org.apache.axis, so it's not in the runtime because nothing else in the runtime requires it

image

I think this is the state after the change I did 3 weeks ago.

So two issue, maybe the viewer should remain in the runtime feature (not to be confused with the two other runtimes), but then I think the ui dependencies should all be optional if it's possible to use this without necessarily dragging in the entire Eclipse UI. Though I suppose you're using it with a flat classpath here so the OSGi dependencies are just ignore and hopefully/actually none of the UI classes need to actually be loaded...

speckyspooky:
You have written "...org.apache.axis,..because nothing else in the runtime requires it...".
But the BIRT-viewer requires the axis from technical side to be runnable.
I do not know why this isn't referenced directly.

merks:
What's not clear to me is where in the runtime zips the viewer itself is located.

In any case, as I mentioned, the runtime content is defined purely by the org.eclipse.birt.engine.runtime product's included features.  The viewer is not part of that.

It's important to observe again that anything that isn't tested isn't guaranteed to keep working so some type of test would be of significant value here.  I have no idea how to test what you are testing...

PR #1838, opened 2024-08-05
Documentation of the JDBC connection property validation

speckyspooky:


speckyspooky:
The documentation is part of the change for #1832

hvbtup (commented):
I don't have the time ATM, but a brief look shows that the documentation contains some typical mistakes of Germans speaking English, for example "must not" probably means "doesn't have to" here, etc..

PR #1839, opened 2024-08-06
Remove org.eclipse.birt.jetty.overlay

merks:


merks:
This bundle was not included in any feature nor product so was not in the p2 repository nor in the runtimes.

PR #1843, opened 2024-08-08
Only disable the failing rendering tests for windows

merks:
- Support the `system property birt.rendering.tests.enabled` set to `true` to enable the tests even on Windows and enable it in the IDE launchers where the tests function correctly.

https://github.com/eclipse-birt/birt/issues/1828


merks:
This is just an initial try just to see if these tests that fail for me in the Windows Maven build actually fail building in the servers on Linux.

merks:
This centralizes the condition checking and allows the tests to be easily re-enabled on Windows and in the IDE-launched tests.

PR #1860, opened 2024-08-18
Test org.eclipse.birt.axis.utility.UTF8Encoder is used and is correct

merks:
- Include org.eclipse.birt.axis.overlay in the org.eclipse.birt.engine.runtime feature.
- Test both runtimes that they use the overlay correctly.
- Reconcile the difference between the features org.eclipse.birt.osgi.runtime and org.eclipse.birt.engine.runtime.
  - Remove org.eclipse.birt.resources because it only provides cheatsheet content for in the IDE.
  - Add org.eclipse.birt.chart.device.pdf and org.eclipse.birt.report.engine.emitter.prototype.excel.

### DIFF
--- a/build/birt-packages/birt-runtime-test/src/test/java/org/eclipse/birt/sdk/BaseTestTemplate.java
+++ b/build/birt-packages/birt-runtime-test/src/test/java/org/eclipse/birt/sdk/BaseTestTemplate.java
@@ -13,6 +13,9 @@ package org.eclipse.birt.sdk;
 import java.io.File;
 import java.io.FilenameFilter;
 import java.io.IOException;
+import java.io.StringReader;
+import java.io.StringWriter;
+import java.io.Writer;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
 import java.net.URL;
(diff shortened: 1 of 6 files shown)


merks:
@speckyspooky 

It's bothered me for a while that I didn't know what org.eclipse.birt.axis.overlay actually does.  It was introduced without any tests.  So I investigated and indeed the Axis encoding of surrogate pairs is broken without the fix provided by org.eclipse.birt.axis.utility.UTF8Encoder so I added a test for that.  It also bothered me that the org.eclipse.birt.axis.overlay bundle (a fragment with org.apache.axis as its host) is not included in the runtime, so I added the same encoding test for both runtimes and modified the feature to include the fragment so that the tests pass; this inclusion transitively includes the host bundle as well then.

At this point there was almost no difference between the org.eclipse.birt.engine.runtime feature and the org.eclipse.birt.osgi.runtime feature so I removed org.eclipse.birt.resources (a UI thing) and add the missing plugins such that the two features are now identical.

I think at this point I can simplify the OSGi runtime product to use the org.eclipse.birt.osgi.runtime, removing the org.eclipse.birt.chart.osgi.runtime feature because all its includes are also in the org.eclipse.birt.osgi.runtime feature already.

I'll do that as a separate step...

speckyspooky:
@merks
Good to know that there are such kind of problems will fixed through such workaround.
But if it works then all is fine.

PR #1861, opened 2024-08-18
Simplify the org.eclipse.birt.engine.runtime product

merks:
- Include only the org.eclipse.birt.osgi.runtime feature.

### DIFF
--- a/build/birt-packages/birt-runtime-osgi/reportengine.product
+++ b/build/birt-packages/birt-runtime-osgi/reportengine.product
@@ -173,8 +173,7 @@ United States, other countries, or both.
    </plugins>
 
    <features>
-      <feature id="org.eclipse.birt.engine.runtime"/>
-      <feature id="org.eclipse.birt.chart.osgi.runtime"/>
+      <feature id="org.eclipse.birt.osgi.runtime"/>
    </features>
 
    <configurations>


merks:
@speckyspooky 

I believe this produces the same runtime contents as before.  It's more logical and simpler that runtime content is defined by the `org.eclipse.birt.osgi.runtime` feature.

speckyspooky:
@merks 
I like your simplefication and after the build I can test the OSGi-version with my tomcat.

merks:
@speckyspooky 

I interpret silence as "I tested it and it there are no problems.  

I'd like to build our first milestone soon based on the SimRel m3 content that will be finalized this week.  Of course it's best if that version is known to function well.  😄

speckyspooky:
@merks 
No, the silence was due to my vacation an some actions with my family 😄
But now I was able to test the latest build of OSGi with my windows system, tomcat 9.x.
The demo reports can be created successfully, tested with standard reports and reports with charts.

**Technical details:**
- MS Windows 11
- Tomcat 9.0.93
- BIRT 4.17 - build N202408181237

From my side we can go ahead.

merks:
Sorry you’re we’re all entitled and have earned a personal life.  Thanks for the confirmation.

speckyspooky:
No problem, in the most cases I find a little bit time :o)

Issue #1867, opened 2024-08-21
Reports exported in DOC format does not show report content properly


hvbtup:
### Discussed in https://github.com/eclipse-birt/birt/discussions/1818

<div type='discussions-op-text'>

<sup>Originally posted by **PenghaiZhang** July 23, 2024</sup>
Hi everyone, I have report template created probably 10+ years ago. When my app integrated with BIRT 3.7.2, we could use this template to generate report in both different formats (e.g. HTML, PDF & DOC). Recently, we have upgrade BIRT to 4.16.0,  and then we found that the report in DOC format does not show contents properly. 

Basically, words in a `text` tag are shrunken to super narrow and become invisible. 

![image](https://github.com/user-attachments/assets/0fed117b-7340-4217-bb90-b3710b88c989)

![image](https://github.com/user-attachments/assets/de357b3a-18cc-40e1-809c-9b962c20aaa7)

You can see from the above screenshot where the top line and bottom line are broken.

Looking further to the table configuration, I saw the width is only 0.04cm.

![image](https://github.com/user-attachments/assets/7e414cc0-c013-4430-91dc-cbe57f6ae6d7)

We found that if we change the content type from HTML to auto or plain, the width will become normal again. However, we do want to support some HTML tags in our reports.

We also found that if we keep the content type as HTML but remove the text alignment, the content will be OK as well.

Can someone give me some ideas about this issue ?? </div>

hvbtup:
The discussion contains a radically simplified report (full source in XML is only a few lines) which shows that there is an issue with the DOCX emitter as well. The report is a 90cm wide master page, the content is only a 1x1 grid with a text item.
This is rendered as expected with the PDF emitter, but the DOCX emitter creates a file where the text is shown in an extremely narrow column, make it unreadable (nearly invisible).

speckyspooky:
The problem here is according of the MS Word definition that a colum have a width range between 0 and 55.87 cm.
Columns which are smaller or larger won't be displayed correctly.
(Familiar limitation is given with MS Excel and the column width.)

The problem of BIRT is that the maximum value should be set (if no width is that and the master page is larger like 55.87 cm)
and currently BIRT switch large values e.g. 60 cm to negative values which should be avoid.

speckyspooky:
I provided a PR #1869 to handle the maximum limit.

speckyspooky:
The PR #1871 is merged to the master branch for BIRT 4.17.

PR #1869, opened 2024-08-21
Fix the word column width which is larger like 22in (#1867)

speckyspooky:


hvbtup (approved):
Two suggestions:

When clipping the column widths, you use the Elvis operator `?:`. When clipping the table width, you use `Math.min`. I think using `Math.min` in both places looks a bit cleaner.

My gut feeling is that the `for` loop with the comment "validate the maximum of column width", which clips the calculated column widths, is not needed, because the column widths and the table width have already been clipped before.
But I'm just looking at the code, I did not test it.
Maybe you could explain why the `for` loop is necessary in the comment.

speckyspooky:
The Elvis operator is my standard comparer for such situation for all data types therefore I used it.

The loop is necessary because if no column width is set the calculation will be done before the loop.
If the column width is set the first check will bei used but the return is earlier.
The reason for the 2 checks at the end is the 2 returns and the different calculation methods.

hvbtup:
OK.

Issue #1876, opened 2024-08-25
Charts, image selection: image selection doesn't show the path/file name


speckyspooky:
Based on the ticket #1841 I tested the color selection of the chart.
The special option "color selection image based" which is shown at the original ticket has the topic
that the file path will not displayed if a file is selected. The file name of embeded image won't be displayed to.
To improve the handling here the dialog will be improved:
- full file path will be displayed for URI selection
- the URI selection need the file chooser too (according to the embedded image selection)
- file name will be displayed for mebedded images

### Chart, color selection display
![grafik](https://github.com/user-attachments/assets/287a2ab8-852c-4aab-bfa9-0c321d708f38)

speckyspooky:
This small enhancment is a reminder may be for an upcoming changes.

PR #1877, opened 2024-08-25
Enhance the meta data display of image color selection (#1876)

speckyspooky:


chart/org.eclipse.birt.chart.ui.extension/src/org/eclipse/birt/chart/ui/swt/composites/ImageDialog.java
```diff
@@ -104,9 +106,13 @@ public class ImageDialog extends TrayDialog {
 	protected SelectType selectType;
 
 	/**
-	 * The constructor.
+	 * The constructor 1
```
merks, line 109: This looks like a type.
```suggestion
	 * The constructor.
```

chart/org.eclipse.birt.chart.ui.extension/src/org/eclipse/birt/chart/ui/swt/composites/ImageDialog.java
```diff
@@ -296,6 +309,40 @@ public void widgetSelected(SelectionEvent event) {
...
+				});
+				try {
+					String fullPath = fileChooser.open();
+					if (fullPath != null) {
+						String fileName = fileChooser.getFileName();
+						if (fileName != null) {
+							imageData = null;
+							imageURI = fullPath;
+							if (selectType.equals(SelectType.EMBEDDED_TYPE)) {
+								uriEditor.setText(fileName);
```
merks, line 332: The place where you know it's a file name, I highly recommend properly converting it to a URI:
```suggestion
								uriEditor.setText(Path.of(fileName).toUri().toString());
```

merks (changes requested):
@speckyspooky 

I really don't understand what this is trying to implement in detail.    The description mentions color selection but the changes all have to do with "imageURI" so I'm not sure how that's related.

In any case, all the string manipulation to convert between a file system path and a file URI look really very bad and very error prone.  It doesn't handle conversion between \ and / or conversion between ' ' and %20 and the conversions are sprinkled everywhere which I don't really understand why that has become necessary.

I think it would be best that where you know you have a file system path, you use Path..of("C:/folder/file.png").toUri().toString() to convert **properly** to a URI.   The proper to and from URI conversion is like this:
```
Path.of(Path.of(fileName).toUri());
```
This makes sure segment separators and special characters are properly converted/encoded and decoded back.

Don't sprinkle that decision everywhere.  

Does this mechanism only support file URIs and not https or whatever?

speckyspooky:
I will cancel this PR and start an additional research on my side.

merks:
Path/url conversion is so complex and so poorly implemented even in the older jdk methods. The platform has all kinds of utilities (hacks) to work around such problems.  Such a mess. But I think the nio stuff has provided modern reliable utilities.

PR #1885, opened 2024-09-01
Enhance the update site to include all all-in-one product features

merks:
- This allows an all-in-one product installation to be provisioned purely from the BIRT update site.
- With that possibility, Oomph can generate a product definition for use in the installer just like it does for the EPP products and the Platform's SDK product.

### DIFF
--- a/build/birt-packages/birt-report-all-in-one/BIRT.product
+++ b/build/birt-packages/birt-report-all-in-one/BIRT.product
@@ -37,7 +37,8 @@ Powered by the Eclipse Platform: https://eclipse.org/eclipse
       <programArgs>-perspective org.eclipse.birt.report.designer.ui.ReportPerspective
 -product org.eclipse.birt.branding.birt_all_in_one
       </programArgs>
-      <vmArgs>-Xms512m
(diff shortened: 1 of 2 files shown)


merks:
@speckyspooky  @wimjongman 

If the update site is complete, we (Oomph) can generate a product setup like this:

image

This can be made available in a product catalog so that it's available in the installer:

image

image

I tested with a local build, and that works well.  Also, this way products can properly update all dependencies from one release to the latest release so hen I will add https://download.eclipse.org/birt/updates/release/latest to the available update site preferences for that purpose.

I assume there are no concerns with this.

merks:
I've also added the update site to the product definition so that it's available here:

image

speckyspooky:
Sounds very good for me.

merks:
I'm updating the generator to use the real BIRT update sites.  In the installer's simple mode BIRT will appear like this:

image

There are currently two versions generated:

image

The 4.16 release repository does not contain the all-in-one product so we can't generate a release version until 4.17 is release.

wimjongman:
Nice. Thanks, Ed.

Issue #1895, opened 2024-09-05
DOCX: margin of element "Label" not implemented (but available for element "Text" and "Dynamic Text"


speckyspooky:
The current DOCX-emitter support margin handling for the elements "Text" and "Dynamic Text"
but for the element "Label" it is missing. The reason of it is that this margin handling wasn't implemented.

The margin option should be available for the Label-element to.

### Current situation of label-margin at DOCX:

![grafik](https://github.com/user-attachments/assets/182cfa62-df1c-4133-a454-9e91c5a05010)

### Target of the upcoming change for DOCX with new handling:

![grafik](https://github.com/user-attachments/assets/3bd9ffd7-763c-4938-a999-dbf6f4edcb7e)

PR #1896, opened 2024-09-05
Add margin handling to the element "Label" (#1895)

speckyspooky:


merks (approved):
Looks okay.  No tests will ever verify that this behaves correctly.

Issue #1903, opened 2024-09-11
New Preview Prototype tab must be called Preview


wimjongman:
The Preview tab has a weird name.

Issue #1902, opened 2024-09-11
Ask to save the editor when it is dirty and Preview is requested.


wimjongman:
The preview does not work on unsaved content.

Please add something like this:

![image](https://github.com/user-attachments/assets/96f19486-6493-4c7d-8b1e-63e4bc62c47f)

Issue #1901, opened 2024-09-11
The preview does not work anymore:


wimjongman:
The preview does not work anymore:

![image](https://github.com/user-attachments/assets/780a34ac-6e86-4fc9-8280-313c374fe27f)

_Originally posted by @wimjongman in https://github.com/eclipse-birt/birt/issues/1732#issuecomment-2344152020_

wimjongman:
![image](https://github.com/user-attachments/assets/1a68f101-f4b9-417b-a59e-7efe5886a1b0)

PR #1904, opened 2024-09-11
The preview does not work anymore: #1901

wimjongman:
After switching the default browser to Edge in issues #1736 and #1737 the preview pane was broken because Edge does not like plain filenames.

Fixed:

1. Convert the plain file name to a file URL

This also fixed:

* Ask to save the editor when it is dirty and Preview is requested. #1902
* 'New Preview Prototype' tab must be called Preview #1903

### DIFF
--- a/UI/org.eclipse.birt.report.designer.ui.preview/src/org/eclipse/birt/report/designer/ui/preview/editors/ReportPreviewFormPage.java
+++ b/UI/org.eclipse.birt.report.designer.ui.preview/src/org/eclipse/birt/report/designer/ui/preview/editors/ReportPreviewFormPage.java
@@ -22,11 +22,14 @@ import org.eclipse.birt.report.designer.ui.preview.extension.IViewer;
 import org.eclipse.birt.report.designer.ui.preview.extension.ViewerExtensionManager;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.gef.ui.actions.ActionRegistry;
+import org.eclipse.jface.dialogs.MessageDialog;
+import org.eclipse.swt.SWT;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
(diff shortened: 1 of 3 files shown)


merks (approved):
Looks good.  Indeed best to use a proper URL!

UI/org.eclipse.birt.report.designer.ui.preview/src/org/eclipse/birt/report/designer/ui/preview/editors/ReportPreviewFormPage.java
speckyspooky, line 1: It is an option to use the message file for the new message (title, message, button).
The title sounds so dirty -> "The editor is dirty..", suggestion "The editor is unsaved"

speckyspooky:
@wimjongman 
I checked the changes an I have only a small question to the new message dialog - see comment.
Do you know which developer implements the "preview, prototype"?
I like the idea but the element support is not given over all elements and it would be good to know if there are changes upcoming.

speckyspooky:
Of course, with the preview we have a half solution of it in compare to the web-preview (e.g. parameter display, images, etc.).
And at the end yes, ask the original developer if it is possible to add there some details that it can be used in a good way.

I work with lot of reports with dynamic parameters and images. But the "preview" is not helpful with the unsupported parts.

wimjongman:
It processes parameters but no Image. It is weird because it should show the rendered html page. Please file an issue for it.

image

speckyspooky (approved):

merks:
@wimjongman 

FYI, you've been pushing branches to the main repository instead of to your clone again:

image

And not deleting it when done.

wimjongman:
Consider the environment before creating a fork! ;) 

I merged through the Github app and that did not delete the branch. It won't happen again. 😞

merks:
It's not a big deal.   Just a small deal.  👅

Issue #1905, opened 2024-09-12
DOCX: MS Word with uncorrect style converting of font size "12pt" based on BIRT produced DOCX-documents


speckyspooky:
Based on the discussion #1868 I started a research of the font-size styling topic to test different situations
what the font-size result will be given in compare of HTML -> (PDF ->) DOCX.

The styling from HTML to PDF is ok but a special situation with the font-size "12pt" cause a wrong style conversion in DOCX-documents.

The problem is that the HTML-content of the BIRT-elements "text", "dynamic text", "data" will not parsed into full supported DOCX-elements. Instead of a full parser the archive-/intermediate-format MHT(ML) will be used. These files embedded like external documents into the DOCX and will be converted through Word into the DOCX-elements.

**example of the internal DOCX-structure with MHT-files created by BIRT**
![grafik](https://github.com/user-attachments/assets/8a7820fc-5330-46a4-94ce-a2f3d5f15690)

And there exists a kind of standard error on MS Word side because the font size "12pt" will be converted in every case into "10pt".
This topic is given at the HTML-tags "div" & "span" but also the HTML4.0-element font is not converted correctly for result size "12pt".
(BIRT produce the DOCX-Version: WordProcessingML, Version 2016)

I tried to replace the "div"-tag with the "p"-tag (paragraph). But the converting of Word is not the same for these both elements
so we would get side effects of styling. In special cases the styling of a p-tag will override the style of following elements.

Normally the best way would be to implements a DOCX-parser to avoid MHT-files but for that large efforts are needed to get it.
(Note: DOC-files from BIRT 4.6 was created correctly, for that file format a kind of HTML-parser is implemented but cannot used for DOCX.)

The smallest workaround is to use another font-size instead of "12pt" to get a better result.
So I will provide a PR that the font-size will be changed from "12pt" to "12.5pt".
I know this isn't the same font-size and therefore I call it workaround. But it is better like the currenty situation.

Note: The behavior of #1868 was not reproducable on my side.

### Example 01 of wrong converting
![grafik](https://github.com/user-attachments/assets/6408d36d-7542-4998-97a4-dfd3749bc6e7)

### Example 02 of wrong converting
![grafik](https://github.com/user-attachments/assets/8a75390a-2374-4065-a06a-54d10f6a646c)

speckyspooky:
The test results of the new solution are tested.
The following screens provide the new result for the font-size for DOCX-documents.

### Example 01
[...]

PR #1906, opened 2024-09-12
Fix DOCX font size converting issue (#1905)

speckyspooky:


merks (commented):
There are many magic numbers.  Moreover the range of all size values is not covered, except I suppose to default to 10pt.   What if size is 8, -3, +5, or 100?  Is 10pt fine of all those?  Can the sizes specified in the DOM have units, like pt or px or are they purely numbers?  What's special about size.length > 2 such that you know to ignore all but the first two values?  I guess you assume then to consider only with +/- prefix and only up to the value 9.

It's not possible for fontSize to be null or have length not > 0 as tested later one.

speckyspooky:
The W3C documentation describe the allowed values: [https://developer.mozilla.org/en-US/docs/Web/HTML/Element/font](url)

grafik

Therefore values different from this range are not allowed and the default "10pt" is based on the BIRT default which is "10pt".
(On a next step I will provide user property class and then I can make the default configurable.)

Further answers to your questions:
- "size"-values couldn't have units (see documentation).
- Relative values: The prefix "+" & "-" is specific for relative values
- Relative values: only allowed for "-[0|..|3]" or "+[0|..|4]" (the +4 is a special case which is supported for the browsers but not listed at the documentation)
- Relative values: values of +2.5 will be handle through the browsers like "+2", there is no rounding up or down given

- The variable "fontSize" will be set into the second if-statement to the default "10pt" so in my point of view it cannot be null. Do you think I there is a constellation where it could be remain with the null-value?

merks (approved):
The explain makes it all clear now.  Thanks!  👍

merks:
@speckyspooky 

Will this be the final change for 4.17 such that I can do another milestone build after this is merged?

speckyspooky:
Yes, this is the last change of 4.17 from my side :o)

PR #1913, opened 2024-09-16
Tagged pdf

merks:


merks:
@hvbargen 

Apparently the changes cause test failures.

speckyspooky:
@hvbargen 
Hi Henning, is there a plan to publish these changes of tagged PDF to our main BIRT branch?
I have seen your amazing changes at this topic and it would be a very interesting feature for BIRT.

hvbtup:
The changes are not ready for the main branch.
It is in a POC state, showing that it is in fact possible to create a valid PDF/UA file with BIRT.

Apart from the fact that the code is a mess and that is is really, really incomplete, there are many places where an if statement should be used to do PDF-tagging work only when it is really needed.
And I'm only working on this topic as a hobby project, without getting paid.
So I want to concentrate on the topic, without being distracted by build issues, other Git commits, not breaking existing functionality.

**I really need co-workers on this topic.**

That's why I wanted to work on this on a new branch in *this* repository.

The plan is to first concentrate on making this work until we can say that it is at least good enough to work for others as a POC, too, including documentation of what is expected to work and what is not, examples and documentation.
Only then should we plan to polish it up in order to merge it into the main branch.

For those interested, at the moment, besides necessary code cleanup etc, the current state is like this:

**I am able to create valid PDF/UA including simple tables for very simple cases.**

* Lables and data items and simple tables are working.
* Grids are not yet tested.
* I (personally) am not interested in adding support for charts.
* The ROM seems to be prepared for complex tables (that what the `headers` and `scope` attributes are meant for), but I did not yet manage to make this work. Interestingly, these two attributes are handles differently from other attributes. Maybe that's why it does not work.
* Colors do not work: The PAC 2024 validator complains about the RG operator. So ATM I can only create black and white. I'm in contact with their support and already got an answer, but it's not clear to me what they mean. The answer was (translated):
 'We are in a Path Object State, and as such the operator "RG" is not allowed. There has to be a Path Painting Operator first.'
* HTML dynamic text "works", at least it doesn't fail, but except unordered lists, ATM I ignore all HTML tags. It is not clear anyway how this *should* be handled at all, eg. if the HTML contains a `H2` tag.

I think all of this is solvable.

**There's one topic which I don't know yet how to handle it, though: Page-Breaking.** 
In a Tagged PDF, if any structure element spans multiple pages, it still must be one structure element.
For example, if the report contains one table and it doesn't fit on one page, it must be *one* `Table` structure element. But Currently, the emitter creates one `Table` per page. 
Ideas are welcome.

hvbargen:
@speckyspooky 
Thomas, there's one thing that you could do which would be helpful:
There are several properties for report items which are related to accessiblity.
It would be great if these properties could be edited in a dedicated property tab (like as border, page-break etc) and could be grouped in the advanced properties tab.

I don't know if this list is complete, but at least these properties are directly related to accessibility:

- alt text
- scope
- headers
- PDF tag

And later we may need additional properties (I'm not yet sure) which would be useful for accessibility:

- "actual text" (this is meant if a text is rendered as an image, e.g if we use SVG to create rotated text). See the PDF spec for more info.

This list only contains one item as of now.

I thought about adding a boolean property "artifact" to this list, but I think declaring something as an artifact can also be achieved by writing "Artifact" as the PDF tag, so I think this is not necessary as an additional property.

I'm undecided about adding a property for additional PDF tag attributes, so to keep things simple for now, I did not add it to the list.

hvbargen:
Still working on complex tables here...
I found an issue which I don't know how to solve:

For complex tables, I need to define a bookmark for the TH cells and a comma-separated list of bookmarks names in the headers attribute of TD cells. This defines the relationship between TD cells and corresponding TD cells.

When I create a table with a group (say, Products grouped by product line), I want to use the expression row["PRODUCTLINE"] as the bookmark expression. By default, this is the same expression that is used for the TocEntry.

But now, the code which was added 14 years ago with 4e393b9ce22cb076cde3c316e7174eaeeef7b6b1 stands in the way: In ReportItemExecutorJava, the following lines were added:

if ( bookmarkManager.exist( bookmark ) )
{
  bookmark = bookmarkManager.createBookmark( bookmark );
}

createBookmark generated a new unique bookmark name.

This would probably be OK for TOC only, but if we want actually want to *use* the bookmark as a in-report link destination, this cannot work, because we don't know the generated bookmark name where we want to point at it.

This may even be a bug in BIRT if we just want to use actually bookmarks and TOC entries of the same name in general, independent of my attempt to support complex tables.

Any ideas except that the report developer has to take care about using distinct bookmark and TOC entry names?

Does anyone know about the commit or the issue which it tried to fix?

speckyspooky:
@hvbtup 
Hello Henning, I can take a look into the additiona feature which you mentioned.
I would start with the advanced register and then with the own property tab.

Could you assign the technical keys to your listed properties:
- alt text
- scope
- headers
- PDF tag

Issue #1921, opened 2024-09-25
BIRT Version >= 4.8 DOCX Export: NullPointerException Caused by Overflow Hidden Enhancement


kamlesh302:
Hi Team,

I am exporting a DOCX file using BIRT version **4.8**. The DOCX generation utilizes the **AbstractWordXmlWriter.java** class. During this process, an issue arises due to the enhancement of the overflow hidden property on a table cell in DOCX (**see enhancement #369**). This leads to a **NullPointerException** because the method attempts to get the length of a text field, which may potentially be **NULL**.

When exporting a DOCX file using BIRT version **4.6**, it works fine because that enhancement is not present. In that version, it directly calls the **AbstractWordXmlWriter.java > writeString(txt, style)** method, which handles null values properly.

The enhancement code in **cropOverflowString()** was added before the **writeString(txt, style)** method (which is also called from **writeTextInRun()**). This enhancement methods does not include a **NULL** check for the text field.

**Explanation of How NULL Can Occur:** 
In the **AbstractWordXmlWriter.java** class, the **writeText()** method handles a txt field that can be NULL. If txt is NULL, it calls the **writeContent()** method in **DocWriter.java**, passing the **NULL** value as a parameter. This, in turn, calls the **writeTextInRun()** method in A**bstractWordXmlWriter.java**, which also takes the txt parameter.

Inside this method, if the overflow hidden property is set, it calls **cropOverflowString()**, passing txt as the text parameter. In the **cropOverflowString()** method, there is a line that initializes a **StringBuilder**:
**`StringBuilder sb = new StringBuilder(text.length() + 1);`**
Here, **text.length()** is called when txt is **NULL**, resulting in a **NullPointerException**.

**The latest version of BIRT also contains the same code.**

**Please find the stacktrace below:**
`Cause:
java.lang.NullPointerException
	at org.eclipse.birt.report.engine.emitter.wpml.writer.AbstractWordXmlWriter.cropOverflowString(AbstractWordXmlWriter.java:1204)
	at org.eclipse.birt.report.engine.emitter.wpml.writer.AbstractWordXmlWriter.writeTextInRun(AbstractWordXmlWriter.java:1166)
	at org.eclipse.birt.report.engine.emitter.wpml.writer.AbstractWordXmlWriter.writeTextInParagraph(AbstractWordXmlWriter.java:1077)
[...]

PR #1923, opened 2024-09-28
Fix NPE if string for crop overflow is null (#1921)

speckyspooky:


speckyspooky:
This BugFix is related to #1921. If an example report is available the DOCX-emitter can be tested if there are further issues with the NPE or this fix is enough.

Issue #1941, opened 2024-10-20
DOCX: padding of element "Label" not implemented


speckyspooky:
The current implementation of the DOCX emitter has no implementation of padding for the BIRT element "Label".
The padding of the label should be added according to margin to set the according indents.
( This enhancement is familiar with #1895)

The difference from "label" to dynamic text will be that no special grid will be used, only the indent will be used.
MS Word doesn't support a real padding.

Additional on it some methods of the existing margin-padding-calculation are doubled with multiple implementations.
I will provide a PR to add the padding-calculation and optimize the methods.

hvbtup:
I'm not sure what you want to achieve.
If I understand correctly, the old version supported margin, but didn not support padding, and with your change, it supports padding, but not fully. It treats padding the same as margin, but the HTML box has distinct values for border, margin and padding for a reason. For example, to where background color extends.

Maybe it would be better to just document that the DOCX emitter doesn't support padding for labels (and plain text dynamic text items?).

hvbtup:
I understand and that it will be useful in common cases.

However, it will not fully support the box model (e.g. background color would not work as intentend, border will not work at all, I suppose) and it will change the behavior of existing reports.

I was going to write: "Also, please keep in mind that items may use `display: inline` instead of the default `display: block`". But AFAIK this feature doesn't work anyway with the DOCX emitter.

speckyspooky:
The enhancement is added with PR #1942

PR #1942, opened 2024-10-20
Enhance the docx emitter to add padding calculation to BIRT element "label" (#1941)

speckyspooky:


speckyspooky:
Hi @merks,

I have seen that my PR was crashing very early with the message
that a dependency wasn't available at repository side: org.mortbay.jasper.apache-el/9.0.90

Original message:
Failed to resolve target definition file:/home/runner/work/birt/birt/build/org.eclipse.birt.target/org.eclipse.birt.target.target: Could not find "org.mortbay.jasper.apache-el/9.0.90" in the repositories of the current location

Have you an idea how we can solve the missing dependency?

Thanks in advance
Thomas

merks (approved):
Looks okay, but I'm not good to judge what's correct just looking at the code...

Issue #1944, opened 2024-10-21
Fix the runtime example scripts


wimjongman:
The runtime example scripts have some out of the box quality issues.

wimjongman:
This has been fixed and verified.

PR #1945, opened 2024-10-21
 Fix the runtime example scripts #1944

wimjongman:
Fixed the runtime scripts for a better OOTB experience.


speckyspooky (approved):
The script changes and readme looks good.

PR #1953, opened 2024-10-24
Fix PDF syntax error (cm not allowed after BT)

### LINKED ISSUE #1951: PDF syntax error
BIRT (as well as obviously a lot of other programs) creates PDF files that are not strictly valid PDF.

The VeraPDF validator shows this as a warning:

**Special graphics state operator (cm) inside Text object**

The [PDF 1.0 specification](says: https://opensource.adobe.com/dc-acrobat-sdk-docs/pdfstandards/pdfreference1.0.pdf) says on page 93:

> None of the graphics state operators described in this section can be used within a
> path object. All except those that save and restore the graphics state (q and Q) or set
> the CTM (cm) can be included within text objects.

In other words: The `q`, `Q` and `cm` operators must not be included within text objects.

This is what VeraPDF complains about, but (IIRC) this error is so common that until now, it is only considered a warning.

I created a solution for this in the (now deleted) tagged-pdf branch.
This is available at my fork https://github.com/triestram-partner/birt/tree/tagged-pdf, see 373925c6cae69b95a27b3698d4119465ce7f6619

hvbtup:
This fixes #1951.

By the way: The author hvbargen is me, too.
I'm using that account for hobby projects, whereas I use hvbtup at work.


speckyspooky (approved):
Looks good and make sense!

Issue #1969, opened 2024-11-12
DOCX: Add the emitter configuration for the new layout option to the advanced register (#1957)


speckyspooky:
With the change of the enhanced layout options of the WordEmitter for docx new user properties are added.
The user properties should be directly accessible at the "advanced property" register according to the existing PDF- & Excel-emitter configuration.

Additional on it the header-footer configuration must be checked because the layout switch not correctly.

I will provide a PR for both parts.

speckyspooky:
The changes is merged to the master branch with PR #1969

PR #1970, opened 2024-11-12
The designer advanced register provide the configuration for the word layout (#1969)

speckyspooky:


engine/org.eclipse.birt.report.engine.emitter.docx/src/org/eclipse/birt/report/engine/emitter/docx/writer/Header.java
```diff
@@ -6,7 +6,7 @@
  * https://www.eclipse.org/legal/epl-2.0/.
  * 
  * SPDX-License-Identifier: EPL-2.0
- * 
+ *
```
wimjongman, line 9: Please add your contribution.

engine/org.eclipse.birt.report.engine.emitter.wpml/src/org/eclipse/birt/report/engine/emitter/wpml/IWordWriter.java
```diff
@@ -377,4 +377,19 @@ void writeContent(int type, String txt, IStyle style, IStyle inlineStyle, String
 	 * End page
 	 */
 	void endPage();
+
```
wimjongman, line 380: Is this breaking the API?

wimjongman (commented):
Hi Thomas, this looks good to me, but what about the API breakage? Do we not bring implementers into problems with these changes, or are all these classes internal?

hvbtup:
> Is this breaking the API?

Probably yes. But should we really care about that?

My personal opinion here is: So what?
Let me explain a bit more: I think the idea of commercial plugins for BIRT is dead anyway. The implementers you are talking about should really be contributors to the project. If they develop plugins as closed source, their companies can make a little money as a short-term effect, but in the long run neither their companies nor their customers nor the Eclipse project will benefit from it.

merks:
If we break API but only increment the minor version then the semantic version numbers become meaningless.  That's questionable.

engine/org.eclipse.birt.report.engine.emitter.docx/src/org/eclipse/birt/report/engine/emitter/docx/writer/Header.java
speckyspooky: I will add them

engine/org.eclipse.birt.report.engine.emitter.wpml/src/org/eclipse/birt/report/engine/emitter/wpml/IWordWriter.java
speckyspooky: The challenge is that the internal handling of the word emitter is based on the IWordWriter for doc and docx.
But both implementations have another structure. Therefore I had to add the 2 methods on the original interface.
But my point of view is that we can have new methods on interface side with new versions.

wimjongman:
I am not against breaking, but we must try to avoid it. And when we must do it then we have to increment the major version number of the plugin.

Also changing _**any**_ public method will break, not just interfaces. (It is only a break if the MANIFEST.MF exports the package containing the interface or the class.)

### Constructor and methods
For the constructors there are ways around. e.g. instead of changing the method you can do like this (overloading):

```java
// Old constructor calls the new constructor
Header(IPart part, Document document, int headerHeight, int headerWidth) throws IOException {
      this(part, document, headerHeight, headerWidthm, true or false);
	}

Header(IPart part, Document document, int headerHeight, int headerWidth, boolean wrapHeader) throws IOException {
		super(part);
		this.document = document;
		this.headerHeight = headerHeight;
		this.headerWidth = headerWidth;
		this.wrapHeader = wrapHeader;
	}

// in the same way the any old methods can call the new method.
```


### Interface
The problem is that the two methods must be implemented by all that use the interface. So it is breaking. Maybe you can provide a default implementation in the interface? This way others will still use the existing functionality.

```java
/**
 * Set the layout attribute for header and footer wrapping.
 * 
 * @param useWrappedTable use layout grid to wrap header and footer 
 */
default void setWrappedTableHeaderFooter(boolean useWrappedTable) {
    // Default implementation: does nothing or logs a message
    System.out.println("Default implementation: setWrappedTableHeaderFooter invoked with value: " + useWrappedTable);
}
```




https://github.com/eclipse-platform/eclipse.platform/blob/master/docs/Evolving-Java-based-APIs.md#Example-4---Adding-an-API-method

wimjongman:
> Probably yes. But should we really care about that?

Yes, we should care about it. Did you not take the oath? ;)

speckyspooky:
I have removed the header-constructor because at our project it changed all calls.
But you are right if some body use only parts of the projects it could be make trouble.
So I will re-add the header constructor.

I can also add the default method but I struggle with the MANIFEST,MF change of the plugin.
(I take a look into the docx-manifest but I'm not firm with the version numbers which must be changed.)

My idea is to implements an instance-check at AbstractEmitterImpl.java because the logic is only address to docx.
So I will try the object and can call the method only if it is an object based on class "DocxWritter".
In this case I can undo the interface change.

engine/org.eclipse.birt.report.engine.emitter.docx/src/org/eclipse/birt/report/engine/emitter/docx/writer/Header.java
speckyspooky: Done.

engine/org.eclipse.birt.report.engine.emitter.wpml/src/org/eclipse/birt/report/engine/emitter/wpml/IWordWriter.java
speckyspooky: Added default methods

speckyspooky:
@wimjongman, @merks 
I added the default methods to the interface. My idea wasn't working due to circle import-exports within the plugins.

Please let me know which version number I have to change at the MANIFEST.MF to fulfill your requirement of version handling for the plugin.

merks:
FYI, I'm out of office and will not have time until next week.

speckyspooky:
@wimjongman 
Please advise for the next step - addition changes necessary(?)

engine/org.eclipse.birt.report.engine.emitter.docx/src/org/eclipse/birt/report/engine/emitter/docx/writer/DocxWriter.java
```diff
@@ -66,12 +66,10 @@ public class DocxWriter implements IWordWriter {
 	 * @param compressionMode compression mode
 	 * @param wordVersion     word version
 	 */
-	public DocxWriter(OutputStream out, String tempFileDir, int compressionMode, int wordVersion,
-			boolean wrappedTableHeaderFooter) {
+	public DocxWriter(OutputStream out, String tempFileDir, int compressionMode, int wordVersion) {
```
wimjongman, line 69: This public method has changed. Can someone get in trouble with it? If so, can you please overload the method?

engine/org.eclipse.birt.report.engine.emitter.docx/src/org/eclipse/birt/report/engine/emitter/docx/writer/Footer.java
```diff
@@ -1,15 +1,16 @@
...
- * 
+ *
  * SPDX-License-Identifier: EPL-2.0
- * 
+ *
  *
  * Contributors:
- *  Actuate Corporation  - initial API and implementation
+ *  Actuate Corporation	- initial API and implementation
+ *  Thomas Gutmann		- add option to handle footer wrapper
```
wimjongman, line 13: Remember to add this to everything you change in the future.

engine/org.eclipse.birt.report.engine.emitter.docx/src/org/eclipse/birt/report/engine/emitter/docx/writer/Header.java
```diff
@@ -32,25 +33,39 @@ public class Header extends BasicComponent {
 	Document document;
 	int headerHeight;
 	int headerWidth;
+	boolean wrapHeader;
 
+	/**
+	 * Constructor 1
+	 */
 	Header(IPart part, Document document, int headerHeight, int headerWidth) throws IOException {
```
wimjongman, line 41: Thanks!

wimjongman (changes requested):
One last round.

engine/org.eclipse.birt.report.engine.emitter.docx/src/org/eclipse/birt/report/engine/emitter/docx/writer/DocxWriter.java
speckyspooky: No changes necessary because the latest version is the original version!
With the interface changes I was able to remove my change here.

engine/org.eclipse.birt.report.engine.emitter.docx/src/org/eclipse/birt/report/engine/emitter/docx/writer/Footer.java
speckyspooky: I will remember of it, till now it was not a style restriction to have it at all.
So yes, no problem :o)

speckyspooky:
I added my comments, thanks Wim!

engine/org.eclipse.birt.report.engine.emitter.docx/src/org/eclipse/birt/report/engine/emitter/docx/writer/Footer.java
merks: FYI. This is no longer strictly necessary because the information is in the commit history of git.

wimjongman (approved):
Thank you, Thomas! Great code.

wimjongman:
I was just looking for a good commit message because I did not understand the commit message:

"The designer advanced register provide the configuration for the word layout (#1969)"

Next time we have to make sure to have a good commit message.

speckyspooky:
@wimjongman I changed the commit title, hope it is more clear now.

wimjongman:
The PR title is not super important, because we also have the issue. It is about the commit comment.

https://github.com/eclipse-birt/birt/commits/master/

Just before we do a squash and merge. We have to make sure that there is a good commit comment.

PR #1972, opened 2024-11-14
Preliminary PDF/UA support

hvbtup:
As I said yesterday, I think my changes are now in a state that they do no harm to the master branch; creating normal PDFs should work exactly as before.
It is now possible to at least generate valid single-page PDF/UA files.
The smiling sun image is my IP, I created it a few years ago for a game experiment.

I cannot split this into more smaller commits in a reasonable way.

Note: The Java code certainly has some room for improvements. 
In several cases I used the `instanceof` operator, because it was the easiest solution and avoids having to change interfaces.

BTW most of the changes in EngineIRVisitor.java were suggested by the IDE to remove warnings.


hvbtup:
@wimjongman 

One of the tests is failing:
EngineIRIOTest.testIO:39->doTestIO:72 EngineIRIOTest, doTestIO(compare: golden, value) - designName: action_test.rptdesign expected:<...     <label id="8" t[ag-type="P" text="uri-expr"> ... (output shortened) ... but got:<...     <label id="8" t[ext="uri-expr"> ...

This obviously means that a tag-type attribute got lost. This might be related to one of my changes: AFAIK the tags are case-sensitive, for example it must be `P` and not `p`.
I changed this in model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/elements/rom.def

Probably now the value of `P` is not written to the file because `P` equals the default value.

So I think I have to modify the "golden" file, do I?

hvbtup:
Nah. By examining the test, I see that the "golden" image is not a file. It is the report, loaded with `ReportParser().parse`.
This is then serialized with `EngineIRWriter` and deserialized with `EngineIReader`. The first and the final report are both saved with `ReportDesignWriter`. And obviously `ReportDesignWriter` actually writes the tags into the rptdesign file, but when using the long way through the IR, they get lost.

speckyspooky:
@hvbtup
I have dine a first review to understand your changes and I think I understand your steps. But I will do a second read of it. The only thing in generally I saw some comments in German and file name "Sonne.png". It would be good if we have all comments in English and the image name is optional for me.

hvbtup:
I kept some German in the example reports and the file name, but I removed the German comments and cleaned up the code.

hvbtup:
@speckyspooky 
The build fails with errors that seem unrelated to my changes. I don't know why.

wimjongman:
> I hope this can be squashed into a single commit rebased commit rather than these 25 commits with merge commits before this is completed...

Yes, our strategy is to SQUASH.

merks:
BTW, this is one of the reasons why during most of the develop cycle I use the most recent available dependencies:

image

I.e., to  catch problems as soon as possible after they occur upstream.

Very soon the 2024-12 cycle will come to a close; rc1 is this week and rc2/final is next week.  At that point I will try to lock down the repos to permanent location so that release tags can be built in the future.  And hopefully BIRT can release shortly after the December 4th release...

wimjongman:
> BTW, this is one of the reasons why during most of the develop cycle I use the most recent available dependencies

Yes. I was impressed by the speed you caught this. 👍

merks:
@hvbtup

FYI, the PR is functional again with the GEF breakage fixed now.

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/README.md
```diff
@@ -33,6 +33,17 @@ The following list get an overview of all supported user properties, the content
...
 
+
+**PdfEmitter.UAConformance**
+
+	Content    	define the PDF/UA conformance of the created PDF document, e.g. PDF.UA-1
+	Location   	report
+	Data type  	string
+	Values     	PDF.Standard, PDF.UA-1
+	Default    	PDF.Standard
+	Since      	4.16
```
speckyspooky, line 44: The version numbers of "Since" & "Designer" is 4.18 and mark the availability of the configuration option.

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/layout/pdf/util/HTML2Content.java
```diff
@@ -631,6 +633,7 @@ static void handleElement(Element ele, Map<Element, StyleProperties> cssStyles,
 				if ("".equals(text.getText())) // add default list type when tag <ul> attribute is empty.
 				{
 					text.setText("\u2022"); // the disc type
+					text.setTagType("Lbl"); // TODO HVB check if this is correct
```
speckyspooky, line 636: Is here a verification ongoing or done?

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/layout/pdf/util/HTML2Content.java
```diff
@@ -639,6 +642,7 @@ static void handleElement(Element ele, Map<Element, StyleProperties> cssStyles,
 			childCell.setColumn(1);
 			childCell.setColSpan(1);
 			childCell.setInlineStyle(style);
+			childCell.setTagType("LBody");
```
speckyspooky, line 645: Suggestion: Make it sense to create constants on an interface for all specialized PDF/UA tag types, may be at your interface ITagType?

speckyspooky (changes requested):

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/README.md
hvbtup: > mark the availability of the configuration option.

I don't know what you mean.

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/layout/pdf/util/HTML2Content.java
hvbtup: Neither nor. I don't know if this is corect. I hope someone else who knows more about tagged PDF can tell us.

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/layout/pdf/util/HTML2Content.java
hvbtup: That's a good idea. But I don't find the time ATM.

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/README.md
speckyspooky: The version numbers have to be changed

- Since 4.18
- Designer 4.18

merks: If this doesn’t land for the December release it will change again.

model/org.eclipse.birt.report.model/src/org/eclipse/birt/report/model/elements/rom.def
```diff
@@ -1914,6 +1918,9 @@
       <Property detailType="emitterPdfConformance" canInherit="false" displayNameID="Element.ReportDesign.Emitter.pdf.conformance" name="pdfConformance" runtimeSettable="false" since="4.17" type="choice" >
         <Default>PDF.Standard</Default>
       </Property>
+      <Property detailType="emitterPdfUAConformance" canInherit="false" displayNameID="Element.ReportDesign.Emitter.pdf.ua.conformance" name="pdfUAConformance" runtimeSettable="false" since="4.17" type="choice" >
```
speckyspooky, line 1921: Change "since" to your plant BIRT version like "4.18" (or later)

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/layout/pdf/util/HTML2Content.java
hvbtup: I suggest to do this refactoring later.

hvbtup:
For the release notes, it should be clearly stated that PDF/UA support is a Work-In-Progress which just started.

In particular, it is not yet possible to generate accessible PDFs where items (eg tables) span more than one page, even though the PAC 2024 says the generated PDF is valid.

Furthermore, existing BIRT reports won't conform to PDF/UA just like that. It is always necessary to carefully set advanced properties in many places.

speckyspooky (approved):
Go from my side.
Yes, we should address the prototype character for that feature.
But it is a very good starting point and more like starting point!

wimjongman (approved):

wimjongman:
Thanks, Henning!

Issue #1979, opened 2024-11-16
DOCX, optimized version: measure for intent incorrect if measure "mm" was used for margin & padding


speckyspooky:
The optimized version of the word structure has a problem with the margin & padding calculation.
The problem is that the indent value is to high if "mm" was used.

After research I figured out that the convert-method calculate the measure wrongly.

I will create a fix for that.

speckyspooky:
After fixing of the mathematics behind the indent will be calculated correctly.
This is a scren of right-indent before and after of the word test document:

![grafik](https://github.com/user-attachments/assets/5daaf61a-06b3-47d3-b239-fad55497aad9)

PR #1980, opened 2024-11-16
Fix DOCX calculation of margin and padding (#1979)

speckyspooky:


hvbtup:
Calculations with several divisions should use brackets. Better even, aren't there constants defined somewhere for converting between different units, eg. `MM_TO_POINT` or something like that?

Issue #1981, opened 2024-11-17
List box, display text: enable the parameter task to handle the display text of multi selections for list box parameters


speckyspooky:
The current handling of parameters and there display text doesn't support the option to get the selected display text results of a list box which is marked as multi selection.
The display text implementation allows only to get the value of a single selection list and for multi selections is the return value "null".
This handling is offical documented at source code side.

This enhancement is created to implements the option of display text for both versions of single selection and multi selection for list boxes.
There are different tickets where the feature was addressed and on of the latest is #1967.

I will create a PR for the enhancement.

speckyspooky:
For the testing I used a test report with 3 list boxes:
- 2 multi selection lists
- 1 single selection list

The following options are tested to get the display text:

```
params["paramMultiSelectionList"].displayText
params["paramMultiSelectionList"].displayText.toString()
params["paramMultiSelectionList"].displayText.join("~")
reportContext.getParameterDisplayText("paramMultiSelectionList").join("~");
```
The following 2 screens display the different report creation options of Run-task & RunAndRender-task.
For a better understanding I add the "value"-option and the "display text"-option of the parameters at the test report.

### Run-task
![grafik](https://github.com/user-attachments/assets/8d149d33-07b7-4cef-8db9-823a0819ccee)

### RunAndRender-task
![grafik](https://github.com/user-attachments/assets/96eb0b2d-cb95-48db-bba9-ec062eeb2180)

PR #1982, opened 2024-11-17
Enhance display text for multi selections (#1981)

speckyspooky:
The implementation is done for the both processing task:
- RunAndRender-task
- Run-task

The handling of the parameters and the display text is different. Additional on it the assignment of the display text is enhanced so that the result of the display text can be a string-array which was already implemented but never used.
For the internal handling I used a ArrayList<String> with will be converted to a standard string-array.


wimjongman (approved):
Looks good! Thanks, Thomas.

Issue #1986, opened 2024-11-26
Security-Issue sonatype-2018-0590 (possible XXE (eXternal XML Entity)-attack when Using XML as Data Source)


mr0reed137:
Hello,

Our security-scan recently reported a security vulnerability in the rhino library, namely **sonatype-2018-0590** with the following explanation:

<img width="354" alt="2024-11-26 10_29_56-IQ Server - Component Details – Mozilla Firefox" src="https://github.com/user-attachments/assets/c047e81f-885f-4383-9918-0d525a6e0179">

This library is included in the Birt Engine (v4.14.0). 

I can confirm that the Birt Engine processes XMLs extended by XXE accordingly and that the vulnerability can therefore be exploited.

An example for a manipulated XML with XXE-attack:

<img width="156" alt="2024-11-26 10_41_02-New Issue · eclipse-birt_birt – Mozilla Firefox" src="https://github.com/user-attachments/assets/3dbd9e15-f81b-4024-85fd-40982f79d936">

The result of the exploited vulnerability in this case is that the contents of the /etc/passwd - file are printed in the pdf.

Although the rhino library has been updated in Birt version 4.17, the vulnerability can still be exploited. I suspect that the processing of XMLs in the Birt Engine as a datasource is not implemented with rhino.

Is this known and can the processing of XML with external XML entities be prevented?

merks:
I'm not sure the basis for asserting that the exploit you describe is in some way related to the version of rhino that's being used.  The issue that you show says 1.7.12 fixes the problem and given BIRT 4.17 uses 1.7.15  there is some kind of contradiction or confusion there.

As @hvbtup suggests, if the sources of your report inputs are not secure then you need to secure those inputs such that they do not contain content that can comprise your server or can produce compromised reports.

I expect there are places where BIRT processes XML where we could and probably should disable some of the vulnerable aspects in the existing XML libraries, e.g., entity resolution and external DTDs.  But who is going to do that work or fund that work?  I expect no one will do either.

hvbtup:
As @merks said, based on the information you gave us, the issue is already fixed in 4.17.

mr0reed137:
The issue is fixed in rhino, yes. 
[...]

PR #1987, opened 2024-11-27
Disable entity-related security features in CommonUtil.createSAXParser

merks:
https://github.com/eclipse-birt/birt/issues/1986


hvbtup (approved):

hvbtup:
grafik

@merks This prevents XXE in some other places (eg. when parsing *.rptdesign files), but probably it does not fix #1986.

merks:
@hvbtup 

Yes, I'm only plugging obvious holes based on what the Platform did not so long ago...

PR #1989, opened 2024-11-29
Update the setup and the target platform specify permanent sites

merks:
- The setup change is temporary until 4.19 development begins.
- Similarly the target platform will be regenerated to moving-target sites when 4.19 begins.
- Only the datatools site is still a moving target pending a 1.16.2 release.

https://github.com/eclipse-birt/birt/issues/1910


hvbtup:
I have no further changes to suggest, and if I understand correctly, your changes will (starting from 4.18) enable me to choose (e.g. for my own branches) to choose if I use the latest stable BIRT release as a base or use the newest BIRT.
Cool!

speckyspooky:
I have no further changes too.
So we can go ehaed with the new build version 4.18

PR #1994, opened 2024-12-04
BIRT OSGi, copy org.apache.commons.commons-logging to the lib-folder

speckyspooky:
The logging factory is switched from "org.apache.commons.logging" to "org.apache.commons.commons-logging" and must be copied from platform/plugins-folder to the WEB-INF/lib-folder.


merks:
@speckyspooky 

This looks promising then.  I will keep and eye on the build.  If this goes well, we can do another milestone after you've tested the nightly build kicked off when this merged.  Certainly I have no concerns about the change itself...

Issue #1993, opened 2024-12-04
How to apply dataset filter expression while get parameter selection list?


chille716:
Hello,  

i have a report design with a dataset and a filter expression like the following one:
![grafik](https://github.com/user-attachments/assets/82405fda-21d5-4f09-9f01-b44f599ee055)

This data set will be used as dynamic values source for a report parameter.  

My problem is now, that the filter expression will not apply, while fetching the values of the parameter with help of the java method: IGetParameterDefinitionTask#getSelectionList

This behavior was changed with version 4.8 and is already mentioned in https://github.com/eclipse-birt/birt/commit/e19c5f5c4f839ada0a5a0f2e16d22bfd00adc821#r39507859

Now my question: Is there any way to change the DataEngineFlowMode for this use case? Or is there another way to fetch the parameters, considering the filter expressions of a data set?

speckyspooky:
Currently there is no configuration to change the behavior.
Which kind of dataset do you use? If you would use a sql-statement you can do it directly there.
But based on the screen it seems that you doesn't use sql.

speckyspooky:
If we would add a configuration there it must be a kind of global parameter because it has an impact for all elements of a report.
And if it is the requirement so we it should work for all reports with the according configuration.
So we can talk about to add a global configuration for the JVM like `-Dbirt.driver.property.validation.excluded.drivers`.

But be aware it could be that you get in trouble with the topics of the discussion which cause the change.

speckyspooky:
I understand. So it if I have time, I can spend time for research of it for BIRT 4.19.
The target would be to have a set- & get-method to set the "DataEngineFlowMode" which shall be accessible through the interface class.

chille716:
This would be great. We only made the following changes locally... 
[diff.txt](https://github.com/user-attachments/files/18010917/diff.txt)
but as already mentioned, i'm not sure if this is a 100% solution.

wimjongman:
@chille716 can you make a pull request?

chille716:
@wimjongman yes, i will do it in the next few days.

wimjongman:
> @wimjongman yes, i will do it in the next few days.

Thanks @chille716. In case you did not do it. Please sign the Eclipse Contributor Agreement. Make sure to use your github email when you create an Eclipse Account.  

https://www.eclipse.org/legal/eca/

![image](https://github.com/user-attachments/assets/697c43d9-8d62-478a-8a41-d3ca34ce41ea)

[...]

PR #2005, opened 2024-12-09
How to apply dataset filter expression while get parameter selection …

chille716:
…list? #1993

I added a boolean to enable/disable the normal data engine flow mode to apply filter expressions in datasets.

### DIFF
--- a/engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/api/IGetParameterDefinitionTask.java
+++ b/engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/api/IGetParameterDefinitionTask.java
@@ -159,4 +159,20 @@ public interface IGetParameterDefinitionTask extends IEngineTask {
 	 */
 	Collection getSelectionTreeForCascadingGroup(String parameterGroupName);
 
+	/**
+	 * Enable or disable the normal flow mode. The normal flow mode can be used to
+	 * apply filters on data sets in data engine execution.
+	 *
+	 * @param normalFlowModeEnabled Set to true to enable the normal flow mode.
+	 */
+	void setNormalFlowModeEnabled(boolean normalFlowModeEnabled);
+
+	/**
+	 * Is the normal flow mode enabled or not? If enabled, filters can be applied on
+	 * data sets.
+	 *
+	 * @return Returns true if the normal flow mode is enabled. Otherwise false.
+	 */
+	boolean isNormalFlowModeEnabled();
+
 }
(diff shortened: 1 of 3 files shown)


merks (commented):
Wouldn't it be more general to have a getting and setting that returns a value of type DataEngineFlowMode?

chille716:
@merks I was not sure about it... that's why I implemented it microinversively. Can't find any places where the other modes (IV, DATA_EXTRACTION) will be used.  But if it useful to set these modes at this place, then i can change it.

merks:
FYI 

@speckyspooky  @wimjongman 

I don't really know the APIs well.  You're right that the other enumerators appear unused.  So I'll wait for other folks to comment.   Note that adding methods to API breaks the API so whatever we do here,  I think it would be better to define default methods in the interface, which means the mock class needs no changes....

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/api/IGetParameterDefinitionTask.java
```diff
@@ -159,4 +159,20 @@ public interface IGetParameterDefinitionTask extends IEngineTask {
 	 */
 	Collection getSelectionTreeForCascadingGroup(String parameterGroupName);
 
+	/**
+	 * Enable or disable the normal flow mode. The normal flow mode can be used to
+	 * apply filters on data sets in data engine execution.
+	 *
+	 * @param normalFlowModeEnabled Set to true to enable the normal flow mode.
+	 */
+	void setNormalFlowModeEnabled(boolean normalFlowModeEnabled);
```
speckyspooky, line 168: Add "default"-method for the new method - PR #1970, see IWordWriter.

grafik

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/api/IGetParameterDefinitionTask.java
```diff
@@ -159,4 +159,20 @@ public interface IGetParameterDefinitionTask extends IEngineTask {
...
+
+	/**
+	 * Is the normal flow mode enabled or not? If enabled, filters can be applied on
+	 * data sets.
+	 *
+	 * @return Returns true if the normal flow mode is enabled. Otherwise false.
+	 */
+	boolean isNormalFlowModeEnabled();
```
speckyspooky, line 176: Add "default"-method for the new method - PR #1970, see IWordWriter.

speckyspooky (changes requested):
Please add there default-methods to avoid broken API

speckyspooky (approved):
Looks good

Issue #2001, opened 2024-12-07
HTML preview: border style changes with scripting isn't displayed (but displayed on other outputs)


speckyspooky:
If a border is definded on an element like a grid or table and the style information like the border color or the border width will be changed then is the change not displayed at the preview. But the change will be displayed on all other outputs like PDF, DOCX and PPTX.

The topic here is that the original style information are not inherit to the new style which is based on the scripting.
I will create a change for it.

### Grid definition
![grafik](https://github.com/user-attachments/assets/43c87dd8-2414-42d5-bea7-56fededd3320)

### Script to change the color - result HTML & PDF
![grafik](https://github.com/user-attachments/assets/e4dde018-abc3-48d4-91ea-217b5deb357c)

speckyspooky:
The change is merged to the master with PR #2007

PR #2007, opened 2024-12-12
Enhance the HTML border handling for style scripting (#2001)

speckyspooky:


engine/org.eclipse.birt.report.engine.emitter.html/src/org/eclipse/birt/report/engine/emitter/html/HTMLEmitter.java
```diff
@@ -316,11 +318,74 @@ protected IStyle getElementStyle(IContent content) {
...
+	/**
+	 * The border style needs inherit properties based on the designer configuration
+	 * The style information will be transferred to the style if it set on designer
+	 * but unset at style object
+	 *
+	 */
+	private IStyle inheritBorderStyle(IStyle style, IStyle computedStyle) {
+
+		// properties: border top
+		if (computedStyle.getBorderTopStyle() != null && !computedStyle.getBorderTopStyle().equals(BORDER_STYLE_NONE)) {
```
merks, line 334: These days Object.equals is a nice way to detail with all the extra null guards.  Of course in this case you can compare to BORDER_STYLE_NONE which is not null.

merks (approved):
Looks generally sensible and fine.

Issue #2008, opened 2024-12-14
PDF/UA tagging, remove eclipse warnings and small clean up


speckyspooky:
I have started a small clean up of some eclipse warning of the new tagging feature for PDF/UA.
The core change is to remove some warnings and put tag-types and pdf-names into static properties.

speckyspooky:
The change won't be merged to the master because further changes are upcoming to this topic.

PR #2009, opened 2024-12-14
Remove eclipse warnings and add tag properties (#2008)

speckyspooky:


hvbtup:
It's the right idea at the wrong time....

I did some cleanup work in my hobby account recently, which conflicts heavily with your changes here.
See https://github.com/hvbargen/birt/tree/pdf-tag-page-break

I suggest to cancel this PR and we can cleanup the code once I come up with the next PR.

In particular, I think about renaming the pushTag and popTag methods, because the structure is a tree, not a stack.
So probably they should be called openTag and closeTag or beginTag and endTag (not sure, the idea is to use similar names as one would use for creating an XML structure).

speckyspooky:
I'm fine with it then I will cancel the request to avoid problems on your side.

PR #2020, opened 2025-01-06
PDF emitter VerticalTab user property

hvbtup:
This adds two user properties which can be used in Cell or in Text/Label items to achieve a functionality like a vertical tab stop.
Example reports for both properties are provided, which two different values "10cm" and "20cm" which show the effect.

For "10cm", the current y position already is more than 10cm from the top, so there is no effect. For "20cm", the current y position is less than 20cm from the top, so the effect is to go down until 20cm from the top.

"top" means the start of the body, eg. when the master page has 1cm margin-top and 1cm header height, it is 22cm from the top of the sheet.

The property is useful in some special use cases, which are not frequent enought to justify the introduction of an additional property to the ROM. That's why I think that user properties are sufficient.

### DIFF
--- a/engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/nLayout/area/impl/BlockContainerArea.java
+++ b/engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/nLayout/area/impl/BlockContainerArea.java
@@ -78,12 +78,9 @@ public class BlockContainerArea extends ContainerArea implements IContainerArea 
 				// Variant 1: The property has to be set for a Label, a Dynamic Text Item or
 				// similar (not for a table row or cell).
 				// A possible structure:
-				// Table
-				// Header
-				// Details
-				// Footer Row with Dynamic Text Item with FixYPosition=20cm.
-				// Footer Row with data which should be displayed starting at y=20cm.
-				String fixYPosition = (String) containerContent.getUserProperties().get("FixYPosition");
+				// ...
+				// A dynamic Text Item with FixYPosition=20cm.
+				String fixYPosition = (String) containerContent.getUserProperties().get(PDF_VERTICAL_TAB);
 				if (fixYPosition != null) {
 					int limit = getDimensionValue(
 							content.getCSSEngine().parsePropertyValue(StyleConstants.STYLE_MARGIN_TOP, fixYPosition));
--- a/engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/nLayout/area/impl/BlockContainerArea.java
+++ b/engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/nLayout/area/impl/BlockContainerArea.java
@@ -32,6 +32,8 @@ import org.eclipse.birt.report.engine.nLayout.area.style.BoxStyle;
 import org.eclipse.birt.report.engine.util.BidiAlignmentResolver;
 import org.w3c.dom.css.CSSValue;
 
+
+
 /**
  * Implementation of block container area
  *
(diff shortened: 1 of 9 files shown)


engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/nLayout/area/impl/BlockContainerArea.java
```diff
@@ -69,6 +71,32 @@ public BlockContainerArea() {
...
+			if (containerContent != null && containerContent.getUserProperties() != null) {
+				// Variant 1: The property has to be set for a Label, a Dynamic Text Item or
+				// similar (not for a table row or cell).
+				// A possible structure:
+				// Table
+				// Header
+				// Details
+				// Footer Row with Dynamic Text Item with FixYPosition=20cm.
+				// Footer Row with data which should be displayed starting at y=20cm.
+				String fixYPosition = (String) containerContent.getUserProperties().get("FixYPosition");
```
speckyspooky, line 86: The user property should have more a familiar name with the content of the functionality and the prefix-naming should be equal to the 2. optin of the vertical tab e.g.:
- VerticalTabTextFixYPositon
- VerticalTabCellFixYPositon

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/nLayout/area/impl/RowArea.java
```diff
@@ -230,6 +231,28 @@ public void add(AbstractArea area) {
...
+		// Variant 2 . The property must be set for a Cell.
+		// A possible structure:
+		// Table
+		// Header
+		// Details
+		// Footer-Row with Cell with VerticalTab=20cm.
+		// Footer-Row with information that should be shown starting with y=20cm.
+		IContent cellContent = cArea.getContent();
+		if (cellContent != null && cellContent.getUserProperties() != null) {
+			String verticalTab = (String) cellContent.getUserProperties().get("VerticalTab");
```
speckyspooky, line 243: See comment above to optimize the user property name

speckyspooky:
Have you tested the effect for the other emitters DOCX, PPTX and XLSX for example?
Have we there impacts or no effects given on these output results?

hvbtup:
The change only affects the PDF emitter. I tested DOCX, XLSX and PPTX and HTML.

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/nLayout/area/impl/BlockContainerArea.java
hvbtup: Too long to type without typos. But I wonder now (I introduced this years ago in my branch), why the user properties should have different names at all. Why not just `VerticalTab` in both cases?

speckyspooky:
If it is only for PDF then the user properties should have an according user Property name with "PdfEmitter.VerticalTab..."
and should be documented on the REAME.MD too.

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/nLayout/area/impl/BlockContainerArea.java
speckyspooky: I'm not with you, if you take a look into the README.MD file you will see also complex namings.
And especially if you use it only for PDF it shall be added with "PdfEmitter.VerticalTab...".

Therefore we have the documentation and the next point is the relation between both propeties which can be documentet at the README.MD.

speckyspooky (approved):
Looks good with the new naming.

PR #2021, opened 2025-01-06
Add support for ExcelEmitter.AutoColWidthsIncludeTableHeader and Footer

hvbtup:
... user properties.
The boolean properties `ExcelEmitter.AutoColWidthsIncludeHeader` and `ExcelEmitter.AutoColWidthsIncludeFooter`, when set to `true` for a table item (default is `false`), cause the inclusion of the table header rows and (one) footer row in the row range for the automatic column width calculation. This is useful if the header content can be wider than the maximum detail content.

Like before, a maximum of 12 rows are considered.

A report demonstrating the feature is included as an example.

The original code was from Jim Talbut alias "yaytay" and part of the "tup_main" branch in my fork.


speckyspooky:
The README.MD means with property "Designer" that the property can be configured via designer property.

Please use the Attribute "Since" for 4.19 or add the class changes for the eclipse designer handling to it. You can use the pattern from my changes for the designer.

speckyspooky (approved):
Looks good

Issue #2026, opened 2025-01-09
Two DOCX emitter page header issues


hvbtup:
I noticed some issues when I compared the DOCX output of my old customized DOCX emitter with the output of the DOCX emitter from master:

If `showHeaderOnFirst` is not set to `true` for the master page, then the DOCX emitter does not show the page header at all.

If I set this property to true for testing purpose, then the page header is created, but a second issue occurs:

In my test report, the content of the page header is a single dynamic text item with a border-bottom line (not a grid or whatever).
But the DOCX emitter does not create this line.

![grafik](https://github.com/user-attachments/assets/f3b535e1-7c73-408e-a25f-baf45edcf434)

![grafik](https://github.com/user-attachments/assets/9fc3853b-9686-4798-9890-13a9e7de8a30)

My customized emitter supported `showHeaderOnFirst`, see https://github.com/triestram-partner/birt/commit/403e3af83cf74d7d9ccc5362a1bca6f5bb6b4616

Fixing this is not  a top priority ATM. Maybe I can try to carefully create a PR from this once I find the time.

hvbtup:
I don't have any idea what I could do regarding the missing line issue, however...

speckyspooky:
The changed behavior was added from my DOCX changes with BIRT 4.18 and the try to remove the additional formatter-grids.

This cause the effects and there are 2 effects:
1.) The bottom-line is unprinted because the layout-grid was avoid because it exists not really such style-element for docx
2.) The header-footer-height is unset because an empty grid was used to produce the correct height. This is not the right way but the solution to avoid the grid is only the first step. The docx-property of header-height/footer height must be set (currently not implemented)

Solution to get the old behavior since 4.18 & 4.19 (current version/nightly build):
- set 2 properties on report level on the "advanced property register"

![grafik](https://github.com/user-attachments/assets/9720ec5e-17d1-4bd1-b0e6-545a5c318db9)

With your ticket it would be better to switch the configuration from disabled (grid-usage=false) to enabled (grud-usage=true).
Please let me know.

hvbtup:
For the missing line, I did not follow your suggestion, but your suggestion gave me the idea to use a 1x1 grid, put the dynamic text item into that cell and set the border bottom property at the cell level.

I think this is a reasonable workaround: My test report is special int that it uses just a dynamic text item in the page header.
In production, the reports usually contain a grid in the page header and footer anyway.

[...]

PR #2027, opened 2025-01-10
DOCX, optimize header & footer handling (#2026)

speckyspooky:
The optimization is for DOCX and the new option of header/footer without layout grid.
For this option the positions of the header and footer will be re-calculated.
The 2 methods `writeSectionInBody()` and `writeSectionInPara()` will use the same logic to handle the header and footer calculation.

The original behavor can be reconfigured very simple with the property register "Advanced" at the property editor on report-level. If the word-option "Use layot grid of header & footer" then the old behavior will be used.

### DIFF
--- a/engine/org.eclipse.birt.report.engine.emitter.wpml/src/org/eclipse/birt/report/engine/emitter/wpml/AbstractEmitterImpl.java
+++ b/engine/org.eclipse.birt.report.engine.emitter.wpml/src/org/eclipse/birt/report/engine/emitter/wpml/AbstractEmitterImpl.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2008, 2024 Actuate Corporation and others
+ * Copyright (c) 2008, 2024, 2025 Actuate Corporation and others
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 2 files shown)


hvbtup (commented):
The code change seems reasonable.
But when I tried to run a simple report as DOCX, this caused a crash in BIRT.

[birt-2027-crash.zip](https://github.com/user-attachments/files/18394348/birt-2027-crash.zip)

I don't know if this is actually related to the code changes.
If I change the DS setting to fetch max. 3 rows (so that it fits on one page), there is no crash.

speckyspooky:
The documenten report error (NPE) with docx will be fixed with PR #2029. The build got now green light.

hvbtup:
Can you add an example DOCX output before and after the change?

speckyspooky:
Here are the compare of the version BIRT 4.18 and BIRT 4.19 with optimization:

grafik

The original-layout-grid of docx which was used till BIRT 4.17 produced the header/footer height.
BIRT 4.18 avoid the additional grid so it can be done manually or switch on the oly behavior via "adavnced register".
Nevertheless it is better to have the herader/footer height by default therefore the optimization of BIRT 4.19.

hvbtup (approved):
Seeing is believing...

Issue #2028, opened 2025-01-13
DOCX, font-tag can cause NPE and the font familiy & font color will not fetched correctly


speckyspooky:
The HTML-tag "font" can cause an NPE - demo report: [birt-2027-crash.zip](https://github.com/user-attachments/files/18401645/birt-2027-crash.zip) .
During the testing I figured out that the font-family and font-color will ignored at some cases.

I will provide a PR to fix both situations.

speckyspooky:
The issue is fiexed with PR #2029

PR #2029, opened 2025-01-13
DOCX, fix font-tag to avoid NPE and optimize the handling of font family and coloring (#2028)

speckyspooky:


speckyspooky:
@merks
Hi Ed, I offered this fix yesterday evening. The problem is that the build isn't running due to an issue with "org.mozilla.rhino". The message means that rhino couln'd be found at repo side. I tried the build yesterday evening 2times and this morning.
Have you directly an idea why this effect appears because my change doesn't effect "rhino" or the target file.
So I checked the "Orbit" project and the line of rhino has a special markup. May be we have from this side an effect:

grafik

**Original error build message:**
2025-01-14T05:14:37.7817224Z [ERROR] Failed to resolve target definition file:/home/runner/work/birt/birt/build/org.eclipse.birt.target/org.eclipse.birt.target.target: Could not find "org.mozilla.rhino/0.0.0" in the repositories of the current location

merks:
Yes, I had to move it from one the OSGi target platform to the BND target platform:

https://github.com/eclipse-orbit/orbit-simrel/commit/eb25760a449d0d7c6af4901b1993e4ece70b04b0

And while the builds are cascaded such that the aggregation build is kicked off by the OSGi and the BND build it can happen that the build seems older p2 update site content, which appears to have happened in this case because doing another build shows the expected new content:

image

merks:
I have these errors from the new version:

image

Investigating.

merks:
FYI, I have rebased the branch:

image

merks:
It looks like we are back on track...

speckyspooky:
Many thanks Ed!

PR #2030, opened 2025-01-14
Accommodate org.mozilla.rhino 1.8.0 API breakage

merks:


merks:
I'm not the expert here.  I'm just trying to get builds to work and keep libraries up-to-date.  

I didn't get an answer to this question:

https://github.com/mozilla/rhino/discussions/1788

I don't see a 1.7.5 release here:

https://repo1.maven.org/maven2/org/mozilla/rhino/

merks:
@hvbtup @speckyspooky @wimjongman 

I'm not 100% sure I should merge this.  Builds will remain broken until this is merged, and I don't think trying to stick to rhino 1.7.15 is a great option.  Note that the API changes in 1.8.0 are such that I changed the lower bound to 1.8.0, i.e., the new constructor we are using did not exist in 1.7.15:

image

merks:
@wimjongman 

In case it's not clear, I've already workaround around the missing OSGi metadata for the 1.8.0 version so we don't strictly require anything from rhino.  I was mostly just curious about the intent and about future expectations for future releases.

hvbtup:
> I don't see a 1.7.5 release here

Sorry, that was a typo. I meant1.7.15. Seems like the maintainers of Rhino have difficulties with updating their docs for new releases.

wimjongman:
@merks I'm, not sure what the issue is and what to do. Are you saying your changes are not needed, or are they needed when we upgrade to 1.8

merks:
I'm saying that I don't need the rhino folks to do anything in order to produce a bundle for 1.8.0 because that's already done in Orbit.

All the changes in this PR are necessary in order to compile against rhino 1.8.0 and in order for the tests to pass when running using rhino 1.8.0.  The only alternative to all these changes is to continue to use rhino 1.7.15 which in my opinion is a bad option.  So I think we should just hit the big green `Rebase and merge` button and move forward.

wimjongman:
Yes, let's go!

speckyspooky:
@hvbtup
Hi Henning, we have no issue with the new standard of ES6 because this is our default since I added the global settings of it.
So I expact no broken reports due to javascript ES6.

PR #2031, opened 2025-01-20
Replace o.e.gef.editparts.ZoomListener with o.e.draw2d.zoom.ZoomListener

merks:


merks:
@speckyspooky 

FYI, this was the reason for the failure of the master branch build.

speckyspooky:
Thanks for the fix!
Was it luck that you now exactly the change and what you have to do or had you others issues too that you know it.

merks:
The build builds against GEF’s latest build so when GEF removes a deprecated method we notice immediately.  It’s a bit painful but better sooner rather than later.

speckyspooky:
Sounds really special. Very good that the fixed could be done so fast from your side :o)

Issue #2037, opened 2025-01-30
PPTX: Slide number doesn't use text alignment (currently not implemented)


speckyspooky:
The pptx emitter support the slide number field (auto text) for page number.
The problem is that the text alignment of the slide numbering of element "a:fld" isn't implemented.

Therefore the text aligment isn't really good configurable.
I will add a change for the support of the text alignment.

PR #2038, opened 2025-01-30
Enhance the pptx emitter to support text alignment of page numbering (#2037)

speckyspooky:


merks:
I see yet a

merks:
@speckyspooky 

It looks like things are back on track.

Issue #2048, opened 2025-02-08
PDF, SVG background images converted to raster images (type PNG) and not used directly


speckyspooky:
The PDF emitter use SVG files for backround images. But the original SVG will be converted to raster images (type PNG).
So the quality of the images loose the precision. So I tried to created rounded border frames but the convertign is visible for the user with according zoom - see screen below.

I will add a change to integrate the SVG usage directly. For fall back reasons I will add a user property that the user can switch back,
but this won't be a standard option of the advanced register.

![Image](https://github.com/user-attachments/assets/6df4d16d-4233-4367-abe6-55820fe0da40)

hvbtup:
This annoyed me for so many years!

Great to see that you improve this!

speckyspooky:
The changes is finished with PR #2049

PR #2049, opened 2025-02-08
PDF emitter enhance the usage of SVG background images (#2048)

speckyspooky:


speckyspooky:
@hvbtup 
Hi Henning, this change enable the PDF-support of SVG images for background images.
The change include 2 classes which are also part of your current change #2018 

It contains the changes:
- PDFPage: contains the activation of the SVG usage directly
- PDFPageDevice: contains a new user property to disable the direct use of SVG images

The fallback if the SVG converting get's trouble is to use the raster image according to the current behavior.

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPage.java
```diff
@@ -178,20 +181,37 @@ protected void drawBackgroundImage(float x, float y, float width, float height,
...
+						&& (new File((new URL(imageUrl)).toURI().getPath())).exists()) {
+					image = transSVG(imageUrl, null, imageHeight, imageWidth);
+					if (image != null) {
+						drawImage(image, x, y, imageHeight, imageWidth, null);
+					}
 				}
-				imageWidth = img.getPlainWidth() / resolutionX * 72;
-				imageHeight = img.getPlainHeight() / resolutionY * 72;
+			} catch (Exception e) {
+				// default handling, use raster image
```
hvbtup, line 195: I think a warning message should be logged at least.

speckyspooky: log warning is added

hvbtup (approved):
Great!

merks:
@speckyspooky 

I'm not sure exactly your setup, but in general it's best to clone the original repo, have master checked out, do a pull, and only then start to make changes to avoid ending up in this state:

image

So I generally have things set up like this:

image

So master comes from origin and origin points at the original repository.  I have a different remote that points at my fork, and when I create a PR I create a new branch from master, modify the branch to use the merks remote instead of the origin remote to push changes to the changes to fork where I can create the PR.  This way I never need to concern myself with keeping my fork's master in sync with the original master.  I just have to remember to do a pull before I start changing things...

speckyspooky:
I don't know if we have a problem currently.
The situation is that I use allways the Master branch and sync for new PR.
Here we have the situation that 2 changes were made at same time on the same ressources.

But I hope we won't have any confilicts.

merks:
I see.  I wasn't sure.  I always avoid doing more than one change at a time.   😞 

In any case this cannot be merged so you will need to do something manually to move this forward...

Issue #2025, opened 2025-01-09
DOCX emitter images shown either distorted or truncated


hvbtup:
In one of my reports, I am using an image item with the following properties:

* width: not set
* height: not set
* proportionalScale: true
* fitToContainer: true

In the `onCreate` event of the image, I set the height to 22cm:

    this.height = "22cm";

In the PDF emitter, this has the desired effect that the full image is shown and it is not distorted (the aspect ratio is intact), independent of the image's actual image dimensions (eg if it's portrait or landscape). The setting causes a maximum width of the container available width and a maximum height of 22 cm, and the image uses at most this given space.

In my case, the image is more or left quadratic, and the available width is 161.9 mm, so this is rendered with a height of ~ 161mm:
![grafik](https://github.com/user-attachments/assets/ee404afa-8567-468b-9fbb-b15c40b635d9)

In the DOCX emitter, the effect is not as desired:
The emitter seems to ignore the combination that `proportionalScale` and `fitToContainer` are both true.
The image is rendered truncated:

![grafik](https://github.com/user-attachments/assets/84d6e528-47b0-4610-8b3b-8bb3fc274e51)

I attempted to fix this by also setting the maximum width to 161mm in the onCreate event.
In this case, the image is shown in full (not truncated), but it is distorted:

![grafik](https://github.com/user-attachments/assets/b151770c-892f-4fe6-b255-6c7fdb86a45f)

If you wonder, why I don't set the height property directly? Well, I have half a dozen of places in the report where an image comment can be shown. Setting the height directly causes a huge empty space in the designer, which makes editing the design a torture.

Note that the images in this report are not fixed, they are stored in the database and their dimensions can differ, some can be portrait, others landscape etc.

hvbtup:
Thomas, I hope you don't mind that I assigned this to you. You are the one with the best knowledge about how images are rendered in BIRT.

speckyspooky:
I will take a look into it if I have a free slot.

hvbtup:
This issue looks tightly related to #1549.

speckyspooky:
Short result of first tests, I started a retest based on the reports of #1406 which includes a report with "hook"-images and a report with a "butterfly"-image and it looks good.

I tested PDF & DOCX in version BIRT 4.14 & BIRT 4.18.

The result is that both BIRT-versions create the same output for PDF & DOCX.
I can confirm that the option "fit to container" hasn't the correct effect of DOCX like for PDF.
[...]

PR #2050, opened 2025-02-10
Enhance the output to support image option "fit to container" for HTML-preview and DOCX (#2025)

speckyspooky:


hvbtup:
I think it makes sense if you add the rptdesign that you used for testing (with the output shown #2025) to the example reports.
The photo of the butterfly is mine, feel free to use it in example or demo reports.

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/executor/ImageItemExecutor.java
```diff
@@ -138,15 +136,15 @@ private void processProportionalScale(ImageItemDesign imageDesign, IImageContent
...
 				double rw = 0;
 				double rh = 0;
-				String targetUnit = UNITS_IN;
-				if (width.getUnits().equalsIgnoreCase(height.getUnits())) {
+				targetUnit = DimensionType.UNITS_IN;
+				if (width != null && width.getUnits().equalsIgnoreCase(height.getUnits())) {
 					targetUnit = width.getUnits();
 					rw = width.getMeasure();
 					rh = height.getMeasure();
 				} else {
```
hvbtup, line 144: In the if statement, you added a check for width != null, probably for a reason.
If width is actually null, wouldn't this result in a NPE in the else branch?

engine/org.eclipse.birt.report.engine.emitter.html/src/org/eclipse/birt/report/engine/emitter/html/HTMLPerformanceOptimize.java
```diff
@@ -545,16 +547,58 @@ public void buildImageStyle(IImageContent image, StringBuffer styleBuffer, int d
...
+				}
+				DimensionValue compareImageWidthValue = DimensionUtil.convertTo(compareImageWidth.getMeasure(),
+						compareImageWidth.getUnits(), DimensionType.UNITS_MM);
+
+				if (compareImageWidthValue.getMeasure() > compareColumnWidthValue.getMeasure()) {
+					imgWidth = new DimensionType(compareColumnWidthValue.getMeasure(), DimensionType.UNITS_MM);
+					useHeightAutoValue = true;
+				}
+			}
+		} catch (Exception e) {
```
hvbtup, line 592: I don't like this "silently ignore all errors". 
What exceptions are to be expected here?
It may be ok to silently ignore some expected exceptions, but this is too general.
At least, a message should be logged.

hvbtup (changes requested):
I see some minor issues in the code (see my comments).
Great work overall.

engine/org.eclipse.birt.report.engine.emitter.html/src/org/eclipse/birt/report/engine/emitter/html/HTMLPerformanceOptimize.java
speckyspooky: Hi both, I confirm, unknown errors are not very friendly. 
The reason of it was that we have many unit-options and it could be that there are some options which can crash the calculation. I will double check for the logging option, this would be better instead of "silent dead".

speckyspooky:
Yes, I will add the used report too.

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/executor/ImageItemExecutor.java
speckyspooky: Removed.

engine/org.eclipse.birt.report.engine.emitter.html/src/org/eclipse/birt/report/engine/emitter/html/HTMLPerformanceOptimize.java
speckyspooky: After some additional testing I removed the try-catch completely.

hvbtup (approved):

Issue #2058, opened 2025-02-23
PDF, advanced glyp handling: Khmer Unicode (Hanuman Font) Not Rendering Correctly


speckyspooky:
### Discussed in https://github.com/eclipse-birt/birt/discussions/2040

<div type='discussions-op-text'>

<sup>Originally posted by **sinsarath** February  3, 2025</sup>
**Hello everyone,**

I'm using Eclipse BIRT Report Designer on macOS to generate reports in PDF format, but I’m facing issues with rendering Khmer Unicode (Hanuman Font). The Khmer characters either:

- Appear as blank or missing text, or
- Show incorrect  khmer characters.

**What I’ve Tried:**

1. Installed Khmer Hanuman Font on macOS and confirmed it works in other applications.
2. Set Khmer Hanuman Font in BIRT: In the report design, I applied Hanuman under Font properties of text elements.
3. Modified fontsConfig.xml:

- Located the file in my BIRT runtime folder and added:

```
<?xml version="1.0" encoding="UTF-8"?>
<font>
	<font-aliases>
		<mapping name="serif" font-family="Times-Roman" />
		<mapping name="sans-serif" font-family="Helvetica" />
		<mapping name="monospace" font-family="Courier" />
		<mapping name="cursive" font-family="Times-Roman" />
		<mapping name="fantasy" font-family="Times-Roman" />
	</font-aliases>

	<search-sequences>
		<sequence locale="zh_CN">
			<catalog name="Western" />
			<catalog name="Chinese_S" />
			<catalog name="Chinese_T" />
			<catalog name="Chinese" />
			<catalog name="Japanese" />
			<catalog name="Korean" />
		</sequence>
		<sequence locale="zh">
			<catalog name="Western" />
			<catalog name="Chinese_T" />
			<catalog name="Chinese_S" />
			<catalog name="Chinese" />
			<catalog name="Japanese" />
			<catalog name="Korean" />
		</sequence>
		<sequence locale="ja">
			<catalog name="Western" />
			<catalog name="Japanese" />
			<catalog name="Chinese" />
			<catalog name="Korean" />
		</sequence>
		<sequence locale="ko">
			<catalog name="Western" />
			<catalog name="Korean" />
			<catalog name="Chinese" />
			<catalog name="Japanese" />
		</sequence>
	</search-sequences>

	<composite-font name="all-fonts" font-family="Times-Roman">
	</composite-font>

	<font-paths>
		<path path="C:/windows/fonts" />
[...]

PR #2059, opened 2025-02-23
Enhance the PDF glyph handling to handle advanced kerning and ligatures (#2058)

speckyspooky:


hvbtup (changes requested):
This may feel like nit-picking, but enabling kerning and ligatures involves a little overhead and of course it changes the width of text, thus possibly resulting in different line-breaking.

There is probably a reason why this is not enabled by default in OpenPDF.
So I think it should not be enabled by default in BIRT, instead it should be an option for the PDF emitter.

speckyspooky:
Ok, make sense, so I will take a look into the emitter configuration.

hvbtup:
Furthermore, enabling kerning and ligatures must be considered in the font width calculation. I don't know if OpenPDF takes this into account.

Besides, it is somewhat strange that OpenPDF has a LayoutProcessor class with static methods.
So using kerning and ligatures is all or nothing and cannot be decided per font instance!?

Decades ago I added kerning and ligature support for the ReportLab Python PDF creation library in my "wordaxe" repo https://deco-cow.sourceforge.net/. However, I only ever tested this for latin languages.

speckyspooky:
I wouldn't see it to negative with the handling of openPDF.
Yes, there is may be a reason way the glyph handling is not enabled by default. I assume that it is more a performance topic instead of handling topic. Because the glyphs was handled before based of the different fonts but not correct add all.

The according article/reference-document of openPDF explain the topic and reference to different examples.
That a static way is used can be a reaction due to the complexity of openPDF and font-handling.

Please be aware the font handling of BIRT isn't easy too and I invested lot of time to figure out why the config-switch wasn't working.

hvbtup:
If we change things here, we should at least:
- Test and document what happens for an extreme text like "AV AV AV AV AV AV AV AV AV AV AV AV AV" when this text is used in a right-aligned paragraph or a justified paragraph.
- Test what happens when text is copied from the PDF into a text editor using Windows-1252 encoding when the text contains ligatures like ffl or ff.
- Test the tagged output with PAC 2024 if we create PDF/UA (the idea is similar to the previous test, but the input is based on the tagged PDF. Hint: I saw that `LayoutProcessor.java` contains a method to output text with a boolean argument actualText (probably meant to create an `ActualText` attribute in the PDF).

I don't say that all of this must work, but we should at least know what works and what not.
And if some of these tests show issues, then IMHO the default should be "no kerning and ligatures".

speckyspooky:
There are mixed some details, the copy of text to editor is only a hint of text display but isn't a test case.
PDF/UA will work due to embedded fonts, I don't see problems there.
The only Thing would be the case with AV and here till my current testing I have seen no effect.

hvbtup:
I agree with Thomas insofar as it it impossible to change the setting per font.
This is because the LayoutProcessor class uses static properties to store state and BIRT is multi-threaded.
If BIRT is used in a multi-threaded fashion, then calling a LayoutProcessor static method in thread A influences the rendering in thread B. 
**Thus LayoutProcessor is not thread-safe.**
A switch should be based on a system property to make it clear that it cannot be changed at runtime.

hvbtup:
[Temp.zip](https://github.com/user-attachments/files/18961093/Temp.zip)

A test report showing that enabling kerning breaks right-aligment and centered alignment, while justified text works.
This report is also a test case for https://github.com/eclipse-birt/birt/issues/2057

wimjongman:
> 
> @wimjongman & @merks What is your opinion to this siwtch option.
> 

Guys, I am just baffled with your knowledge about this topic. 🙏  Reading you comments, I think the logical thing to do is to make it an global switch which is off by default.

Great work!! ⭐⭐⭐⭐⭐

speckyspooky:
So, I figured out an alternative instead of a global switch with a system property.
I have added a new font-configuration-tag "font-kerning" to the fontsConfig*.xml.
With this configuration the kerning can be enabled or disabled (default: disabled/unused).

This switch can be configured at every font-xml because it is for all setups/configurations the same internal logic,
means all possible configuration will be checked at all font-config-files.

Let me know I can add in addition a system property but I think the config solution should be good enough.

engine/org.eclipse.birt.report.engine.fonts/fontsConfig.xml
```diff
@@ -146,6 +149,10 @@
 	this issue, please add the font path to this section.
 -->
 	<font-paths>
+		<path path="C:/Users/gutma/AppData/Local/Microsoft/Windows/Fonts/Hanuman-Regular.ttf" />
```
hvbtup, line 152: Adding your personal user directory to the BIRT fontsConfig.xml is probably not a good idea...

wimjongman:
How do people set this property? Does it have easy access?

engine/org.eclipse.birt.report.engine.fonts/fontsConfig_pdf.xml
```diff
@@ -44,4 +46,10 @@
 		<font font-family="HeiseiKakuGo-W5" catalog="Japanese" />
 		<font font-family="HYGoThic-Medium" catalog="Korean" />
 	</composite-font>
+	<font-paths>
+		<path path="C:/Users/gutma/AppData/Local/Microsoft/Windows/Fonts/Hanuman-Regular.ttf" />
```
hvbtup, line 50: Again, these personal directories should not go into default configuration.

engine/org.eclipse.birt.report.engine.fonts/fontsConfig.xml
```diff
@@ -48,6 +48,9 @@
 	currently "Times-Roman", to denote a place of missing character. 
 -->
 <font>
+	<!-- font advanced kerning ligatures will be used, values: false/unused (default), true/used -->
```
hvbtup, line 51: To be precise, kerning and ligatures are two different things, there is no "kerning ligature".
It's just that kerning and ligatures are both enabled with `LayoutProcessor.enableKernLiga()`.
Thus I think the wording (not only here) should be `KerningAndLigatures` in Java and `<kerning-and-ligatures enabled="false" />` in XML.
Or maybe even two distinct options, one for kerning and one for ligatures.

Besides, I think using fontsConfig.xml for this configuration is a great idea!

hvbtup (changes requested):
Please remove your personal directories from the config file.
I think the wording should be improved.

Apart from that: 
The overall code looks good to me and I like the idea of configuring this in fontsConfig.xml.

hvbtup:
> How do people set this property? Does it have easy access?

The fontsConfig.xml file is supposed to be editable, e.g. to add custom TTF font paths, so I would say: Yes, easy enough.

engine/org.eclipse.birt.report.engine.fonts/fontsConfig.xml
speckyspooky: removed

engine/org.eclipse.birt.report.engine.fonts/fontsConfig.xml
speckyspooky: removed

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/layout/pdf/font/FontMappingManager.java
```diff
@@ -54,10 +63,14 @@ public class FontMappingManager {
 			this.fontAliases.putAll(parent.getFontAliases());
 			this.fontEncodings.putAll(parent.getFontEncodings());
 			this.compositeFonts.putAll(parent.getCompositeFonts());
+			if (!this.fontKerningAdvancedUsage)
```
hvbtup, line 66: I think what you mean is: if it is not explicitly defined, get it from parent.
But then the variable should probably be a `Boolean` object and the test should be `fontKerningAdvancedUsae == null`.

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/layout/pdf/font/FontMappingManager.java
```diff
@@ -54,10 +63,14 @@ public class FontMappingManager {
...
+				this.fontKerningAdvancedUsage = parent.fontKerningAdvancedUsage;
 		}
 		this.fontEncodings.putAll(config.fontEncodings);
 		this.searchSequences.putAll(config.searchSequences);
 		this.fontAliases.putAll(config.fontAliases);
+		if (!this.fontKerningAdvancedUsage)
```
hvbtup, line 72: See above

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPage.java
```diff
@@ -599,7 +599,10 @@ private void drawText(String text, float textX, float textY, FontInfo fontInfo,
 					// PDF/A fallback font must be entered fully qualified with path and file name
 					String defaultFontPdfA = this.pageDevice.getDefaultFontPdfA();
 					if (defaultFontPdfA != null) {
-						font = BaseFont.createFont(defaultFontPdfA, BaseFont.IDENTITY_H, true);
+						if (!LayoutProcessor.isEnabled()) {
+							LayoutProcessor.enableKernLiga();
```
hvbtup, line 603: I think the font configuration should be checked in the `if` condition.
Without that check, this makes the whole configuration pointless.

hvbtup (changes requested):
In PdfPage.java, if I understand it correctly, the LayoutProcessor.enableKernLiga() is called without checking the option in fontsConfig.xml.
This makes the option pointless!?
And since the LayoutProcessor is basically a singleton, I think it makes sense to enable it (or not, depending on the option) at best exactly once, and not inside a frequently called method like `PdfPage.drawText`.

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/layout/pdf/font/FontMappingManager.java
```diff
@@ -70,26 +83,60 @@ public class FontMappingManager {
 		}
 	}
 
+	/**
+	 * Get the font mapping manager
```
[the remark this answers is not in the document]
speckyspooky: will be done

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPage.java
speckyspooky: eagle eye - will be done

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/layout/pdf/font/FontMappingManager.java
speckyspooky: The property is not the Object `Boolaen` it is the type `boolean` which is also located at parent side therefore this compare (and without null)

merks: I think the point is that you can't distinguish the three states:

- Explicitly set to false.
- Explicitly set to true.
- Not set at all.

with a boolean value that can only have two states...

speckyspooky: ok, may be I'm really slowly today, my property definition is done with an initial value and this is "false".
and a boolean (type) must have a value so which moment should switch false to "null". (I try only to understand this point.)

hvbtup: Line 66:
My first thought that your intention was "the property should inherit from parent only if not set explicitly".
Then the case "this should not use advanced kerning, even though the parent does" cannot work.
But now I wonder if the condition is necessary at all?
In lines 62 .. 65 the other properties are copied from parent without a condition. So why do you use a condition in line 66?

hvbtup (approved):
I still don't understand the two if statements in FontMappingManager.java, but if it works, so what...

PR #2060, opened 2025-02-25
Fix issue 2057: Take letter-spacing into account for width calculation

hvbtup:
Issue 2057 shows that letter-spacing was ignored in width calculations.
This PR solves this.
[issue-2057.zip](https://github.com/user-attachments/files/18963269/issue-2057.zip)


speckyspooky (approved):
Looks good!

Issue #2069, opened 2025-03-01
Font handling, optimization of missing collection types and java-docs (based on #2058)


speckyspooky:
During the change implementation of #2058 appears lot of untyped collections, maps and lists.
I will try to improve the untyped collections and add the missing java-doc.

PR #2070, opened 2025-03-01
Enhance font classes to set types at collections and maps and add java-doc (#2069)

speckyspooky:


engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/layout/pdf/font/FontSplitter.java
```diff
@@ -21,6 +21,12 @@
 
 import com.ibm.icu.text.Bidi;
 
+/**
+ * Font splitter
```
hvbtup, line 25: I think a bit more description would be useful, if our goal ist to improve the documentation and not just reduce the warnings count.

hvbtup (approved):
Some comments could be improved.
Apart from that, great work!

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/layout/pdf/font/FontSplitter.java
speckyspooky: Improved

PR #2076, opened 2025-03-06
Update the setup and the target platform to specify permanent sites

merks:
- Use Tycho 4.0.12 and CBI 1.5.2.

https://github.com/eclipse-birt/birt/issues/1997

### DIFF
--- a/build/org.eclipse.birt.releng/BIRT.setup
+++ b/build/org.eclipse.birt.releng/BIRT.setup
@@ -416,13 +416,13 @@
   <setupTask
       xsi:type="setup:VariableTask"
       name="birt.version"
-      value="latest">
+      value="4.19">
     <choice
         value="latest"
         label="BIRT Latest"/>
     <choice
-        value="4.18"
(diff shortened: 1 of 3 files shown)


speckyspooky:
@merks 
Hi Ed, after updateing my local workspace and execute the setup-task of BIRT modules
together with the cleaning task I get always the following error - see my screen.

So I switched the JDK from 17 to 21 (locally) but the message wasn't cleaned up.
Special is that the message is located to a message plugin. 

grafik

merks:
It looks like some error on the pom.xml.  Just delete them and hope they don't come back.  They probably won't.  We don't rely on m2e building anything for us in the IDE...

Also it looks fine for me:

image

Maybe it's something missing in the `~/.m2/toochains.xml`.  Let me know if the errors come back after deleting them...

speckyspooky:
Thanks for your checks.
Based on your hint I checked the pom.xml which was ok. But I had a changed preference file of this plugin in my staging area. After the deletion of it the errors were gone.

Issue #2102, opened 2025-03-21
DateFormatISO8601, Date Time converting with "parse(String source, TimeZone timeZone)" throw NullPointerException in source is empty


speckyspooky:
Discussion #2100

The mentioned discussion describe a problem that the date parsing method parse(String source, TimeZone timeZone) will throw a NullPointerException if the value of parameter "source" is empty.

This change is comming from PR aef9954 on Aug 3, 2018.

The problem is that the empty-check was moved to another method.
An additional change doesn't recognized the option of empty source-value and added an additional TimeZone call.
But if the source is empty then the converted date is empty and the TimeZone call produce an NPE.

I will create a PR to solve the problem.

speckyspooky:
@kbgentile
After reserach of your description I figured out what you mean. I created the PR #2103 to solve the problem.
It will be available with BIRT 4.20.

speckyspooky:
Is fixed with PR #2103

PR #2103, opened 2025-03-21
Fix NPE of date parsing at usage of DateFormatISO8601 (#2102)

speckyspooky:


kbgentile:
I did patch 4.19.0 runtime jar locally with an updated DateFormatISO8601 class by oputting source emoty/null check at top of parse method.  If fixed all of the issues I was seeing.  So doing the same for 4.20 will be great.  Thanks for the quick response.

speckyspooky:
The fix is merged. So with the next nigthly build of 4.20 you geht the fix.

PR #2104, opened 2025-03-22
Enhance outline search dialog 2101

speckyspooky:


speckyspooky:
Hello Steve & Ed,

I have tested the new search dialog and it is really a very good change.
During my testing I figured out small UI topics. This PR provide the changes of the findings.

My overview of the small changes:
1. CheckBoxes:
- changed the labes to clickable lables of the checkbox
- added tooltip-text to assist the understanding of the options

2. Result Table, Display:
- There was a refresh issue with the result table after filling, which was displayed if the cursor was hover over the line
- The table will be re-painted after filling

4. Result Table, Columns:
- swithced the column labels to the message file

5. Result Table, Element-Content:
- Element content contains originally the "handle.index", I changed it to element-id

7. Search, ID - Enhancement:
- The new option to search the design element-id is added
- This is a specific change because the element-id is not given by default at the property lists

Please check the changes and your are free to decide which of these changes will be merged or have to be removed.

merks:
I’m traveling and won’t get to this until Tuesday.

merks:
FYI it’s sufficient for me if @SteveSchafer-Innovent  approves.

speckyspooky:
Good trip, thanks for the note.

merks:
It seems better to amend a commit and force push it than to push multiple commits to a PR.

speckyspooky:
Ed, I know and I have used this, otherwise you would see more commits.
The first commit was created through git automatically because I had struggle with my local branch.

PR #2117, opened 2025-04-05
Regenerate the target platform for org.mortbay.jasper.apache 9.0.102

merks:


merks:
@speckyspooky 

FYI, this should fix the build failure you had in https://github.com/eclipse-birt/birt/pull/2116

PDE and Tycho now support version ranges in the *.target file so I'll eventually improve the generator to specify [9.0.0,10.0.0) so that this won't happen again when Orbit updates the 9.x version again.  There are both 9.x and 10.x versions so I need to ensure that the build picks up the one we're actually needing/using:

image

PR #2121, opened 2025-04-14
Provide *.aggr and corresponding *.aggran for Maven publishing

merks:
- Update the setup to install the CBI aggregator editors.


speckyspooky:
Hi Ed, it seems to be a problem with our PR, because it seems to be unaccessible:

2025-04-14T11:05:23.7953236Z ##[error]fatal: couldn't find remote ref refs/pull/2121/merge

merks:
Note that I merge this right after force pushing a fix which generated some errors.   There are current only to PR builds:

image

So I don't think there remain any problems.

PR #2122, opened 2025-04-15
Add command and key binding for SearchAction

SteveSchafer-Innovent:
In response to Thomas' request for a hotkey I've added a command and key binding for the SearchAction.  I don't like this solution  but I haven't found another way to do it.  It establishes an undesirable dependency from ViewContextMenuProvider to SearchHandler.

In order for SearchAction to work it needs a handle to TreeViewer.  A Command only has access to the Event object when executing and there doesn't appear to be a way to get to the TreeViewer through the Event object.  So I had to create a static variable.   ViewContextMenuProvider is pretty close to where the TreeViewer is instantiated.  Since ViewContextMenuProvider isn't visible from SearchHandler, but SearchHandler is visible from ViewContextMenuProvider, I had create the static variable in SearchHandler and set it in ViewContextMenuProvider.  

Please let me know if there's a better way I could have done this.

### DIFF
--- a/UI/org.eclipse.birt.report.designer.ui/plugin.properties
+++ b/UI/org.eclipse.birt.report.designer.ui/plugin.properties
@@ -100,6 +100,7 @@ command.name.28 = reloadCssStyleCommand
 command.name.29 = CopyCellContentsAction
 command.name.30 = CreateChartViewCommand
 command.name.31 = resetImageSizeCommand
+command.name.32 = searchAction
 context.name = reportdesignereditpart.context
 extension-point.name = elementAdapters
 extension-point.name.0 = BIRT Designer Report Item EditPart extension
--- a/UI/org.eclipse.birt.report.designer.ui/plugin.xml
+++ b/UI/org.eclipse.birt.report.designer.ui/plugin.xml
@@ -500,6 +500,11 @@
             id="org.eclipse.birt.report.designer.ui.command.resetImageSizeCommand"
             name="%command.name.31">
       </command>
+      <command
+            defaultHandler="org.eclipse.birt.report.designer.internal.ui.command.SearchHandler"
+            id="org.eclipse.birt.report.designer.ui.command.searchAction"
+            name="%command.name.32">
+      </command>
    </extension>
    <extension
          point="org.eclipse.birt.report.designer.ui.DNDServices">
@@ -876,6 +881,14 @@
                type="org.eclipse.birt.report.designer.internal.ui.dialogs.parameters.IHyperlinkParameterProvider">
          </adapter>
       </adaptable>
+   </extension>
+   <extension
+         point="org.eclipse.ui.bindings">
+      <key
+            commandId="org.eclipse.birt.report.designer.ui.command.searchAction"
+            schemeId="org.eclipse.ui.defaultAcceleratorConfiguration"
+            sequence="M1+F">
+      </key>
    </extension>
    
 </plugin>
--- a/UI/org.eclipse.birt.report.designer.ui/src/org/eclipse/birt/report/designer/internal/ui/command/SearchHandler.java
+++ b/UI/org.eclipse.birt.report.designer.ui/src/org/eclipse/birt/report/designer/internal/ui/command/SearchHandler.java
@@ -1,0 +1,44 @@
+/*******************************************************************************
+ * Copyright (c) 2004 Actuate Corporation.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * https://www.eclipse.org/legal/epl-2.0/.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ *
+ * Contributors:
+ *  Actuate Corporation  - initial API and implementation
+ *******************************************************************************/
+
+package org.eclipse.birt.report.designer.internal.ui.command;
+
+import org.eclipse.birt.report.designer.internal.ui.views.actions.SearchAction;
+import org.eclipse.core.commands.ExecutionEvent;
+import org.eclipse.core.commands.ExecutionException;
+import org.eclipse.jface.viewers.TreeViewer;
+
+/**
+ *
+ */
+
+public class SearchHandler extends SelectionHandler {
+	public static TreeViewer treeViewer = null;
+
+	/*
+	 * (non-Javadoc)
+	 *
+	 * @see
+	 * org.eclipse.core.commands.AbstractHandler#execute(org.eclipse.core.commands.
+	 * ExecutionEvent)
+	 */
+	@Override
+	public Object execute(ExecutionEvent event) throws ExecutionException {
+		super.execute(event);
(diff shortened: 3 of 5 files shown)


UI/org.eclipse.birt.report.designer.ui/plugin.properties
```diff
@@ -100,6 +100,7 @@ command.name.28 = reloadCssStyleCommand
 command.name.29 = CopyCellContentsAction
 command.name.30 = CreateChartViewCommand
 command.name.31 = resetImageSizeCommand
+command.name.32 = searchAction
```
merks, line 103: This is displayed to the user and even the other names I see above are not really appropriate as an English name to be displayed to a user.

UI/org.eclipse.birt.report.designer.ui/plugin.xml
```diff
@@ -500,6 +500,11 @@
             id="org.eclipse.birt.report.designer.ui.command.resetImageSizeCommand"
             name="%command.name.31">
       </command>
+      <command
+            defaultHandler="org.eclipse.birt.report.designer.internal.ui.command.SearchHandler"
+            id="org.eclipse.birt.report.designer.ui.command.searchAction"
+            name="%command.name.32">
+      </command>
```
merks, line 507: I don't think you should try to do this with a global handler.

UI/org.eclipse.birt.report.designer.ui/plugin.xml
```diff
@@ -877,5 +882,13 @@
...
       </adaptable>
    </extension>
+   <extension
+         point="org.eclipse.ui.bindings">
+      <key
+            commandId="org.eclipse.birt.report.designer.ui.command.searchAction"
+            schemeId="org.eclipse.ui.defaultAcceleratorConfiguration"
+            sequence="M1+F">
+      </key>
+   </extension>
```
merks, line 892: You can't take over this key for the entire Eclipse application.

UI/org.eclipse.birt.report.designer.ui/src/org/eclipse/birt/report/designer/internal/ui/command/SearchHandler.java
```diff
@@ -0,0 +1,44 @@
...
+package org.eclipse.birt.report.designer.internal.ui.command;
+
+import org.eclipse.birt.report.designer.internal.ui.views.actions.SearchAction;
+import org.eclipse.core.commands.ExecutionEvent;
+import org.eclipse.core.commands.ExecutionException;
+import org.eclipse.jface.viewers.TreeViewer;
+
+/**
+ *
+ */
```
merks, line 24: Don't add useless comments.  Either no comment or something meaningful.

UI/org.eclipse.birt.report.designer.ui/src/org/eclipse/birt/report/designer/internal/ui/command/SearchHandler.java
```diff
@@ -0,0 +1,44 @@
...
+
+public class SearchHandler extends SelectionHandler {
+	public static TreeViewer treeViewer = null;
```
merks, line 27: It's never necessary to initial a field to its default value (null, false, 0, and so on).

UI/org.eclipse.birt.report.designer.ui/src/org/eclipse/birt/report/designer/internal/ui/command/SearchHandler.java
```diff
@@ -0,0 +1,44 @@
...
+	 * @see
+	 * org.eclipse.core.commands.AbstractHandler#execute(org.eclipse.core.commands.
+	 * ExecutionEvent)
+	 */
+	@Override
+	public Object execute(ExecutionEvent event) throws ExecutionException {
+		super.execute(event);
+		SearchAction searchAction = new SearchAction(treeViewer);
+		searchAction.run();
+		return Boolean.TRUE;
```
merks, line 41: If there is not enough context then you are no doubt not using the right design.  After all, the treeViewer might be null and it might be disposed.

merks (changes requested):
Unfortunately this is just so wrong in too many ways.  This takes over Ctrl-F for the entire Eclipse application affecting all editors and views.  The lifecycle dependencies via static variables is just not a reasonable design.

Such a thing is normally done in an override of  org.eclipse.ui.part.EditorActionBarContributor.init(IActionBars) e.g., 

image

This ensures that whatever key binding the user chooses for Find will be used locally in that editor to do a find.  With multi-page editors the life cycle is still tricky to manage that you search properly in the current active editor's current active page...

SteveSchafer-Innovent:
I have questions.  Let me know if I should move this to a discussion.
1st question:  Once I subclass EditorActionBarContributor do I instantiate it or register it somewhere?  How does eclipse become aware of it?

merks:
No problem to do it here.  Technically you could force push an entirely new set of changes here to replace the ones you have.  What I do for that is remain the local branched used to create this PR, switch to master, do a pull, and then create a new branch with exactly the old name used for this PR.   You can commit to that branch and force push it (push with the force check box enabled) to update this PR.

---

No problem to ask here.  I should have been more proactive. All editors must have an action bar contributor, so I just set a breakpoint to see how it's constructed:

image

I see that org.eclipse.birt.report.designer.ui.editors.ReportActionBarContributor already exists.

But even here, it's really complicated with the fact that the action bar contributor is shared by the editors and given the lifecycle of the page activation and the fact that each page also must manage its contributions.

---

Not entirely a great design either, but certainly the quick and dirty approach is to add a key listener to the outline tree viewer:

image

Harshit99999123:
Hi all,
I am using 
<dependency>
    <groupId>org.eclipse.birt.runtime</groupId>
    <artifactId>org.eclipse.birt.runtime</artifactId>
    <version>4.4.2</version>
</dependency>  in my pom.xml for my Springboot application that just  creates and renders rpt design file and produces report via task. I have noticed that the first report after deployment or the report after long time of inactivity takes huge time. Can someone suggest me how to fix this issue?

SteveSchafer-Innovent:
It doesn't seem like using an ActionBarContributor is appropriate since I don't think this should be in an action bar or a menu since it's already in the context menu.  However I did find a way to prevent the hotkey from being globally enabled by implementing the canEnable function in the handler.   In that function I'm checking if the context is in the correct editor, the correct plugin, and the correct items are selected.  Seems to work correctly in my testing.    I have a couple of comments though:
1) In checking the editor I'm comparing the class name since the ReportEditorProxy class is not accessable from the handler.  Not sure if that's good practice.  
2) When I hit the hotkey on a tree item that shouldn't allow it, such as Variables, it correctly recognizes it as not enabled, however the selection goes away.  I.E. Variables is selected; I hit the hotkey; the search dialog does not come up (correct behavior); Variables is no longer selected.   I've tried identifying what's deselecting it but the selection event seems to be caused by a GTK native method, so I have no clue.

PR #2125, opened 2025-04-18
Enhance documentation of font configuration

speckyspooky:


hvbtup:
It should be noted that it is also possible to search for the font configuration files elsewhere, by calling

    // for example, -Dbirt.fonts.configuration=file:///C:/path/to/conf/birt on the command line

    fontsConfigurationURL = new URI(System.getProperty("birt.fonts.configuration") + "/fontsConfig.xml").toURL();
    engineConfig.setFontConfig(fontsConfigurationURL);

This will read the file `C:\path\to\conf\birt\fontsConfig.xml`

However, I don't know if/how the other font configuration files (eg the PDF specific) can be used at the same time.

speckyspooky:
@hvbtup I have added your example as an additional option.

PR #2127, opened 2025-04-20
Update *.aggr for Maven publishing

cristatus:
- Add maven-wrapped-<groupId|artifactId|version> as fallback
- Add mapping for apache derby
- Fix uk.co.spudsoft.birt.emitters.excel version


merks:
I looked at this in the IDE and it looks good to me.  This has no build impact, i.e., it's untested, so I skipped verification builds.

PR #2128, opened 2025-04-20
Update *.aggr for Maven publishing

cristatus:
- Exclude sampledb
- Exclude chart examples
- Remove `.feature.group` suffix from runtime artifacts

### DIFF
--- a/build/org.eclipse.birt.releng.maven/BIRT-Runtime-Minimized.aggr
+++ b/build/org.eclipse.birt.releng.maven/BIRT-Runtime-Minimized.aggr
@@ -40,6 +40,7 @@
   <mavenMappings namePattern="uk\.co\.spudsoft\.birt\.emitters\.excel" groupId="\$maven-groupId\$" artifactId="\$maven-artifactId\$" versionPattern="([^.]+)\.([^.]+)\.([^.]+)(\..*)?" versionTemplate="$1.$2.$3"/>
   <mavenMappings namePattern=".*" groupId="\$maven-groupId\$" artifactId="\$maven-artifactId\$" versionPattern=".*" versionTemplate="\$maven-version\$"/>
   <mavenMappings namePattern=".*" groupId="\$maven-wrapped-groupId\$" artifactId="\$maven-wrapped-artifactId\$" versionPattern=".*" versionTemplate="\$maven-wrapped-version\$"/>
+  <mavenDependencyMapping namespacePattern="osgi\.bundle" namePattern="org\.apache\.derby" groupId="*" artifactId="*" versionRangePattern=".*" versionRangeTemplate="[major.minor.micro,major+1)"/>
   <mavenDependencyMapping iuNamePattern="(?!.*(org\.eclipse\.)).*|org\.eclipse\.emf.*|org\.eclipse\.ecf.*|org\.eclipse\.jetty.*|org\.eclipse\.orbit.*" namespacePattern=".*" namePattern=".*" groupId="!" artifactId="!"/>
   <mavenDependencyMapping namespacePattern="osgi\.bundle|org\.eclipse\.equinox\.p2\.iu" namePattern="(?!.*(org\.eclipse\.(emf|ecf))).*" groupId="*" artifactId="*" versionRangePattern=".*" versionRangeTemplate="major.minor.micro-qualifier"/>
   <mavenDependencyMapping namespacePattern="java\.package" namePattern=".*" groupId="*" artifactId="*" versionRangePattern=".*" versionRangeTemplate="major.minor.micro-qualifier"/>
--- a/build/org.eclipse.birt.releng.maven/BIRT-Runtime.aggr
+++ b/build/org.eclipse.birt.releng.maven/BIRT-Runtime.aggr
@@ -38,6 +38,7 @@
   <mavenMappings namePattern="uk\.co\.spudsoft\.birt\.emitters\.excel" groupId="\$maven-groupId\$" artifactId="\$maven-artifactId\$" versionPattern="([^.]+)\.([^.]+)\.([^.]+)(\..*)?" versionTemplate="$1.$2.$3"/>
   <mavenMappings namePattern=".*" groupId="\$maven-groupId\$" artifactId="\$maven-artifactId\$" versionPattern=".*" versionTemplate="\$maven-version\$"/>
   <mavenMappings namePattern=".*" groupId="\$maven-wrapped-groupId\$" artifactId="\$maven-wrapped-artifactId\$" versionPattern=".*" versionTemplate="\$maven-wrapped-version\$"/>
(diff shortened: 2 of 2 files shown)


merks:
I don't think we should exclude axis overlay because these

merks:
FYI, I am current building a new version of the aggregator:

- https://github.com/eclipse-cbi/p2repo-aggregator/commit/2280b81c35c3aeb4808fa01e65cce9efb246058b
- https://ci.eclipse.org/cbi/view/p2RepoRelated/job/cbi.p2repo.aggregator-build/

The build infrastructure has been a bit flaky the last days, so keeping my fingers crossed.

Locally this lets me build a repository containing axis:

image

And without the feature IUs:

image

I prefer that we not remove any dependencies that are specified in the actual IU metadata at this point in time.  In the future, if they really aren't needed they should be removed from the actual bundles, or made optional,  but not simply filtered later on...

speckyspooky:
I would prefer to exclude the sampledb only for the minimized version and not for both version?
Because this db is very helpful and allows to create reports w/o specific db connection and the sampledb isn't so huge.

build/org.eclipse.birt.releng.maven/BIRT-Runtime.aggr
```diff
@@ -38,6 +39,7 @@
   <mavenMappings namePattern="uk\.co\.spudsoft\.birt\.emitters\.excel" groupId="\$maven-groupId\$" artifactId="\$maven-artifactId\$" versionPattern="([^.]+)\.([^.]+)\.([^.]+)(\..*)?" versionTemplate="$1.$2.$3"/>
   <mavenMappings namePattern=".*" groupId="\$maven-groupId\$" artifactId="\$maven-artifactId\$" versionPattern=".*" versionTemplate="\$maven-version\$"/>
   <mavenMappings namePattern=".*" groupId="\$maven-wrapped-groupId\$" artifactId="\$maven-wrapped-artifactId\$" versionPattern=".*" versionTemplate="\$maven-wrapped-version\$"/>
+  <mavenDependencyMapping namespacePattern="osgi\.bundle|org\.eclipse\.equinox\.p2\.iu" namePattern="(.*)\.(feature\.jar|sampledb.*|examples.*)" groupId="!" artifactId="!"/>
```
speckyspooky, line 42: I would not exclude the sampledb for this version.

merks:
Given the aggregation now excludes all features, defining a mapping for the feature is a no-op.  These are the contents of the repo currently:

image

These are the contents of the repo without the feature excluded:

image

So I think this PR should just be closed.

build/org.eclipse.birt.releng.maven/BIRT-Runtime-Minimized.aggr
```diff
@@ -40,6 +41,7 @@
   <mavenMappings namePattern="uk\.co\.spudsoft\.birt\.emitters\.excel" groupId="\$maven-groupId\$" artifactId="\$maven-artifactId\$" versionPattern="([^.]+)\.([^.]+)\.([^.]+)(\..*)?" versionTemplate="$1.$2.$3"/>
   <mavenMappings namePattern=".*" groupId="\$maven-groupId\$" artifactId="\$maven-artifactId\$" versionPattern=".*" versionTemplate="\$maven-version\$"/>
   <mavenMappings namePattern=".*" groupId="\$maven-wrapped-groupId\$" artifactId="\$maven-wrapped-artifactId\$" versionPattern=".*" versionTemplate="\$maven-wrapped-version\$"/>
+  <mavenDependencyMapping namespacePattern="osgi\.bundle" namePattern="org\.apache\.derby" groupId="*" artifactId="*" versionRangePattern=".*" versionRangeTemplate="major.minor.micro.1"/>
```
merks, line 44: Hard coding this to `.1` is eventually going to be wrong, so this is just a problem waiting to happen.  I believe it's currently mapping to a version range.   Is that the "problem" you are trying to address?  (Is it a problem?)  Maybe `.qualifier` would be more correct if you are trying to avoid a range.

build/org.eclipse.birt.releng.maven/BIRT-Runtime-Minimized.aggr
```diff
@@ -16,6 +16,7 @@
   <configurations operatingSystem="macosx" windowSystem="cocoa" architecture="aarch64"/>
   <configurations operatingSystem="macosx" windowSystem="cocoa" architecture="x86_64"/>
   <configurations architecture="x86_64"/>
+  <mavenMappings namePattern="org\.eclipse\.birt\.(.*)\.feature\.group" groupId="org.eclipse.birt" artifactId="org.eclipse.birt.$1"/>
```
merks, line 19: Features are excluded, they don't need mapping.

merks (changes requested):
One change looks pointless and the other not correctly.

build/org.eclipse.birt.releng.maven/BIRT-Runtime-Minimized.aggr
cristatus: Now, we don’t have an aggregating pom artifact, so we need to list all the dependencies manually.

In my opinion, we should only skip `.feature.jar` and not `.feature.group`. It’s generally more convenient to depend on the `.feature.group` and then exclude any unnecessary transitive dependencies, rather than having to manually specify every required dependency.

merks:
Derby needed quite special handling to turn it into an OSGi bundle.  😱  So we don't even know the actual exact maven version via the available metadata:

Perhaps we could go with this instead:

>   `<mavenDependencyMapping namespacePattern="osgi\.bundle" namePattern="org\.apache\.derby" groupId="*" artifactId="*" versionRangePattern=".*" versionRangeTemplate="[major.minor.micro,major+1)"/>`

I.e., use a range.

There are no features on any kind in the aggregation:

image

merks:
> I think excludeFeatures="true" should skip only the related .feature.jar artifacts, not the .feature.group itself. It's much easier to depend on the pom type and avoid listing all dependencies manually.

We will not publish features.  Not the jar nor the metadata.  It may be be easier to depend on a group of things (everything), but it's more correct to depend on exactly and only what you need.  (And, as we've seen, the people wanting things and needing things are not really so much interested in actually doing things.  As such, I'm more interested in doing what's easiest and most expedient.)

cristatus:
> Derby needed quite special handling to turn it into an OSGi bundle. 😱 So we don't even know the actual exact maven version via the available metadata:
> 
> Perhaps we could go with this instead:
> 
> > `<mavenDependencyMapping namespacePattern="osgi\.bundle" namePattern="org\.apache\.derby" groupId="*" artifactId="*" versionRangePattern=".*" versionRangeTemplate="[major.minor.micro,major+1)"/>`
> 
> I.e., use a range.
> 

It works!

merks:
Thanks for finding this problem.  👍 👏

Issue #2143, opened 2025-05-10
PDF, usage of "Webdings" font cause an exception (4.20 milestone)


speckyspooky:
The PDF creation cause an exception if the webdings font is used, exception: Only supported for True Type Unicode fonts. The problem is discovered at the enhancement of #2119 (discovered by @mr0reed137).

The reason of this effect is that specific symbol fonts can only be displayed with OpenPDF if the LayoutProcessor is enabled. With the change of #2119 the handling is disabled by default.

I will add a specific handling which will recognize the usage of specific symbolic fonts. If such font is used the kerning handling will be enabled.

speckyspooky:
The automatic is implemented and my testcases with Webding font were successfully.

Issue #2133, opened 2025-04-29
Content with Arabic Characters is incorrect for output format PDF since 4.17


rdoppler:
Upgrading our application with BIRT 4.19 (coming from 4.13) resulted in incorrect PDF output if content contains Arabic Characters. 
After trying out several BIRT versions we found out that the issue started with 4.17.
[Arabic_Character_Issue.zip](https://github.com/user-attachments/files/19958831/Arabic_Character_Issue.zip)

**Correct** PDF output (created with 4.13): 
![Image](https://github.com/user-attachments/assets/fef39e24-2b3d-4c38-a748-2403531e4a72)

**Incorrect** PDF output (created with 4.19):
![Image](https://github.com/user-attachments/assets/35b67ee2-1806-4a5d-8ece-e7d245510dd5)

For other output formats the content is processed correctly.

speckyspooky:
I can confirm that this issue occur with BIRT 4.17 and till BIRT 4.16 all is fine.
Currently I have no direct solution to fix it. An important hint is that the library which create the PDF-output-stream was updated from "OpenPDF 1.4.2" to "OpenPDF 2.0.3".
I don't know if the this version caused the topic. So it would need further investigations.

speckyspooky:
I have started next validation steps and the reason ist located into the combination of the used fonts and the changed font-glyph handling of OpenPDF (1.4.2 -> 2.0.3).

To figure out this fact, I restored the BIRT sources from BIRT 4.16 for PDF-output in combination with the current OpenPDF-version.
The result is the same according to thecreens.

**Solution**
Based on your screen I saw that you used a kind of default handling without specific defined font for PDF-creation.
I installed 2 Arabic fonts on my side:
-  Noto Sans Arabic Font: [Download](https://fontmeme.com/fonts/noto-sans-arabic-font/)
- DIN Next LT Arabic Regular: [Download](https://www.freefontdownload.org/en/din-next-lt-arabic-regular.font)

The 2nd step is to configure the fonts at the fon configuration file `fontsConfig.xml` and added the 2 arabic font installation path:
```
	...
	<font-paths>
		<path path="C:/Users/fonts/noto-sans-arabic/static" />		<!-- installation path of ttf-file -->  
		<path path="C:/Users/fonts/din-next-lt-arabic" />		<!-- installation path of ttf-file -->
		<path path="C:/windows/fonts" />
		<path path="d:/windows/fonts" />
		...
```

[...]

Issue #2119, opened 2025-04-09
PDF, ligatures will be used at specific fonts with letter spacing and disabled ligatures


mr0reed137:
Hi,

we introduced Birt 4.18 in our J2EE-Application and now face the problem, that the font contains ligatures, which was never a problem before. Is there a way to deactivate that, so the ligatures won't show up? I have never dealt with ligatures before, so I am not sure how to handle that.
Thanks for some hints and/or help on that.

Marco

speckyspooky:
Can you add a screen of your problem and for which output you have trouble PDF-format or other one?

With BIRT 4.19 there is added a new option to enable/disable ligatures for PDF.
But the default handling wasn't changed to handle ligatures in a special way at BIRT 4.18.

mr0reed137:
The problem occurs in pdf-format. I am not sure, if I can give a test case because we use an own font, which I might not be able to provide.

Can you tell me more about this option to enable/disable ligatures in pdfs? I am currently testing the latest snapshot of version 4.20, because we are interested in updating to 4.20, when released.

speckyspooky:
The special topic is to chage the fontsConfig.xml or you can use the fontsConfig_pdf.xml.
There is a new setup given very simple to switch on/off the kerning & ligatures for the font handling.

The configuration can be done with the following tag: `<kerning-and-ligatures enabled="false" />`

This is also given as a short documentation at the xml-file: [fontsConfig.xml](https://github.com/speckyspooky/birt/blob/05bdc9daefa01f8a91dd7de7eedd6232de89c6e4/engine/org.eclipse.birt.report.engine.fonts/fontsConfig.xml)

mr0reed137:
Thanks for the help.

So I did set the kerning-and-ligatures - tag and it raises two more questions:

1. When I have three configuration-files: fontsConfig.xml, fonstConfig_linux.xml and fontsConfig_pdf.xml, will they all be merged together (with the higher priorized settings to win) or will in this case only the fontsConfig_pdf.xml be used and the other two files ignored?

2. I did set the <kerning-and-ligatures enabled="true"/> and it works with it set to "true", not "false". Means the ligatures in my pdf-printoutputs will not show anymore. So am I missunderstanding the options-name? I would think with it set to true the ligatures will be printed and with it set to false the ligatures won't be printed...

speckyspooky:
Here the answers to your questions:
1. If you have 3 files which is the standard then there is an internal cascading used.

[...]

Issue #2043, opened 2025-02-07
BIRT PDF Khmer Unicode Font Not Rendering Correctly


sinsarath:
### Hello everyone,

I'm using Eclipse BIRT Report Designer to generate reports in PDF format, but I’m facing issues with rendering Khmer Unicode Font. The Khmer characters either:

- Appear as blank or missing text, or
- Show incorrect khmer characters.

I've tested generating a simple PDF using OpenPDF with the Hanuman font. Here’s the code I used:

```
import com.lowagie.text.Document;
import com.lowagie.text.DocumentException;
import com.lowagie.text.Font;
import com.lowagie.text.Paragraph;
import com.lowagie.text.pdf.BaseFont;
import com.lowagie.text.pdf.PdfWriter;
import java.io.FileNotFoundException;
import java.io.FileOutputStream;
import java.io.IOException;

public class Main {
	public static void main(String[] args) throws DocumentException, IOException {
		// Output PDF file
        Document document = new Document();
        try {
			PdfWriter.getInstance(document, new FileOutputStream("Khmer_Font_Test.pdf"));
		} catch (DocumentException | FileNotFoundException e) {
			// TODO Auto-generated catch block
			e.printStackTrace();
		}
        document.open();
        // Load Hanuman TTF font
        BaseFont hanumanFont = BaseFont.createFont("/Users/sinsarath/eclipse-workspace/SamplePdf/fonts/Hanuman.ttf", BaseFont.IDENTITY_H, BaseFont.EMBEDDED);
        Font font = new Font(hanumanFont, 14); // Set font size to 14
        // Add Khmer text
        String khmerText = "លេខ";
        document.add(new Paragraph(khmerText, font));

        document.close();
        System.out.println("PDF generated successfully!");
	}
}
```

Result are render not correct:

<img width="407" alt="Image" src="https://github.com/user-attachments/assets/f5a0f68e-35d2-46f3-a42a-334fef5d273c" />

It should be displayed like that:  **លេខ**

hvbtup:
This is an issue of OpenPDF as the discussion in #2040 showed.

PR #2144, opened 2025-05-10
Enhance the handling of symbolic fonts combined with kerning and ligature option (#2143)

speckyspooky:


hvbtup:
@vk-github18 @speckyspooky I'm a bit worried regarding multi-threading.

BIRT is supposed to work in a multi-threading environment.

But OpenPDF uses a **global LayoutProcessor** (a class, with static objects, or a singleton, I don't remember ATM), with some flags.
And we switch these flags on and off possibly several times during processing a BIRT report.

Of course this works in simple test cases, but isn't this going to run into bugs due to race conditions in a MT environement under load? And these kind of bugs are very, very hard to reproduce and to analyze later.

Probably this cannot be solved in BIRT at all, but requires a change in OpenPDF?

vk-github18:
Why would you change the flags? Can you provide a usecase and example?

Henning von Bargen ***@***.***> schrieb am Mo., 12. Mai 2025,
09:37:

hvbtup:
@vk-github18 Regarding "Why would you change the flags?" Because that's what https://github.com/eclipse-birt/birt/commit/2aef7c460b9d7cf8019849951616c18b7d2db04d does.

vk-github18:
LayoutProcessor is ment to be switched on before Processing and not to be
switched on and off

Henning von Bargen ***@***.***> schrieb am Mo., 12. Mai 2025,
10:37:

hvbtup:
Let me summarize the situation and please correct me if I'm wrong.

Since we updated to OpenPDF 2.0.3, several issues were reported regarding different non-latin fonts and ligatures with the Calibri font (#2143, #2133, #2119, #2058, #2043, #2033).

Thomas (with your help) tried his best to fix these issues, which all seem to be related to LayoutProcessor somehow (at least this is my impression).

Fixing the latest issue #2143 seems to require switching LayoutProcessor features on and off on a per-font basis before calling `drawText` (see the commit I mentioned above). This is a low-level internal routine which writes some text fragment in a given font, size and style at a given position to the page.

According to Volker, this is not really supported.

**So what can we do now?**

Note: Even if we had to fiddle with the LayoutProcessor settings only at report level instead of text fragment level, this would not remedy the situation. As long as OpenPDF uses global variables, which have to be set this way or that way for different fonts to achieve correct results, it is not really thread-safe.
The old iText library was thread-safe AFAIK, so shouldn't this be a design goal for OpenPDF as well?

speckyspooky:
If the LayoutProcessor is enabled then the kerning cannot be Set because the processor is already online.
But the LayoutProcessor must be online because the full disabling of the processor is controlled via enabling flag.
So it was my target to do it in this way.

Other ways will not work.

vk-github18:
The LayoutProcessor is only activated, if you call 
LayoutProcessor.enable*(). Otherwise it is not active and can not cause 
problems.

Using LayoutProcessor can help for correct rendering of foreign scripts 
or of Latin letters with accents, see e.g. DIN 91379.

You changed BIRT, included FOP in your classpath and activated 
FOPGlyphProcessor by this.

LayoutProcessor is designed to be called at the start with some options 
and not changed afterwards. It works only with OpenType-Fonts.
It should not be necessary to switch it off.


There was a bug with legacy fonts, that is fixed in the trunk of OpenPdf
and in https://github.com/openpdfsaucer/openpdfsaucer 2.0.9.
With this fix legacy processing is always used for legacy fonts.

So to be consistent, either
A: Don't use LayoutProcessor
B: Use LayoutProcessor for all glyph layout and use only OpenType fonts.

Again, LayoutProcessor is about glyph layout, it replaces the legacy 
glyph layout using the HarfBuzz layout engine built into recent Java 
versions resulting in correct representation of many scripts.

Kerning and ligatures are just minor options that can be switched on per 
font. If LayoutProcessor is disabled, these settings have no effect.



Am 12.05.25 um 14:01 schrieb Henning von Bargen:

speckyspooky:
The communication is to long.
I invested a lot of time for testing and my solution is working.

And I tested a lot of cases.
I can retest the cases of kerning but it the LayoutProcessor is sometimes misteriouse.

The current way is working for all issues which was addressed.

speckyspooky:
There is also e special case where the LayoutProcessor is to be enabled with usage flag "0".
Which means processor is active but without kerning and ligatures.

Because `LayoutProcessor.disabled() `is not like the same like `LayoutProcessor.enable(0)`.
I don't know if this is a bug or a feature or a special constellation of handling.

vk-github18:
> @vk-github18, can you point us to the bug which is fixed in OpenPDF and what change was made in OpenPDFSaucer?
See https://github.com/openpdfsaucer/openpdfsaucer/pull/40
and https://github.com/LibrePDF/OpenPDF/issues/1159
> 
> > So to be consistent, either
> > A: Don't use LayoutProcessor
> > B: Use LayoutProcessor for all glyph layout and use only OpenType fonts.
> 
> If I understand the issues I mentioned correctly, then for some reports we need to disable the LayoutProcessor, and for others we need to enable it. So we can choose neither A nor B.

LayoutProcessor should be able to handle all cases. If you can provide an OpenPDF-only example I can have a look.

vk-github18:
> There is also e special case where the LayoutProcessor is to be enabled with usage flag "0". Which means processor is active but without kerning and ligatures.
> 
> Because `LayoutProcessor.disabled() `is not like the same like `LayoutProcessor.enable(0)`. I don't know if this is a bug or a feature or a special constellation of handling.

I don't understand the argument `0` to enable.
And no -- calling `disable` is not the same `enable`.

hvbtup:
@speckyspooky 
I created an OpenPDF issue 1307 for this and in that issue Volker pointed to an OpenPDF example `GlyphLayoutDocumentKernLigaPerFont.java` which shows that the `LayoutProcessor` can be enabled once on startup and then kerning and ligature support can be specified on a per-font basis.

I think the code from commit https://github.com/eclipse-birt/birt/commit/2aef7c460b9d7cf8019849951616c18b7d2db04d and other places should be changed in such a way that is uses the same approach as the OpenPDF example, that is, enable or disable the LayoutProcessor only once at the beginning and then enable or disable kerning and/or ligatures using the LayoutProcessor methods `LayoutProcessor.setLigatures(font)` and `LayoutProcessor.setKerning(font)`.

@vk-github18  Though I find the API not quite clear: Is it disabled by default? I mean, I would expect a boolean argument as well...

vk-github18:
We should take our time to think about a good solution.
The default for Kerning and Ligatures is Off if you call
LayoutProcessor.enable(), and can be enabled with the set...() Methods.
Switching back and forth is not possible.
If you call Layoutprocessor.enableKernLiga()
Kerning and Ligatures are on and can not be switched off.

Henning von Bargen ***@***.***> schrieb am Mo., 19. Mai 2025,
14:35:

vk-github18:
Do you need the options Kerning and Ligatures? Or do you just need support
for complex scripts?
In this case just use LayoutProcessor.enable()
Without options.

Henning von Bargen ***@***.***> schrieb am Mo., 19. Mai 2025,
14:35:

vk-github18:
@hvbtup 
> enable or disable kerning and/or ligatures using the LayoutProcessor methods 
> `LayoutProcessor.setLigatures(font)` and `LayoutProcessor.setKerning(font)`.
`set` means `enable` here.
These methods should also only called once at the start.

hvbtup:
OK, so, if I understand everything correctly, the best practice would be:
- Call `LayoutProcessor.enable()` once at BIRT startup time.
- For kerning and ligatures, the default is false. Use `LayoutProcessor.setLigatures(font)` and/or `LayoutProcessor.setKerning(font)` (if the user wants to enable it for a specific font)exactly once for that, when the font is loaded for the first time (I recall vaguely that either OpenPDF or BIRT use a font cache).

For BIRT, this could be implemented as follows:

Have three global properties in fontsConfig.xml:

1. whether the LayoutProcessor shall be enabled at all (the default should be true)

Document that all other settings only work if this first property is true.

2. whether kerning shall be enabled by default (default: false)
3. whether ligatures shall be enabled by default (default: false)

These settings only specify the defaults and can be overridden per font.

For each font, optionally override the defaults from 2. and 3. :
4. Override kerning property from 2.
5. Override ligatures property from 3.

In the code, when a font is used for the first time (and only then):
Check if property 4 is specified for this font. If yes, use this value.
Otherwise, use the value from property 2.
If the value is yes, call `LayoutProcessor.setKerning(font)`, otherwise do nothing.

For ligatures, in the same way, use the values from the properties 5 and 3 and `LayoutProcessor.setLigaturesfont)`.

**Never** call `LayoutProcess.enableKernLiga()` or `LayoutProcessor.disable()`.
Note that the properties *cannot be set for individual reports*, but I guess that's OK.

I'm undecided about whether we need a special handling in the code for symbolic fonts.
Using the configuration options I described above, they can be configured in fontsConfig.xml.
OTOH modifying this file is not always easy, depending on the situation.
So the default for symbol fonts should "just work". If not, then we still need a special handling for these fonts, wich should be switched on by default. For these fonts, instead of the values from properties 2 and 3, the logic should use defaults for kerning and ligatures that just work.

speckyspooky:
Hi together,
the core problem was that the kerning and ligatures should be disabled to avoid automated handling at all.
So we cannot set by default the of LayoutProcessor to "true".
And there was also special handling of OpenPDF why `LayoutProcessor.disable()` was not working correctly therefore I used enabled(0).
And of course, the handling of symbolic fonts should be given by default. This was a side effect that I disabled all and the symbolic fonts are not displayed correctly. Which is now fixed.

I'm confused with the description of the comment:
grafik

Means this the method can only be called for this specific font 1 times or over all 1 times?

hvbtup:
I don't mind if the default for the LayoutProcessor is enabled or disabled.

vk-github18:
Again LayoutProcessor is about glyph layout, substitution etc for complex
scripts, not about Ligatures and Kerning, these are just options that can
be set if LayoutProcessor is enabled.
Kerning and Ligatures can be enabled once per font.
Disable works as expected and switches off LayoutProcessor, then you have
the old Layout of OpenPdf.
LayoutProcessor works with Opentype fonts and ignores other fonts.
Possibly you had a problem with a legacy font because of a bug fixed only
in the trunk but not in the latest release


Thomas Gutmann ***@***.***> schrieb am Di., 20. Mai 2025,
09:13:

hvbtup:
@vk-github18 
AFAIK we (BIRT) cannot use the trunk of OpenPDF, we must use a Maven release.
Do you know when the next release of OpenPDF is to be expected?

vk-github18:
I am not a maintainer, you should ask asturio or andreasrosdal.
There is also a fork openpdfsaucer maintained by andreasrosdal, where this
bug is fixed

Henning von Bargen ***@***.***> schrieb am Di., 20. Mai 2025,
09:44:

merks:
FYI, the latest BIRT nightly build is using today's release of openpdf:

https://repo1.maven.org/maven2/com/github/librepdf/openpdf/2.0.4/

PR #2153, opened 2025-05-17
Support Maven publishing

merks:


merks:
The build does not test this.  Merging.

PR #2161, opened 2025-05-20
Add org.eclipse.datatools.connectivity.oda.flatfile to engine.runtime

merks:


merks:
@NicolasPerritaz

Here is how we can add the dependency for aggregation:

I did a local build and tested aggregation locally to see that this works:

image

@speckyspooky 

I assume no one has concerns about augment the runtime a little bit...

speckyspooky:
I'm fine with it 👍

PR #2172, opened 2025-05-27
Synchronize System.getProperties() access

comtel2000:
Create a copy of the global shared `System.getProperties()` to prevent a `ConcurrentModificationException` from being thrown while iterating in `executionContext.registerBeans(systemProperties)`

### DIFF
--- a/engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/util/SecurityUtil.java
+++ b/engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/util/SecurityUtil.java
@@ -22,8 +22,13 @@ public class SecurityUtil {
 		return System.getProperty(name);
 	}
 
+	/**
+	 * Returns a copy of {@link System#getProperties()}
+	 *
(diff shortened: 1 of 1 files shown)


merks:
@comtel2000 

It appears you have not signed the Eclipse Contributor Agreement:

https://github.com/eclipse-ide/.github/blob/main/CONTRIBUTING.md#step-1-setting-up-an-eclipse-and-github-account

We need that to be able to merge PRs.

merks (changes requested):
I think this could be simplified.

merks:
Could you please squash the commits and force push a single commit.

Thanks for contributing!

merks (approved):
This looks good.  Also good to add the comment about the copying; I did check that the existing callers aren't modifying the returned result (unless it happens deeply in the engine somewhere).

merks:
@speckyspooky 

Do you have any concerns.  To me it looks like a good idea.

speckyspooky:
Of course, the change make sense. I'm fine with it.

Issue #2202, opened 2025-06-25
Slower performance with java17+


giangianoulas:
Hello

I am having a strange behaviour. I have a standalone application with all birt dependencies inside it. It is using birt v4.12 and compiled with jdk11. The application itself is very simple, it just sets up the environment for Birt Engine and creates the report.

With a given rptdesign & input files it runs 2-3 times slower with java 21 (~25 minutes) compared to java 11 (~11 minutes). The dataset is large so jvm is able to warm up in both cases.
The problem appears on both windows & linux.

What i tried:
Tried birt v4.17 => same result, but I can't test java 11.
Switch GC & JIT options to mimic java 11, to no avail this far.

The created report is identical in both cases.

Was this reported before? Or any pointers would be greatly appreciated!

giangianoulas:
No it did help, in fact it doesnt even reach the limit I had already set. GC activity looks the same. for jdk 11 & 21. It is either a JIT things, althout i tryed all possible settings i could find, i added all possibe --add-opens & --add-exports i could but still the same.

IO activity is virtually double on java 11, but i am not sure if it the cause or the symptom

giangianoulas:
yes, everything is identical. Maybe it has to do with Nashorn removal with JDK15? But then again i get same runtimes with v4.17.
I will try latest birt and hope for the best, and try to add Nashhorn as dependency

giangianoulas:
Well, final comment, i tried birt 4.20, same story.
I did the profiling again, just in case i got wrong

--------------------------------
Java 11

![Image](https://github.com/user-attachments/assets/d2998d30-c69c-44ae-9488-a2d35451fe4c)

--------------------------------
Java 21

![Image](https://github.com/user-attachments/assets/28d18e19-2ca8-46ca-ba3c-11fbe460a717)

Conlusion the IOUtil.readInt method is much slower with java21 because of the slower ByteArrayInputStream.read().
Maybe `org.eclipse.birt.core.util.IOUtil::readInt` can be adjusted to read what it needs in such a way that it can be optimized.
On the positive side `java.io.DataInputStream::readUTF` is now faster.

That got me curious:
Some benchmark test (Mostly AI generated but shows the result)

```java
package com.example.benchmark;

import org.openjdk.jmh.annotations.*;
import org.openjdk.jmh.infra.Blackhole;

import java.io.ByteArrayInputStream;
import java.io.FilterInputStream;
import java.io.IOException;
import java.io.InputStream;
[...]

PR #2204, opened 2025-06-27
Read 4 bytes in to allow optimization of readInt during runtime #2202 issue

giangianoulas:
Attempt to fix the bottleneck


merks:
Do you know if the stream from which it’s reading is buffered? If not, that would be perhaps a more direct improvement impacting other parts of the read operation.

merks:
In the other   PR it seems a byte array input stream is involved. So definitely buffered.

PR #2205, opened 2025-06-27
Read 4 bytes in to allow optimization of readInt during runtime #2202 issue

giangianoulas:
Try to allow optimization during runtime, as per findings of the issue #2202

### DIFF
--- a/core/org.eclipse.birt.core/src/org/eclipse/birt/core/util/IOUtil.java
+++ b/core/org.eclipse.birt.core/src/org/eclipse/birt/core/util/IOUtil.java
@@ -158,10 +158,17 @@ public class IOUtil {
 	 * @throws IOException
 	 */
 	public final static int readInt(InputStream inputStream) throws IOException {
-		int ch1 = inputStream.read();
-		int ch2 = inputStream.read();
-		int ch3 = inputStream.read();
-		int ch4 = inputStream.read();
+		byte[] bytebuffer = new byte[4];
+		int bytesRead = inputStream.read(bytebuffer, 0, 4);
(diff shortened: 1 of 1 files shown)


giangianoulas:
Complete call tree 
image

merks:
This does appear to because an actual test failure:

https://ci.eclipse.org/birt/job/build/job/PR-2205/1/testReport/org.eclipse.birt.report.engine.executor/ListingElementExecutorTest/testPageBreakInterval/

Other than the synchronized there doesn't really appear to be anything expensive in this call stack:

image

So it's hard to imagine that this would make a significant overall difference...

core/org.eclipse.birt.core/src/org/eclipse/birt/core/util/IOUtil.java
```diff
@@ -158,10 +158,17 @@ public final static void writeShort(OutputStream outputStream, short value) thro
 	 * @throws IOException
 	 */
 	public final static int readInt(InputStream inputStream) throws IOException {
-		int ch1 = inputStream.read();
-		int ch2 = inputStream.read();
-		int ch3 = inputStream.read();
-		int ch4 = inputStream.read();
+		byte[] bytebuffer = new byte[4];
```
hvbtup, line 161: I think one should take DataInputStream.readInt() as a template.
Note that DataInputStream uses only one static 8 bytes buffer, while your code creates a new buffer on the heap *for each call* and then calls ByteArray.getInt to convert the buffer to an int.

giangianoulas:
"Fixing it" will not make a difference. It has to do with removal of biased locking with java 18

PR #2208, opened 2025-06-30
Use a modified version of java.io.ByteArrayInput/OutputStream with re…

giangianoulas:
Copy & modify java.io.BytaArrayInputStream & java.io.BytaArrayOutputStream.
Mainly remove the syncronized keyword in an effort to avoid the overhead of the threads waiting when in this case only one thread accesses the object. With this the throughput and overall execution time is slightly more than what it was with jdk 11, but still far better than java 15+.

I think the licensing notice is ok, and also the classpath exception.

Relevant  #2202

### DIFF
--- a/LICENSE-OPENJDK.txt
+++ b/LICENSE-OPENJDK.txt
@@ -1,0 +1,40 @@
+This file contains the licensing terms for portions of this project derived
+from OpenJDK source code, specifically the classes:
+
+- java.io.ByteArrayInputStream
+- java.io.ByteArrayOutputStream
+
+-------------------------------------------------------------------------------
+GNU GENERAL PUBLIC LICENSE
+Version 2, June 1991
+
+Copyright (C) 1989, 1991 Free Software Foundation, Inc.
+51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA
+
+Everyone is permitted to copy and distribute verbatim copies
+of this license document, but changing it is not allowed.
+
+Full text available at:
+https://www.gnu.org/licenses/old-licenses/gpl-2.0.txt
+
+-------------------------------------------------------------------------------
+"CLASSPATH" EXCEPTION TO THE GPL
+
+Certain source files distributed by Oracle are subject to the following
+clarification and special exception to the GNU General Public License, version 2:
+
+You may link this library with independent modules to produce an executable,
+regardless of the license terms of these independent modules, and you may copy
+and distribute the resulting executable under terms of your choice, provided
+that you also meet, for each linked independent module, the terms and conditions
+of the license of that module.
+
+An independent module is a module which is not derived from or based on this
+library. If you modify this library, you may extend this exception to your
+version of the library, but you are not obligated to do so. If you do not wish
+to do so, delete this exception statement from your version.
+
+-------------------------------------------------------------------------------
+This project includes modified versions of classes originally from the OpenJDK
+project (https://openjdk.org/), redistributed under the terms of GPLv2 with the
+Classpath Exception.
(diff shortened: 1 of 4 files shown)


hvbtup (commented):
I understand the idea (I had it myself).
Regarding naming, maybe the class names should contain the word `Unsychronized` to make it clear that they are not thread-safe and that's the only difference from the standard ones from the SDK.
What do others think?

Regarding the License, I'm not an expert. @wimjongman, can you check if it's ok?

wimjongman:
Comment from Ai:

You're right to be concerned about copying JDK classes—duplicating standard library code is typically discouraged due to:

- **Maintenance Burden:** Updates/fixes to the JDK won't propagate to your code.
- **Licensing:** You must comply with the JDK's (GPL+Classpath Exception) license and track modifications.
- **Code Bloat:** More code to maintain/test/document.

### Alternatives to Copying JDK ByteArray Streams

#### 1. **Use Existing JDK Classes (`java.io.ByteArrayInputStream` / `ByteArrayOutputStream`)**
- **Pros:** Well-tested, maintained, and usually “fast enough” for most applications.
- **Cons:** Methods are synchronized, which can be a (small) bottleneck in extremely high-throughput, single-threaded scenarios.

#### 2. **Use Apache Commons IO / Guava**
- **[Apache Commons IO](https://commons.apache.org/proper/commons-io/javadocs/api-release/org/apache/commons/io/output/ByteArrayOutputStream.html):**  
  `org.apache.commons.io.output.ByteArrayOutputStream` is a drop-in, unsynchronized replacement, and well-tested.
- **[Guava](https://guava.dev/releases/31.1-jre/api/docs/com/google/common/io/ByteStreams.html):**  
  Guava's `ByteStreams` utility methods can help, though they don't provide a full replacement class.

#### 3. **Wrap the JDK Classes for Your Use-case**
- Create a simple wrapper or delegate class that only exposes the methods you need, and use composition to avoid accidental synchronization.
- *However*, the internals will still be synchronized unless you reimplement the logic.

#### 4. **Refactor Code to Avoid the Bottleneck**
- Reuse a single buffer or output stream instance per-thread.
- Profile to confirm that `synchronized` is really the hot spot; often, the overhead is negligible.

#### 5. **Use Thread-local or Pooling**
- Use a `ThreadLocal<ByteArrayOutputStream>` pool if the cost comes from repeated allocation, not synchronization.

#### 6. **Implement a Minimal Version Yourself**
- If you really must, implement a minimal, unsynchronized version tailored to your needs (not a full JDK clone). This is what Apache Commons IO does.

---

### **Recommendation**

**Best:**
- Use `org.apache.commons.io.output.ByteArrayOutputStream` (unsynchronized, widely used, no JDK code copying).
  ```java
  import org.apache.commons.io.output.ByteArrayOutputStream;
  ```

**Or:**
- Profile first; if `synchronized` is not a bottleneck, use the JDK classes.

**Avoid:**
- Copying/modifying JDK source directly unless there is no alternative.

---

**Summary Table**

| Solution                               | Synchronization | Maintenance | Risk | Recommended |
|-----------------------------------------|----------------|-------------|------|-------------|
| JDK Classes                            | Yes            | Low         | Low  | ✔️          |
| Apache Commons IO                       | No             | Low         | Low  | ✔️✔️        |
| Copy JDK Source                        | No             | High        | High | 🚫          |
| Custom Minimal Implementation           | No             | Medium      | Medium | ⚠️        |

If you want, I can show you how to refactor to use Apache Commons IO or discuss profiling strategies. Let me know your preference!

wimjongman (changes requested):
-1 for copying JDK classes. Open to other solutions.

giangianoulas:
> From their naming, it seems like Apache Commons `UnsynchronizedByteArrayInputStream` and `UnsynchronizedByteArrayOutputStream` is what fits best here.

I think the best alternative requiring minimal work and satisfying everyone is the apache commons-io but i have no idea how to add a dependency with this build system. I hope someone here takes the time to do it.

PR #2209, opened 2025-06-30
In ResultObjectUtil.readData synchronize on the ByteArrayInputStream

merks:
https://github.com/eclipse-birt/birt/issues/2202


merks:
@giangianoulas 

Looking more closely this morning at the code and at your call stack in the issue I realize that one should also try to synchronize the `bis` argument which I've force-pushed to this PR.  Could you please try this version of the PR?

merks:
BTW, I'm a reluctant to add a new dependency because we'll have to be quite sure that the horrible ant scripts do the right things with those and I really distrust the dark magic of those scripts...

merks:
It's in the target platform for sure, but I didn't see any explicit use of it in any BIRT bundle or feature.  So it's quite likely that some dependency is using it already because  indeed I see it here:

image

And now I can see that poi is using it:

image

So it really wouldn't be a big deal or risky to add it here.

If someone wants to try that, please use a package import to specify add dependency in the MANIFEST.MF.  I might find time later in the week if we're all agreed this is the desired approach...

PR #2213, opened 2025-07-07
In ResultObjectUtil use unsynchronized byte array stream variants

merks:
- Use org.apache.commons.io.input.UnsynchronizedByteArrayInputStream instead of java.io.ByteArrayInputStream
- Use instead org.apache.commons.io.output.UnsynchronizedByteArrayOutputStream of java.io.ByteArrayOutputStream

https://github.com/eclipse-birt/birt/issues/2202


giangianoulas:
If the black box scripts work then this is the solution.

speckyspooky:
I'm followed the communication of the topic.
I'm fine with the changes. The best result would be if we have a possitve performance effect.

wimjongman:
Thanks, Ed and Giannis. For taking the time to report, patch and test it. 🙏

PR #2227, opened 2025-07-20
Update to Tycho 4.0.13

merks:
- Use export MAVEN_OPTS="-XX:MaxRAMPercentage=30" to avoid heap space errors creating the update site zip.
- Use regenerated target platform with stricter versions for "org.mortbay.jasper to avoid resolving the 10.x versions.


speckyspooky:
Ok, understand thanks for the info. So I know why we use here the specific version number.

Issue #2247, opened 2025-08-01
Report engine, a NPE will be thrown if a report contains invalid items on the report structure


speckyspooky:
Issue based on discussion #2240

If a report contains invalid items then the report engine cannot generate the report and throw an null pointer exception.

<img width="1635" height="780" alt="Image" src="https://github.com/user-attachments/assets/f6557ce9-cbe9-48ba-b8c3-62cc18225337" />

The exception will be thrown through the class "LocalizedContentVisitor". The class tries to fetch the type information of the items. But in the case of invalid item the object is null and so we got the NPE.

I will provide a fix for it.

Demo error-report: [error-report_cmds_2.txt](https://github.com/user-attachments/files/21554364/error-report_cmds_2.txt)

PR #2248, opened 2025-08-01
Fix NPE of invalid items at content visitor (#2247)

speckyspooky:


hvbtup:
What is the runtime behavior in this case now?
Does it fail with a more usable error message?

I mean, I really, really *dislike* BIRT's attitude to silently ignore errors, or in many cases (scripting errors) to just treat them as warnings (like a browser does) and just continue.

Report generation *must* fail if there is an invalid item in the report.
For production quality, it is essential that the report output is correct.
A failed report with a reasonable error message is much better than a wrong or incomplete report, except when the report is run from the BIRT designer during development.

PR #2258, opened 2025-08-07
Revert "restrict access to encrypted odaPassword from java script"

DaniEll-AT:
Revert #575 according to discussion #2255


merks:
We need your account to have an associated ECA to be able to merge PRs:

https://github.com/eclipse-ide/.github/blob/main/CONTRIBUTING.md#step-1-setting-up-an-eclipse-and-github-account

speckyspooky:
@DaniEll-AT & @hvbargen 
With this change it is possible to fetch the encrypted oda-password from script side.
This was the original target of the change to avoid this kind of fetch.

So I have no good arguments why this change should be done if we get a kind of leak for the password-topic.

speckyspooky (approved):

merks:
All checks have passed. Someone should merge.

PR #2262, opened 2025-08-20
Fix problems in org.eclipse.birt.report.viewer for Jetty 12.1.0

merks:
- Add requirement to org.eclipse.jetty.ee.webapp to avoid transitive classpath errors.
- Update all the jetty lower bounds to 12.1.0.
- Change jetty-deploy.xml to use o.e.jetty.deploy.StandardDeployer.StandardDeployer(ContextHandlerCollection)


merks:
@speckyspooky  @claesrosell 

FYI, when I tested the new Jetty 12.1.0 release the viewer was broken yet again.  I made an educated guess on fixing the deployment file so that the viewer works again.

image

I believe this means that Jetty 12.1.0 is required and that older Jetty versions will not work.

speckyspooky:
Thanks for your Information and the changes!

PR #2263, opened 2025-08-25
Fix viewer and remove unnecessary log ouput commands

speckyspooky:


speckyspooky:
@merks
Hi Ed, I have added a very small change to remove log outputs at commandline.
With the PR creation the eca check means I'm not authorized.

I verified ECA and it means I'm ok.
Do you know what I can do to solve it?

grafik

merks:
I expect it's some error in some server that's involved.  Given it's you, I would just ignore it.  If it was a PR from a contributor involved, I would want to ensure that it's not a real problem.

PR #2270, opened 2025-09-06
Fix html execution run method

speckyspooky:


merks:
I just ignored it.

merks:
Yes, looks like a problem with the email address

image

versus

image

I guess I should not have assumed that the ECA was broken yet again (but it happens with enough regularity that one starts to assume it happens again).

PR #2288, opened 2025-09-13
Enable running tests for uk.co.spudsoft.birt.emitters.excel.tests

merks:
- Comment out all failing tests with `@Ignore // FIXME`


speckyspooky:
Thanks for the information.

Issue #2292, opened 2025-09-18
Enhance the scripting option to set the height / width property of report items (based on #2280)


speckyspooky:
The support-ticket #2280 discussed the option to add setter-methods for width / height property to objects which have a getter for one of the both properties.
After a research the following items are figured out and will get the new setter methods for the 2 parameter options "unit string" and "dimension value" as necessary parameter value.
The items are:
- CellHandle
- ColumnHandle
- CrosstabCellHandle
- CrosstabReportItemHandle
- MasterPageHandle
- RowHandle

I will create a PR for the change.

hvbtup:
Setting a width or height for a cell in a grid, table or cross tab doesn't make sense!

PR #2293, opened 2025-09-18
Enhance items with setWidth or setHeight method (#2292)

speckyspooky:


merks:
@speckyspooky 

I really need to get support for version ranges in place for the target platform...

I rebased your PR which should help push it forward.

PR #2324, opened 2025-10-22
Fix infinite loop in PDF generation with oversized avoid rows (#2319) 

### LINKED ISSUE #2319: Infinite loop when generating a PDF with a table with fixed height and page-break-inside: avoid
**1. The Problem: What's Happening?**

An infinite loop can occur during PDF generation under a specific set of conditions:
- A report design contains a table with a fixed height.
- A row within that table has the page break property set to page-break-inside: avoid.
- The content of that single row is too tall to fit on the current page.

When these conditions are met, the BIRT layout manager gets stuck. It tries to place the row, sees that it overflows the page, but is forbidden from splitting it due to the avoid property. It fails to make any progress, retrying the same failed operation endlessly until the print job times out.

**2. Root Cause Analysis** 

The infinite loop is not caused by a single bug but by a harmful interaction between several methods in the layout engine. When a row is too tall to fit on an empty page and cannot be split, the engine fails to advance its state, leading to a loop.

_Key Contributing Components:_
- PDFTableBandLM.addToRoot: This method is the first point of failure. It correctly identifies that the tall row overflows the page, but it refuses to add it, even to an empty page. It simply returns false.
- PDFBlockStackingLM.addArea: This method calls addToRoot but critically ignores its false return value. Because it doesn't know addToRoot failed, it doesn't trigger a page break or advance the layout cursor. The state remains unchanged.
- RowArea.split: When the layout engine forcefully tries to split the content to make progress, this method is called. However, if the row is marked with page-break-inside: avoid, it returns SUCCEED_WITH_NULL, which essentially means "I succeeded by doing nothing." This response doesn't signal an error or a need for a page break.

_How the Loop Occurs:_
The retry mechanism is driven by BlockContainerArea.split.
- The layout engine attempts to place the unsplittable, oversized row. addArea calls addToRoot, which returns false (no progress).
- The engine detects an overflow and forcefully calls split to try and break the content across pages.
- RowArea.split respects the avoid property and returns SUCCEED_WITH_NULL without actually splitting anything (no progress).
- The BlockContainerArea sees this "successful" null split and does not change its internal state.
- The parent caller sees that the content still doesn't fit and triggers the same auto-page-break logic again, returning to step 
- This cycle repeats indefinitely.

neilpelow:
Return BEFORE_AVOID_WITH_NULL when RowArea._split() cannot split due to page-break-inside: avoid, preventing infinite retry loops.

Fixes #2319

### DIFF
--- a/engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/nLayout/area/impl/RowArea.java
+++ b/engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/nLayout/area/impl/RowArea.java
@@ -415,6 +415,11 @@ public class RowArea extends ContainerArea {
 		}
 		updateRow();
 		needResolveBorder = true;
+		// If force is true but we couldn't split anything, return BEFORE_AVOID_WITH_NULL
+		// to signal that the row cannot be placed on the current page, preventing infinite loops
+		if (force) {
+			return SplitResult.BEFORE_AVOID_WITH_NULL;
+		}
 		return SplitResult.SUCCEED_WITH_NULL;
 	}
 


hvbtup:
(Same text as in my comment on the issue):
Since there still are conflicting requirements, at least one of the requirements cannot be fulfilled.
So, what happens now, with your change in effect, ...
* in your original test case?
* in more typical cases, when there is no height specified for anything, but the table row eg contains another table or grid which has page-break-inside: avoid set and, when the set of rows of the inner table doesn't fit into the page?

I think the infinite loop was undefined behavior in an edge case.
It should be documented in the issue how this PR changes the behavior.
I mean, it avoids an endless loop, but what is the outcome instead? Will the content be truncated (text is lost?), does the report generation fail, or does it continue, and if so, with an error message added? What happens in the speical case of a *header* row that is too tall?

hvbtup:
Your text looks like it is generated by AI and I can't imagine how the conflicting goals can be achieved at the same time.
Can you please provide the test reports and their results with the PDF emitter?

PR #2347, opened 2025-12-11
Use com.ibm.icu.text.Collator instead of String.compare by default

merks:
https://github.com/eclipse-birt/birt/issues/2329

### DIFF
--- a/data/org.eclipse.birt.data/src/org/eclipse/birt/data/engine/api/script/BaseScriptEvalUtil.java
+++ b/data/org.eclipse.birt.data/src/org/eclipse/birt/data/engine/api/script/BaseScriptEvalUtil.java
@@ -50,6 +50,8 @@ import org.eclipse.birt.data.engine.i18n.ResourceConstants;
 import org.mozilla.javascript.Context;
 import org.mozilla.javascript.Scriptable;
 
+import com.ibm.icu.text.Collator;
+
 /**
  * Static utility methods to help evaluating Javascript expressions
  *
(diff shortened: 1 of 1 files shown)


merks:
I think this is a good change but I want to be sure that you **both** agree this this is a good change.

hvbtup (dismissed):
Thinking about it, I think there *is* a risk, because if we change the comparison function, this probably also changes which strings are considered equal, and not only the sorting order.

Eg if some of `A`, `a`, `Ä` and `A` plus `COMBINING DIARESIS` are now considered to be equal, this could possibly break existing logic which uses data from data sets and stores it in JS data structures.
It could also harm reports with grouping, because the grouping changes.

What I mean is:
With the old code, there was a one-to-one correspondence between a string (seen as a sequence of Unicode code points) used for grouping and the group, and the sorting. With the new code, if the default Collator considers A and AE to be equal, then two different strings (seen as Unicode code point sequences) could belong tho the same group.

Does sorting a table item result in the same call to `compare` without `compareHints`?
I guess not, because then it would have the same issue.

I think a better solution would be to try changing the cross tab item in such a way that it uses the same solution as the table item for sorting and grouping.

merks:
I don't expect a collator to treat two different characters as equals and in fact that isn't the case:

image

Anyway, my time is limited and I don't have domain knowledge.  I simple see results that scream wrong to me and this is where that comes from.

merks:
I don't see how to "inject" a cleaner solution into this call stack.   I had a hard enough time tracking this down and finally resorted to a conditional breakpoint on String.compare to find it, assuming that was being called.  I don't know which collator is being for the regular table. 

In the end,  I just see something that works poorly for all languages, always, everywhere.   Even the simplest collator will help with things like keeping 'A' and 'a' together in sorted order.   I don't disagree with your concerns, I just don't have knowledge of the big picture, and using String.compare as a default just screams out as a bad idea.

Anyway, I have done my best here with my limited time and knowledge.  If someone wants to fix this is a better way, that will need to be done by the someone who wants that.  No insult to criticism intended!

merks:
@speckyspooky 

Calling doCompare which casts each argument to CharSeuqnce and then calls toString is kind of strange.  We do know both arguments are String so can cast to String which appears would be a bit more efficient, so I changed that.

@hvbtup 

You're review is still a blocking review...

Issue #2338, opened 2025-12-05
Regarding the BIRT version 4.20 engine, there is an issue where the corresponding translation (localization) is not displayed.


LubinLuo:
Locally, the TW (Traditional Chinese/Taiwanese) translation type was selected for the language, but the displayed result is reading content from the zh_CN (Simplified Chinese/Mainland China) file, and not reading the content from the zh_TW file.
<img width="679" height="476" alt="Image" src="https://github.com/user-attachments/assets/fa557732-fc85-4a03-81a8-3d7b9fd3cdd9" />

PR #2353, opened 2025-12-23
Enhance the parameter locale handling (ICU4J) (#2338)

speckyspooky:


speckyspooky:
I have done your ogic and compared all deatiled information over 1.000 which are given and which are mssing.
The Locales are missing at both sides and there is not direct mapping way because some locales are named differently.

And if you have multiple matches like zh_TW (Hans vs. Hant) then a decision must be done.
Additional on it there are some Locales which are supported from standard locale named **...#Latn** which doesn't exists at ICU4J.
Additional on it there are further Locales which are supported from standard locale which contains **...#...** which doesn't exists at ICU4J and no direct (automated) mapping is mpossible.

Of course, I have done a complete compare with the mentioned logic above.

speckyspooky:
Hi Ed, do we need more documentation at code side or what would be the suggestion?

merks (approved):
I did a rebase to run the build again.  The change looks okay now.

PR #2356, opened 2026-01-03
Enhance rhino script handling of version 1.9.0

merks:
https://github.com/eclipse-birt/birt/issues/2354

### DIFF
--- a/core/org.eclipse.birt.core/src/org/eclipse/birt/core/script/JavascriptEvalUtil.java
+++ b/core/org.eclipse.birt.core/src/org/eclipse/birt/core/script/JavascriptEvalUtil.java
@@ -1,5 +1,5 @@
 /**************************************************************************
- * Copyright (c) 2005, 2024 Actuate Corporation and others
+ * Copyright (c) 2005, 2026 Actuate Corporation and others
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -229,6 +229,8 @@ public class JavascriptEvalUtil {
 			}
 		} else if (inputObj instanceof Wrapper) {
 			return ((Wrapper) inputObj).unwrap();
+		} else if (inputObj instanceof NativeArray) {
+			return ((NativeArray) inputObj).toArray();
 		} else if (inputObj instanceof Scriptable) {
 			return ((Scriptable) inputObj).getDefaultValue(null);
 		}
--- a/data/org.eclipse.birt.data.tests/test/org/eclipse/birt/data/engine/api/ScriptedDSTest.java
+++ b/data/org.eclipse.birt.data.tests/test/org/eclipse/birt/data/engine/api/ScriptedDSTest.java
@@ -490,12 +490,16 @@ public class ScriptedDSTest extends APITestCase {
 		init(false);
 		IQueryDefinition queryDefn = getReportQueryDefn(TEST_NONE);
 		try {
-			dsource.setBeforeOpenScript("importPackage(Packages.javax.swing.tree);");
-			dset.setBeforeOpenScript("dmtn = new DefaultMutableTreeNode();");
+			// dsource.setBeforeOpenScript("importPackage(Packages.javax.swing.tree);");
+			jsContext.evaluateString(jsScope, "importPackage(Packages.javax.swing.tree);", "testImportPackage", 1,
+					null);
+			// dset.setBeforeOpenScript("dmtn = new DefaultMutableTreeNode();");
+			jsContext.evaluateString(jsScope, "dmtn = new DefaultMutableTreeNode();", "testImportPackage", 2, null);
 			IPreparedQuery preparedQuery = dataEngine.prepare(queryDefn);
 			preparedQuery.execute(null);
 		} catch (Exception e) {
-			fail("Should not arrive here");
+			e.printStackTrace();
+			fail("Should not arrive here:" + e.getMessage());
 		}
 	}
 
--- a/data/org.eclipse.birt.data.tests/test/org/eclipse/birt/data/engine/binding/ColumnBindingTest.java
+++ b/data/org.eclipse.birt.data.tests/test/org/eclipse/birt/data/engine/binding/ColumnBindingTest.java
@@ -436,7 +436,7 @@ public class ColumnBindingTest extends APITestCase {
 
 		List combinedList = new ArrayList();
 		String script1 = "obj = new Array();obj[0]=\"Shanghai\"; obj[1]=\"Chicago\";obj";
-		String script2 = "obj = new Array();obj[0]= new Array(); obj[0][0]=\"London\"; obj[1]=\"Edinburgh\";obj";
+		String script2 = "obj = new Array();obj[0]= new Array(); obj[0]/*[0]*/=\"London\"; obj[1]=\"Edinburgh\";obj";
 		combinedList.add(script1);
 		combinedList.add(script2);
 		IConditionalExpression filter = new ConditionalExpression("row.testColumn2", IConditionalExpression.OP_IN,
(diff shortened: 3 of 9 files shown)


core/org.eclipse.birt.core/src/org/eclipse/birt/core/script/JavascriptEvalUtil.java
```diff
@@ -229,6 +229,8 @@ public static Object convertJavascriptValue(Object inputObj) {
 			}
 		} else if (inputObj instanceof Wrapper) {
 			return ((Wrapper) inputObj).unwrap();
+		} else if (inputObj instanceof NativeArray) {
+			return ((NativeArray) inputObj).toArray();
```
merks: @speckyspooky 

I traced the problem down locally to poor conversion handling for NativeArray which dropped into the next branch and returned a comma-separated string of the values which of course could not be processed by the called function.

speckyspooky: Tested with success.

merks:
Yes, I don't understand all the changes.  With the old logic it got the "real" scope

image

and ended up correctly here:

image

But with the new logic, none of that happens.  It pretty much seems like a bug in Rhino that it just assumes that `thisObject` is of a certain type when in fact the `scope` is the object on which it should be operating.  Note how in this deprecated method they throw away the `thisObject` and that the `this` in passed as the `thisObject`.  That's probably the basis of the assumption the code, but I don't think it's correct for what should behave like a global static function...

I'm not sure I can invest any more time tracking problems in such a complex framework...

hvbtup:
I don't know if this is helpful, but I think your example `dsource.setBeforeOpenScript("var javaImporter = new JavaImporter(); javaImporter.importPackage(Packages.javax.swing.tree);");` is not a realistic test case.

I'm using Java object from within JS in BIRT quite often, and a typical usage pattern for me is like this:

```
importPackage(Packages.java.util.logging);
var log = {
    _log: Logger.getLogger("birt.js"),
    debug: function(s) { this._log.fine(s); },
    info: function(s) { this._log.info(s); },
    warn: function(s) { this._log.warning(s); },
    warning: function(s) { this._log.warning(s); },
    error: function(args) {
    	this._log.severe("" + args);
    }
};
log._log.setLevel(Level.ALL);

...
log.info("executing script lisa.birt.js");

```

I mean, when I need Java classes inside any scripts, I always make sure to call `importPackage` before in the *same* script.
And why should a script explicitly use a new `JavaImporter` object when the function `importPackage` is available anyway?

merks:
According to the debugger, how BIRT actually uses Rhino 1.9 results in class cast exceptions whereas 1.8 works fine.  The debugger trumps expectations, theories and all AI agents because it reflects how the software really works running on an actual computer.  Note that even when one gets past the import and the use of the import, one gets stuck on `testPrint` not being visible.  Maybe it's all just weird test case artifacts.  Someone else will need to be the judge of that...

merks:
Yes, there is lots of questionable behavior that I see, such as creating more than one ImportTopLevel for the specific failing test; one created directly and one created when creating the engine context.

Staying on old versions of Rhino seems like just delaying the inevitable; eventually there's likely to be a CVE and then moving will become necessary....

Maybe these three failing tests are each just sort of bogus?

https://ci.eclipse.org/birt/job/build/job/PR-2356/1/testReport/

I'm perfectly happy to comment them out and move on... 

Things do appear to generally/typically be working well aside from these "outliers"...

merks:
> . it would be interesting to see where and how exactly the script execution fails

I’ve spent almost 2 days on that and shared my detailed findings.  

I can try to have a quick look at the regex thing.  But I think those who use this technology should feel some obligation to investigate in proportion to their usage and benefit. 

Please please someone needs to

data/org.eclipse.birt.data.tests/test/org/eclipse/birt/data/engine/api/ScriptedDSTest.java
```diff
@@ -490,12 +490,16 @@ public void testImportPackage() throws Exception {
 		init(false);
 		IQueryDefinition queryDefn = getReportQueryDefn(TEST_NONE);
 		try {
-			dsource.setBeforeOpenScript("importPackage(Packages.javax.swing.tree);");
-			dset.setBeforeOpenScript("dmtn = new DefaultMutableTreeNode();");
+			// dsource.setBeforeOpenScript("importPackage(Packages.javax.swing.tree);");
+			jsContext.evaluateString(jsScope, "importPackage(Packages.javax.swing.tree);", "testImportPackage", 1,
+					null);
+			// dset.setBeforeOpenScript("dmtn = new DefaultMutableTreeNode();");
+			jsContext.evaluateString(jsScope, "dmtn = new DefaultMutableTreeNode();", "testImportPackage", 2, null);
```
merks: Note that here I've change the logic to evaluate the two scripts on the global scope.  This reflects the change of behavior of importPackage relative to a global scope versus a local scope.  

I think the test is relatively bogus as @hvbtup has suggested, but it demonstrates that it's still possible to import classes and instantiate them which was the original intent of the test, I believe.

data/org.eclipse.birt.data.tests/test/org/eclipse/birt/data/engine/binding/ColumnBindingTest.java
```diff
@@ -436,7 +436,7 @@ public void testCombinedFilterOnTable2() throws Exception {
 
 		List combinedList = new ArrayList();
 		String script1 = "obj = new Array();obj[0]=\"Shanghai\"; obj[1]=\"Chicago\";obj";
-		String script2 = "obj = new Array();obj[0]= new Array(); obj[0][0]=\"London\"; obj[1]=\"Edinburgh\";obj";
+		String script2 = "obj = new Array();obj[0]= new Array(); obj[0]/*[0]*/=\"London\"; obj[1]=\"Edinburgh\";obj";
```
merks: I'm really not sure of the intent behind `obj[0][0]` but it looks bogus to me.  From other test failures it's clear that Rhino has done some changes to how array index assignment is implemented, i.e., the need to handle `NativeArray` happens only when array index assignments occur in the script.

speckyspooky: Here I would follow your test-change. I adjusted the test on my side too.

merks:
FYI, the GHA builds are failing for network/infrastructur reasons.
```
 Error:  dependency: org.osgi:osgi.annotation:jar:8.1.0 (compile)
Error:  	Could not transfer artifact org.osgi:osgi.annotation:jar:8.1.0 from/to central (https://repo.maven.apache.org/maven2): status code: 403, reason phrase: Forbidden (403)
```
The jar does exist:

- https://repo.maven.apache.org/maven2/org/osgi/osgi.annotation/8.1.0/osgi.annotation-8.1.0.jar

speckyspooky:
Retest of a small script for `importPackage`.
I executed 3 test cases:
- Case 01: execute an import package at dynamic text
- Case 02: move the import package to report-initialize()
- Case 03: move the import package to beforeFactory()
- Script:
```
importPackage(Packages.java.util);
var dmtn = new ArrayList();
dmtn.add(0, "test");
dmtn.get(0);
```
Result is positive for each test case. The report can be created with the correct output.
My hope was that latest with the `beforeFactory()` the same error will appears at report level.

grafik

So I think we have a topic like you mentioned it Ed with the global scope.
This problem would cause issue with scripted data sets.

data/org.eclipse.birt.data.tests/test/org/eclipse/birt/data/engine/binding/ColumnBindingTest.java
hvbtup: I don't understand the intent of the whole test.

Note: I'm using "array" and "list" as synonyms here, because it's about meaning, not about technical implementation.

I'm rarely using filters in BIRT, but for me, I would firmly expect an array of strings as the comparison value for the IN operator, like in SQL: `... where CITY IN ('London', 'Paris', 'Rome')`.
What should be the meaning of passing a list of two arrays in this regard?
It would only make sense if either the IN operator was supposed to operate on arbitrary objects (which would be surprising for me) or if the list of string arrays was automatically flattened to a list of strings (which I still would find surprising).
So, personally I think this whole test is - and always was - crap.

hvbtup:
> My hope was that latest with the `beforeFactory()` the same error will appears at report level.

Sorry, I don't understand what you mean.

merks:
There's now a new build with the results:

PR #2360, opened 2026-01-12
Migrate the BIRT viewer servlets from Apache Axis/javax to JDK 17 with Jakarta EE support

ischindl:
Hi all, this is my first pull request here. 
I was try migrate viewservlets to JDK 17 world. Without Axis and javax -> jakarta move. 

It is proof of concept, I dont know how to proceed with eclipse files like .classpath, .settings preferences in pull request 

It is working in our project where it is integrated into WAR

Main changes - SOAP is emulated over normal webservice, this mean no changes on javascript side.
Created docker for possibility run CBI aggregate locally 
Added possibility to deploy viewservlets.jar as  maven artifact  

Please comments all what is done bad or missing . I will try to fix it.
Thanks
Imrich

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -267,6 +267,11 @@
 							<arch>x86_64</arch>
 						</environment>
 						<environment>
+							<os>win32</os>
+							<ws>win32</ws>
+							<arch>aarch64</arch>
+						</environment>
+						<environment>
 							<os>linux</os>
 							<ws>gtk</ws>
 							<arch>x86_64</arch>
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,6 @@ concurrency:
 
 permissions:
   contents: read
-  actions: write
 
 jobs:
   build:
@@ -52,31 +51,6 @@ jobs:
     - name: Build with Maven
       run: mvn clean verify --no-transfer-progress -U -DskipTests=false
 
-    # Keep only the last run's artifacts: delete all previous artifacts starting with "birt-"
-    - name: Delete previous artifacts (keep only latest run)
-      uses: actions/github-script@v7
-      with:
-        script: |
-          const owner = context.repo.owner;
-          const repo = context.repo.repo;
-
-          const artifacts = await github.paginate(
-            github.rest.actions.listArtifactsForRepo,
-            { owner, repo, per_page: 100 }
-          );
-
-          const toDelete = artifacts.filter(a => (a.name || "").startsWith("birt-"));
-
-          core.info(`Found ${toDelete.length} artifacts starting with 'birt-' to delete.`);
-          for (const a of toDelete) {
-            core.info(`Deleting artifact id=${a.id}, name=${a.name}, created_at=${a.created_at}`);
-            await github.rest.actions.deleteArtifact({
-              owner,
-              repo,
-              artifact_id: a.id
-            });
-          }
-
     # Upload each output as its own artifact (so they appear individually in the UI)
     - name: Upload birt-report-designer linux aarch64
       uses: actions/upload-artifact@v4
@@ -85,7 +59,8 @@ jobs:
         if-no-files-found: error
         compression-level: 0
         path: build/birt-packages/birt-report-all-in-one/target/products/birt-report-designer-all-in-one-*-linux.gtk.aarch64.tar.gz
-
+        retention-days: 3
+        
     - name: Upload birt-report-designer linux x86_64
       uses: actions/upload-artifact@v4
       with:
@@ -93,6 +68,7 @@ jobs:
         if-no-files-found: error
         compression-level: 0
         path: build/birt-packages/birt-report-all-in-one/target/products/birt-report-designer-all-in-one-*-linux.gtk.x86_64.tar.gz
+        retention-days: 3
 
     - name: Upload birt-report-designer macos aarch64
       uses: actions/upload-artifact@v4
@@ -101,6 +77,7 @@ jobs:
         if-no-files-found: error
         compression-level: 0
         path: build/birt-packages/birt-report-all-in-one/target/products/birt-report-designer-all-in-one-*-macosx.cocoa.aarch64.tar.gz
+        retention-days: 3
 
     - name: Upload birt-report-designer macos x86_64
       uses: actions/upload-artifact@v4
@@ -109,6 +86,7 @@ jobs:
         if-no-files-found: error
         compression-level: 0
         path: build/birt-packages/birt-report-all-in-one/target/products/birt-report-designer-all-in-one-*-macosx.cocoa.x86_64.tar.gz
+        retention-days: 3
 
     - name: Upload birt-report-designer windows x86_64
       uses: actions/upload-artifact@v4
@@ -117,3 +95,13 @@ jobs:
         if-no-files-found: error
         compression-level: 0
         path: build/birt-packages/birt-report-all-in-one/target/products/birt-report-designer-all-in-one-*-win32.win32.x86_64.zip
+        retention-days: 3
+
+    - name: Upload birt-report-designer windows aarch64
+      uses: actions/upload-artifact@v4
+      with:
+        name: birt-report-designer-win32-aarch64
+        if-no-files-found: error
+        compression-level: 0
+        path: build/birt-packages/birt-report-all-in-one/target/products/birt-report-designer-all-in-one-*-win32.win32.aarch64.zip
+        retention-days: 3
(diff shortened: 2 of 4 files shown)


ischindl:
I was remove .classpath changes and squash changes into two commits

1.  primary javax-> jakarta and Axis remove changes
2.  mainly viewer changes to emulate SOAP

I register just now one bug - in viewer after change filter (action in toolbar) - servlet return old data from previous filter - may be invalid view session processing.

merks:
@speckyspooky @wimjongman @hvbtup @SteveSchafer-Innovent 

Do you folks have thoughts or comments?  I'm mostly looking at this from a releng point of view...

@ischindl 

FYI, note that the *.target is actually generated when using the setup:

image

Also the aggregor tool is installed by the setup such that one can generate the maven repository in the IDE.

---

I fetched the GitHub PR into my IDE, did Help -> Perform Setup tasks to update the target platform (which uses Targlet which effectively resolve the dependencies on the bundles in the git repository so adjust to the changes you made).  Then I launched self-host launch where things worked quite nicely except the preview server no longer works:

image

Other than this, which is kind of essential, things look good. 

I have no further time before leaving the office...

hvbtup:
I like the idea, but I probably cannot help here, lacking time and expertise.

wimjongman (commented):
Thank you Imrich!! I did a first pass.

merks:
As far as I recall, the javax/jakarta dependencies are not singletons and should be able to coexist in general.   Also WST and JST are not dependencies of the viewer servlet.  Equinox's JSP might be problematic.  I'm not sure about that part.  You should only need to concern yourself with the servlet viewer dependencies:

image

Supporting multiple versions of the servlet should be avoided...  It's all already too complex...

ischindl:
image
First partial success.

Now minimal requirement which i need in orbit is 

<id>org.glassfish.expressly:expressly:5.0.0</id>

ischindl:
With second one 

org.glassfish.jaxb:jaxb-runtime:4.0.1

The report appear.

image

hvbtup:
Are these errors "java.lang.ClassNotFoundException: org.eclipse.birt.report.data.oda.sampledb.Driver" in the failed CI build expected?

hvbtup:
> - it is only referenced in 3 rptdesigns

Yeah, this is what is used for the Classic Models builtin sample DB.
I don't see similar ClassNotFoundException for this class in a build.log of one of my old builds.

Anyway, please make sure that reports using the Class Models sample DB still work.

ischindl:
Ok core trouble is move java11 -> java21 and classpath for tests. I must fix this.

ischindl:
I had to do some changes into javax.olap, into runtime test classloader to be compatible with java21, but OSGI test is now ignored because of java21 classloaders separation and plurl in osgi. I will try to find out way how to run it again.

ischindl:
> FYI, with regard to updating org.eclipse.jetty.servlet-api to 4.0.8 there is this problem:
> 
> [eclipse-platform/eclipse.platform.releng.aggregator#3615 (comment)](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3615#issuecomment-3764047962)
> 
> Even my local BIRT builds of master fail with compile errors that look like the ones from the Platform. I tried changing the package import to a bundle import but even then some tests fail so it all looks suspicious. I'm considering if I should roll back and exclude this version from Orbit because it's kind of broken it would seem...
> 
> Oh, and I think none of this really matters with your changes because it's changing this to jakarta servlet versions.

I added workaround into https://github.com/eclipse-birt/birt/pull/2364

merks:
> And this is trouble for OSGI - PLURL - it is initilalized in static part of code and this failed on java21 to get java.net.URL dependency.

I'm confused though because the PURL implementation is designed expressly for the purpose of the replacing the old implementation that relied on forbidden deep reflection.  

I believe the test was previously running with Java 21 before any changes from this PR via this commit upgraded which everything to Java 21:

https://github.com/eclipse-birt/birt/commit/2c3f1926e83f648e9b2d48c1604f8e238433694e

merks:
@ischindl 

I'm back from vacation so could spend a  bit more time reviewing/testing changes.  But things appears to be stalled for quite a while now.  If you want to continue, please squash and rebase the PR on master.

wimjongman:
> > Is this a script for building a BIRT docker image? If you can put that part in a separate PR then we can track it better.
> 
> This docker is only for CBI call to build Maven repo from builded P2 repo. It call same stuff as is used on Jenkins. It is here [build/org.eclipse.birt.releng.maven/docker](https://github.com/eclipse-birt/birt/pull/2360/changes#diff-9063a0b5828bd598197e451b97223b38c62dd455fb30a414003fb6f9a404fd75)
> 
> From my point of view, it is only helper and not required tool.

@merks WDYT?

viewer/org.eclipse.birt.report.viewer.tests/pom.xml
```diff
@@ -23,4 +26,18 @@
...
+        <version>4.0.0</version>
+			<scope>test</scope>
+    </dependency>
+		<dependency>
+			<groupId>org.glassfish.jaxb</groupId>
+			<artifactId>jaxb-runtime</artifactId>
+			<version>4.0.5</version>
+			<scope>test</scope>
+		</dependency>
+	</dependencies>
```
merks: Please specify these dependencies in the MANIFEST.MF instead.  It's already difficult to maintain dependencies and having these specified in the pom and them becoming outdated in the future is just asking for strange problems in the future.

pom.xml
```diff
@@ -37,7 +37,7 @@
 		<build.helper.maven.plugin.version>3.6.0</build.helper.maven.plugin.version>
 		<cbi-version>1.5.2</cbi-version>
 		<jacoco.maven.plugin.version>0.8.12</jacoco.maven.plugin.version>
-		<jetty-jspc-maven-plugin-version>10.0.24</jetty-jspc-maven-plugin-version>
+		<jetty-jspc-maven-plugin-version>12.1.5</jetty-jspc-maven-plugin-version>
```
merks: I think 12.1.6 has been released so maybe better to use that?

ischindl: I mean help is only missing package in target , I will try.

wimjongman: > Failing all that, I could just merge and them make the 3 changes.

+1 Let's go. We can fix issues when we find them. it will be much easier. If all fails we can always roll back.

merks:
It works for me locally:

image

ischindl:
Ok i can make rebase and force push

wimjongman:
> BTW, squash and merge is also not actually nice because it changes the author of the commit to be some bot thing. 

I squashed and merged 4 commits today. The author is preserved normally. 

image

> As I said, I'm a horrible fussy person.

I know😂. I don't mind at all.

wimjongman:
Thank you Imrich (@ischindl) and Ed @merks for getting this in! What a great contribution.  🎉🎉🎉

merks:
@akurtakov  @HannesWell @laeubi 

FYI, here is concrete a problem for folks trying to use the help system and use newer versions of Jakarta, Jetty, JSP and such.

In principle such things should be able to co-exist, but in practice it appears dependencies are not sufficient constrained in 3rd party bundles.

It makes me wonder if such constraints could be specified/imposed in the help system bundles themselves?  E.g., here:

https://github.com/eclipse-platform/eclipse.platform/blob/master/ua/org.eclipse.help.webapp/META-INF/MANIFEST.MF

Of course in the ideal world, there would be sufficient resource to update the help system to Jakarta and company...

@ischindl 

Thanks for investigating.

Everything in this repo:

https://download.eclipse.org/tools/orbit/simrel/orbit-aggregation/table.html

comes via this *.target:

https://github.com/eclipse-orbit/orbit-simrel/blob/main/maven-osgi/tp/Maven.target

and correspond to jars that are available in Maven Central as OSGi bundles.   

We cannot modify existing OSGi metadata; the project providing the bundle must do that.  For that we'd need to open an issue for the project and hope for the best...

ischindl:
i created https://github.com/eclipse-birt/birt/issues/2372

merks:
@ischindl 

I was wondering about tomcat compatibility?

We have documented (at the very bottom of the page) that tomcat 10 and 11 are not supported due to Jakarta EE dependencies:

https://github.com/eclipse-birt/birt?tab=readme-ov-file#runtime-requirements

But I wonder if that still true for BIRT 4.23 with your awesome upgrades?

@speckyspooky 

This is something you will probably want to try out...

merks:
I agree EE 10 is a big improvement and I don't see any super compelling reason for EE 11.  But hey, I'm just the release engineer.

@AndreasIgelCC 

I don't think there are migration steps.   The runtime(s) simply contain the necessary things.  But then I don't know what code you have that might need changes.   You might consider to help provide a guide when you migrate whatever it is you are migrating.

PR #2408, opened 2026-05-02
Enforce urlReportPathPolicy in ParameterAccessor.getResourceFolder

merks:
- In particular check that the value of __resourceFolder in the request, if present, passes the ParameterAccessor.isValidFilePath test just as is already required for the report itself.

### DIFF
--- a/viewer/org.eclipse.birt.report.viewer/birt/WEB-INF/classes/org/eclipse/birt/report/utility/ParameterAccessor.java
+++ b/viewer/org.eclipse.birt.report.viewer/birt/WEB-INF/classes/org/eclipse/birt/report/utility/ParameterAccessor.java
@@ -39,11 +39,6 @@ import java.util.TimeZone;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-import jakarta.servlet.ServletConfig;
-import jakarta.servlet.ServletContext;
-import jakarta.servlet.http.HttpServletRequest;
-import jakarta.servlet.http.HttpSession;
-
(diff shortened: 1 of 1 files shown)


merks:
For a  URL such as this:

http://127.0.0.1:61819/viewer/frameset?__report=D%3A\Users\merks\birt-master\runtime-workspace\project\hack.rptdesign&__format=html&__svg=true&__locale=en_US&__timezone=ECT&__masterpage=true&__rtl=false&__cubememsize=10&__resourceFolder=http://localhost:8080/user.home/&-254663951

with the guard in place, the report fails like this:

image

Without the guard, and with this instrumented

image

we can see the relative location before resolved like this:

```
###library.rptlibrary->http://localhost:8080/user.home/library.rptlibrary
###library.rptlibrary->http://localhost:8080/user.home/library.rptlibrary
###exploit.js->http://localhost:8080/user.home/exploit.js
###templates/TOCDefaultValue.xml->bundleresource://334.fwk1755588557/templates/TOCDefaultValue.xml
###exploit.js->http://localhost:8080/user.home/exploit.js
```

In general, different policies can be enforced

https://github.com/eclipse-birt/birt/blob/25b31134a74d9210fcb7952c155e92eb985ee6e1/viewer/org.eclipse.birt.report.viewer/birt/WEB-INF/classes/org/eclipse/birt/report/utility/ParameterAccessor.java#L1932-L1988

merks:
I'll roll this out to eliminate the problem.  This could be refined later if need be.

PR #2409, opened 2026-05-08
Implements recognition of SVG images based on the image content.

webwart-bln:
Instead on relying on mime-type or file-ending it is now possible to embed SVG images as PDF images coming from a database blob or similar sources.

### DIFF
--- a/engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPage.java
+++ b/engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPage.java
@@ -185,9 +185,14 @@ public class PDFPage extends AbstractPage {
 
 			// SVG images
 			try {
-				if (this.pageDevice.useBackgroundImageSvg() && SvgFile.isSvg(null, null, imageUrl)
-						&& (new File((new URL(imageUrl)).toURI().getPath())).exists()) {
-					image = transSVG(imageUrl, null, imageHeight, imageWidth);
+				boolean isSvgImage = this.pageDevice.useBackgroundImageSvg()
+						&& SvgFile.isSvg(null, imageUrl, null, imageData);
+				if (isSvgImage) {
+					if ((new File((new URL(imageUrl)).toURI().getPath())).exists()) {
+						image = transSVG(imageUrl, null, imageHeight, imageWidth);
+					} else if (imageData != null) {
+						image = transSVG(null, imageData, imageHeight, imageWidth);
+					}
 					if (image != null) {
 						drawImage(image, x, y, imageHeight, imageWidth, null);
 					}
@@ -283,7 +288,7 @@ public class PDFPage extends AbstractPage {
 		}
 
 		// Not cached yet
-		if (SvgFile.isSvg(null, null, extension)) {
+		if (SvgFile.isSvg(null, null, extension, imageData)) {
 			template = generateTemplateFromSVG(imageData, height, width);
 		} else {
 			// PNG/JPG/BMP... images:
--- a/engine/org.eclipse.birt.report.engine.tests/test/org/eclipse/birt/report/engine/emitter/pdf/PDFSvgEmbeddingTest.java
+++ b/engine/org.eclipse.birt.report.engine.tests/test/org/eclipse/birt/report/engine/emitter/pdf/PDFSvgEmbeddingTest.java
@@ -1,0 +1,190 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * https://www.eclipse.org/legal/epl-2.0/.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   See git history
+ *******************************************************************************/
+
+package org.eclipse.birt.report.engine.emitter.pdf;
+
+import java.io.ByteArrayOutputStream;
+import java.io.ByteArrayInputStream;
+import java.nio.charset.StandardCharsets;
+
+import org.apache.batik.transcoder.SVGAbstractTranscoder;
+import org.apache.batik.transcoder.TranscoderInput;
+import org.apache.batik.transcoder.print.PrintTranscoder;
+import org.openpdf.text.Document;
+import org.openpdf.text.PageSize;
+import org.openpdf.text.pdf.PdfContentByte;
+import org.openpdf.text.pdf.PdfTemplate;
+import org.openpdf.text.pdf.PdfWriter;
+
+import junit.framework.TestCase;
+
+public class PDFSvgEmbeddingTest extends TestCase {
+
+	private static final String SIMPLE_SVG =
+			"<svg xmlns=\"http://www.w3.org/2000/svg\" width=\"100\" height=\"100\">" +
+			"<rect x=\"10\" y=\"10\" width=\"80\" height=\"80\" fill=\"red\"/>" +
+			"</svg>";
+
+	private static final String SVG_WITH_PATH =
+			"<svg xmlns=\"http://www.w3.org/2000/svg\" width=\"200\" height=\"200\">" +
+			"<path d=\"M10,10 L190,10 L190,190 L10,190 Z\" fill=\"blue\" stroke=\"black\" stroke-width=\"2\"/>" +
+			"</svg>";
+
+	private static final String SVG_WITH_CIRCLE =
+			"<svg xmlns=\"http://www.w3.org/2000/svg\" width=\"50\" height=\"50\">" +
+			"<circle cx=\"25\" cy=\"25\" r=\"20\" fill=\"green\"/>" +
+			"</svg>";
+
+	private static final String SVG_WITH_TEXT =
+			"<svg xmlns=\"http://www.w3.org/2000/svg\" width=\"300\" height=\"50\">" +
+			"<text x=\"10\" y=\"30\" font-family=\"Arial\" font-size=\"20\">Hello World</text>" +
+			"</svg>";
+
+	public void testTranscodeSvgByteArrayToPdfTemplate() throws Exception {
+		byte[] svgData = SIMPLE_SVG.getBytes(StandardCharsets.UTF_8);
+		float width = 100, height = 100;
+
+		PdfTemplate template = transSvgToTemplate(svgData, width, height);
+
+		assertNotNull("PDF template must not be null", template);
+		assertTrue("Template width must be positive", template.getWidth() > 0);
+		assertTrue("Template height must be positive", template.getHeight() > 0);
+	}
+
+	public void testTranscodeSvgWithPath() throws Exception {
+		byte[] svgData = SVG_WITH_PATH.getBytes(StandardCharsets.UTF_8);
+		float width = 200, height = 200;
+
+		PdfTemplate template = transSvgToTemplate(svgData, width, height);
+
+		assertNotNull("PDF template must not be null", template);
+		assertEquals("Template width must match", width, template.getWidth());
+		assertEquals("Template height must match", height, template.getHeight());
+	}
+
+	public void testTranscodeSvgWithCircle() throws Exception {
+		byte[] svgData = SVG_WITH_CIRCLE.getBytes(StandardCharsets.UTF_8);
+		PdfTemplate template = transSvgToTemplate(svgData, 50, 50);
+		assertNotNull("PDF template for circle SVG must not be null", template);
+	}
+
+	public void testTranscodeSvgWithText() throws Exception {
+		byte[] svgData = SVG_WITH_TEXT.getBytes(StandardCharsets.UTF_8);
+		PdfTemplate template = transSvgToTemplate(svgData, 300, 50);
+		assertNotNull("PDF template for text SVG must not be null", template);
+	}
+
+	public void testGeneratedPdfIsReadableAndContainsSvgContent() throws Exception {
+		byte[] svgData = SVG_WITH_PATH.getBytes(StandardCharsets.UTF_8);
+
+		Document document = new Document(PageSize.A4);
+		ByteArrayOutputStream pdfOut = new ByteArrayOutputStream();
+		PdfWriter writer = PdfWriter.getInstance(document, pdfOut);
+
+		document.open();
+
+		PdfContentByte cb = writer.getDirectContent();
+		PdfTemplate svgTemplate = transSvgToTemplate(svgData, 200, 200);
+
+		assertNotNull("SVG template must not be null", svgTemplate);
+		cb.addTemplate(svgTemplate, 50, 500);
+
+		document.newPage();
+		document.close();
+
+		byte[] pdfBytes = pdfOut.toByteArray();
+		assertTrue("PDF must be non-empty", pdfBytes.length > 0);
+		assertTrue("PDF must start with %PDF-",
+				new String(pdfBytes, 0, Math.min(pdfBytes.length, 5), StandardCharsets.ISO_8859_1).startsWith("%PDF-"));
+	}
+
+	public void testSvgEmbeddedAsVectorNotRasterized() throws Exception {
+		byte[] svgData = SVG_WITH_PATH.getBytes(StandardCharsets.UTF_8);
+
+		Document document = new Document(PageSize.A4);
+		ByteArrayOutputStream pdfOut = new ByteArrayOutputStream();
+		PdfWriter writer = PdfWriter.getInstance(document, pdfOut);
+
+		document.open();
+
+		PdfContentByte cb = writer.getDirectContent();
+		PdfTemplate svgTemplate = transSvgToTemplate(svgData, 200, 200);
+		cb.addTemplate(svgTemplate, 50, 500);
+
+		document.close();
+
+		byte[] pdfBytes = pdfOut.toByteArray();
+		String pdfContent = new String(pdfBytes, StandardCharsets.ISO_8859_1);
+
+		assertTrue("PDF must contain vector path commands",
+				pdfContent.contains(" m ") || pdfContent.contains(" l "));
+
+		assertFalse("PDF must NOT contain JPEG raster stream (DCTDecode)",
+				pdfContent.contains("/DCTDecode"));
+	}
+
+	public void testSvgWithNamespaceEmbeddedCorrectly() throws Exception {
+		String svgWithNs = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>" +
+				"<svg xmlns=\"http://www.w3.org/2000/svg\" xmlns:xlink=\"http://www.w3.org/1999/xlink\" " +
+				"width=\"100\" height=\"100\">" +
+				"<rect x=\"0\" y=\"0\" width=\"100\" height=\"100\" fill=\"yellow\"/>" +
+				"</svg>";
+		byte[] svgData = svgWithNs.getBytes(StandardCharsets.UTF_8);
+
+		PdfTemplate template = transSvgToTemplate(svgData, 100, 100);
+		assertNotNull("SVG with namespace must transcode to non-null template", template);
+	}
+
+	public void testEmptySvgDataThrowsException() throws Exception {
+		try {
+			transSvgToTemplate(new byte[0], 100, 100);
+			fail("Empty SVG data should cause an exception");
+		} catch (Exception e) {
+		}
+	}
+
+	public void testNullSvgDataDoesNotFail() throws Exception {
+		PdfTemplate template = transSvgToTemplate(null, 100, 100);
+		assertNotNull("Null SVG data should still produce a template (empty)", template);
+	}
+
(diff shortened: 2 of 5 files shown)


merks:
The build error suggests there need package imports of org.apache.batik.transcoder and org.apache.batik.transcoder.print  in MANIFEST.MF of the test bundle where you added the test.

```
09:37:44  [ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:5.0.2:compile (default-compile) on project org.eclipse.birt.report.engine.tests: Compilation failure: Compilation failure: 
09:37:44  [ERROR] /home/jenkins/agent/workspace/build_PR-2409/engine/org.eclipse.birt.report.engine.tests/test/org/eclipse/birt/report/engine/emitter/pdf/PDFSvgEmbeddingTest.java:[7] 
09:37:44  [ERROR] 	import org.apache.batik.transcoder.SVGAbstractTranscoder;
09:37:44  [ERROR] 	       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
09:37:44  [ERROR] The import org.apache.batik.transcoder cannot be resolved
09:37:44  [ERROR] /home/jenkins/agent/workspace/build_PR-2409/engine/org.eclipse.birt.report.engine.tests/test/org/eclipse/birt/report/engine/emitter/pdf/PDFSvgEmbeddingTest.java:[8] 
09:37:44  [ERROR] 	import org.apache.batik.transcoder.TranscoderInput;
09:37:44  [ERROR] 	       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
09:37:44  [ERROR] The import org.apache.batik.transcoder cannot be resolved
09:37:44  [ERROR] /home/jenkins/agent/workspace/build_PR-2409/engine/org.eclipse.birt.report.engine.tests/test/org/eclipse/birt/report/engine/emitter/pdf/PDFSvgEmbeddingTest.java:[9] 
09:37:44  [ERROR] 	import org.apache.batik.transcoder.print.PrintTranscoder;
09:37:44  [ERROR] 	       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
09:37:44  [ERROR] The import org.apache.batik.transcoder cannot be resolved
09:37:44  [ERROR] /home/jenkins/agent/workspace/build_PR-2409/engine/org.eclipse.birt.report.engine.tests/test/org/eclipse/birt/report/engine/emitter/pdf/PDFSvgEmbeddingTest.java:[159] 
09:37:44  [ERROR] 	PrintTranscoder transcoder = new PrintTranscoder();
09:37:44  [ERROR] 	^^^^^^^^^^^^^^^
09:37:44  [ERROR] PrintTranscoder cannot be resolved to a type
09:37:44  [ERROR] /home/jenkins/agent/workspace/build_PR-2409/engine/org.eclipse.birt.report.engine.tests/test/org/eclipse/birt/report/engine/emitter/pdf/PDFSvgEmbeddingTest.java:[159] 
09:37:44  [ERROR] 	PrintTranscoder transcoder = new PrintTranscoder();
09:37:44  [ERROR] 	                                 ^^^^^^^^^^^^^^^
09:37:44  [ERROR] PrintTranscoder cannot be resolved to a type
09:37:44  [ERROR] /home/jenkins/agent/workspace/build_PR-2409/engine/org.eclipse.birt.report.engine.tests/test/org/eclipse/birt/report/engine/emitter/pdf/PDFSvgEmbeddingTest.java:[161] 
09:37:44  [ERROR] 	transcoder.transcode(new TranscoderInput(new ByteArrayInputStream(svgData)), null);
09:37:44  [ERROR] 	                         ^^^^^^^^^^^^^^^
09:37:44  [ERROR] TranscoderInput cannot be resolved to a type
09:37:44  [ERROR] /home/jenkins/agent/workspace/build_PR-2409/engine/org.eclipse.birt.report.engine.tests/test/org/eclipse/birt/report/engine/emitter/pdf/PDFSvgEmbeddingTest.java:[163] 
09:37:44  [ERROR] 	transcoder.addTranscodingHint(SVGAbstractTranscoder.KEY_ALLOW_EXTERNAL_RESOURCES, Boolean.TRUE);
09:37:44  [ERROR] 	                              ^^^^^^^^^^^^^^^^^^^^^
09:37:44  [ERROR] SVGAbstractTranscoder cannot be resolved to a variable
09:37:44  [ERROR] 7 problems (7 errors)
```

webwart-bln:
You are so right. I updated the file and mvn package (including tests) is running locally without any errors. I hope that it works now as expected.

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPage.java
```diff
@@ -185,9 +185,14 @@ protected void drawBackgroundImage(float x, float y, float width, float height,
 
 			// SVG images
 			try {
-				if (this.pageDevice.useBackgroundImageSvg() && SvgFile.isSvg(null, null, imageUrl)
-						&& (new File((new URL(imageUrl)).toURI().getPath())).exists()) {
-					image = transSVG(imageUrl, null, imageHeight, imageWidth);
+				boolean isSvgImage = this.pageDevice.useBackgroundImageSvg()
+						&& SvgFile.isSvg(null, imageUrl, null, imageData);
```
merks, line 189: I was a little concerned that the second and third arguments are exchanged, but I see that's mostly okay, though has a subtle impact on testing for contains `URL_IMAGE_TYPE_SVG`.  But that all looks to be slightly confusing and bogus before anything was modified.

The use of a static variable for `org.eclipse.birt.report.engine.util.SvgFile.isSvg` just looks so wrong and silly.  Not something you clanged, but screams, "Let's find a way to make the class not thread safe for completely gratuitous reasons."  While you are improving this class, could I ask that you please also delete that static variable?

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/util/SvgFile.java
```diff
@@ -65,15 +71,53 @@ public static boolean isSvg(String uri) {
...
 		return isSvg;
 	}
 
+	/**
+	 * Validation of SVG image by content inspection
+	 *
+	 * @param data raw image data (bytes)
+	 * @return true if the content looks like an SVG/XML document
+	 */
+	public static boolean isSvgContent(byte[] data) {
```
merks: Maybe we should make this private to avoid yet more API?

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/util/SvgFile.java
```diff
@@ -65,15 +71,53 @@ public static boolean isSvg(String uri) {
...
+		if (data == null || data.length < 5) {
+			return false;
+		}
+		String content;
+		try {
+			content = new String(data, 0, Math.min(data.length, 512), StandardCharsets.UTF_8);
```
merks: As I understand it, this never throws an exception because the bytes aren't properly encoded UTF-8 but rather always returns something, which might be a completely butchered something.

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/util/SvgFile.java
```diff
@@ -65,15 +71,53 @@ public static boolean isSvg(String uri) {
...
+		} catch (Exception e) {
+			try {
+				content = new String(data, 0, Math.min(data.length, 512), StandardCharsets.ISO_8859_1);
```
merks: I don't think you ever get here for the reason you think it might get here.  Also, given the content you are looking at is mostly just the XML itself and for SVG those are ASCII characters, this sis probably mot needed because ISO latin and UTF-8 overlap in the ASCII range.

What's more likely to be problematic is if someone uses a double byte encoding, e.g., UTF-16.  So this seems not very robust.  (And of course the encoding is specified in the XML header.)

merks (changes requested):
The changes look generally sensible.

Looking at where other isSvg methods are called, one really has to wonder what a kind of  mess this this?  (Not your changes but what exists):

Image

These two places would likely be candidates for your new method:

Image

I would very much prefer that you rebase to a single commit and force push amended commits to maintain the single commit state of the PR.

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/util/SvgFile.java
```diff
@@ -65,15 +71,53 @@ public static boolean isSvg(String uri) {
...
+
+	/**
+	 * Validation of SVG image
+	 *
+	 * @param mimeType  mime type to be validated
+	 * @param uri       URI to be validated
+	 * @param extension extension to be validated
+	 * @param data      raw image data (bytes) for content-based detection
+	 * @return the validation result of SVG image
+	 */
```
merks, line 85: Oh, I forgot to mention there should be an @ since 4.24.

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/util/SvgFile.java
webwart-bln: absolutetly. good thing pointing that out.

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/util/SvgFile.java
webwart-bln: I agree that trying with this fallback is unnecessary. 

I wanted to avoid to try to parse the while byte[] into an xml document (for performance reasons and not having to handle XXE and other possible XML-base attack vectors). I think that having a double byte encoded SVG is allowed but very unlikely because it might cause trouble in other software. 

Hmm. I think I could let Batik try to determine whether the byte{] is a SVG file it can handle like this:

```
private boolean isSvg(byte[] data) {
    try {
        String parser = XMLResourceDescriptor.getXMLParserClassName();
        SAXSVGDocumentFactory factory = new SAXSVGDocumentFactory(parser);
        InputStream inputStream = new ByteArrayInputStream(data);
        Document doc = factory.createDocument(null, inputStream);
        return doc != null;
    } catch (Exception e) {
       return false;
    }
} 


```But I am not sure whether the used XML parser and document instances are safe. What would you suggest?

File: engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/util/SvgFile.java (hunk 5 above)
Comment (webwart-bln) [reply engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/util/SvgFile.java:85]: I will add that!

File: engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPage.java (hunk 1 above)
Comment (webwart-bln) [reply engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPage.java:189]: Sure, I will look into removing that. Do you want to have the  order of the arguments changed back to the original order?

Comment (merks) [reply engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPage.java:189]: No t think was kind of wrong before.

File: engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/util/SvgFile.java (hunk 4 above)
Comment (merks) [reply engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/util/SvgFile.java]: Indeed that's quite heavy weight to build a document for the hole document.  

Looking at how EMF and the Platform do content type recognition, one might instead create a SAX parser, e.g,,
```
	private static synchronized SAXParserFactory createSAXFactoryIgnoringDOCTYPE() {
		SAXParserFactory f = SAXParserFactory.newInstance();
		try {
			f.setFeature("http://xml.org/sax/features/external-general-entities", false); //$NON-NLS-1$
			f.setFeature("http://xml.org/sax/features/external-parameter-entities", false); //$NON-NLS-1$
			f.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false); //$NON-NLS-1$
		} catch (Exception e) {
			throw new RuntimeException(e);
		}
		return f;
	}
```
and then do a SAX parse with an extended org.xml.sax.helpers.DefaultHandler where you simply abort the parsing process as soon as you get to the initial start element at which point you know the namespace and local name of that element.  E.g., sort of like this that we do in EMF (where EMF is much more complex than what you would do for determining if the uri/localName are SVG.
```
                    public void startElement(String uri, String localName, String name, Attributes attributes) throws SAXException
                    {
                      super.startElement(uri, localName, name, attributes);
                      int depth = elements.size();
                      if (depth == 1 && !isXMINameAndNamespace(localName, uri) ||
                            depth == 2 && !isXMINamespace(uri))
                      {
                        endElement(uri, localName, name);
                        if (depth == 2)
                        {
                          endElement(uri, localName, name);
                        }
                        endDocument();
                        throw new RuntimeException();
                      }
                    }
```

The general concern is of course UTF-8, big- and little-endian UTF-16, Byte Order Marks and all the other XML complexity all of which are handled by using a deeply-XML-aware mechanism...

I think this could be made relatively simple and then also be extremely general.

Comment (hvbtup) [reply engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/util/SvgFile.java]: I agree that UTF-16 or UTF-32 are very unlikely, at least unlikely enough to just not support auto-detection in that case. I would use ASCII decoding with undecodable bytes just replaced, because what we want to know here is just if it's SVG content or not.

Comment (hvbtup) [reply engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/util/SvgFile.java]: A regex should suffice, creating a parser is using a sledgehammer to crack a nut.

Comment (merks) [reply engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/util/SvgFile.java]: Unfortunately the nut has many nuances.  Here's an SVG from the Eclipse project:

image

Does that (a comment) work with the pattern?  I suspect not.   What else doesn't work?  It's quite some list.  So yes, it's a big hammer but also a tough nut.  Given that it's really easy to use .svg suffix, so once one get past that really cheap nut cracker, maybe one really does want a more thorough hammer?

Comment (hvbtup) [reply engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/util/SvgFile.java]: Where?

Comment (merks) [reply engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/util/SvgFile.java]: By "Where?" asking where is the pattern? 

This one appears to allow a `DOCTYPE` but not a comment.
```
	private static final Pattern SVG_CONTENT_PATTERN = Pattern.compile(
			"^\\s*(?:<\\?xml[^?]*\\?>\\s*)?(?:<!DOCTYPE[^>]*>\\s*)?<svg[\\s>]",
			Pattern.CASE_INSENSITIVE);
```

I'm really not hell-bent on a sledge hammer, but if we are going to hammer this nut to inspect the contents inside the shell,  we do have to ask ourselves to what purpose and which compromises are we willing to make?  A comment seems common:

image

My generally approach is to consider carefully every dark corner rather than to look primarily under the streetlight where visibility is best.  So even if we do add support for a comment, the 512 characters is getting to be maybe a bit short...

Comment (hvbtup) [reply engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/util/SvgFile.java]: > By "Where?" asking where is the pattern?

For whatever reason I could not see the image in your comment.

What I do in other projects is this:
First, scan the first 500 *bytes* for the <?xml encoding="..." (case-independent, replace non-ASCII characters with dummy values) to detect the encoding. Then, use this encoding to translate these bytes (after stripping non-ASCII bytes at the end) to a string. Then perform further string-based checks.

OTOH BIRT already includes an XML parser, so why not use it...

Comment (webwart-bln) [reply engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/util/SvgFile.java]: I am happy to implement the check based on XML parsing, if you think that this is the safest option.

> OTOH BIRT already includes an XML parser, so why not use it...

I found that one could use org.eclipse.birt.core.util.CommonUtil#createSAXParser();

With this i could do something like this:

```
private static boolean isValidSvg(byte[] data) {
    try {
        SAXParser saxParser = CommonUtil.createSAXParser();
        SvgHandler handler = new SvgHandler();

        saxParser.parse(new InputSource(new ByteArrayInputStream(data)), handler);
        return true;
    } catch (Exception e) {
        return false;
    }
}

private static class SvgHandler extends DefaultHandler {

    private boolean svgTagChecked = false;

    @Override
    public void startElement(String uri, String localName, String qName, Attributes attributes) {
        if (!svgTagChecked && qName.equalsIgnoreCase("svg")) {
            // this is the "normal" case: first real tag of the document is a svg tag
            svgTagChecked = true;
        } else {
            // fail fast!
            throw new IllegalStateException("first element is not a svg tag!");        
        }
    }
}

```

What do you think? is this the way to go?

Comment (merks) [reply engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/util/SvgFile.java]: Yes, that's a **great idea**!  CommonUtil already does the appropriate things to avoid attack vectors!

Just keep in mind that it's better if `startElement` always throws an exception so that you abort all processing as soon as you see the root element.  That will perform much better than letting it handle the whole document.

File: engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/util/SvgFile.java (hunk 2 above)
Comment (webwart-bln) [reply engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/util/SvgFile.java]: I made it private in the current version of this PR.

File: engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/util/SvgFile.java (hunk 4 above)
Comment (webwart-bln) [reply engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/util/SvgFile.java]: I implemented a new version of the recognition by using a SAXParser. Unfortunately I could not use the parser provided by CommonUtil, since it forbids having a DOCTYPE declaration, which is totally fine in SVGs, I think. So I added a new method to create a usable SAXParser, which should be as secure as possible.

I used the recommendations from https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html#jaxp-documentbuilderfactory-saxparserfactory-and-dom4j and https://xmltools.github.io/xml-security-essentials 

I hope that's okay that way.

File: engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/util/SvgFile.java (hunk 5 above)
Comment (webwart-bln) [reply engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/util/SvgFile.java:85]: Done that.

File: engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/util/SvgFile.java [hunk 6]
```diff
@@ -49,11 +59,10 @@ public class SvgFile {
 	 */
 	public static boolean isSvg(String uri) {
 		if (uri != null && (uri.endsWith(FILE_EXTENSION_SVG) || uri.toLowerCase().contains(URL_IMAGE_TYPE_SVG))) {
-			isSvg = true;
+			return true;
```
merks: You could just have a single return statement.

merks (changes requested):
Please add copyright headers to the new files.  Look at other classes for examples.

Otherwise it mostly fine on other one little nitty thing...

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/util/SvgFile.java
webwart-bln: I changed it to return directly

webwart-bln:
I added the missing copyright headers in my latest changes.

engine/org.eclipse.birt.report.engine.tests/test/org/eclipse/birt/report/engine/emitter/pdf/PDFSvgEmbeddingTest.java
```diff
@@ -0,0 +1,189 @@
...
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * https://www.eclipse.org/legal/epl-2.0/.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   See git history
+ *******************************************************************************/
```
merks: Sorry for fussing.  This should be before the package declaration, i.e., at the very start of the file.  Same for the other one below.

webwart-bln: No no, thats fine. My error. Changing it asap.

merks: You're awesome.  Thanks so much!!

merks (approved):
This is really high quality work that includes a nice enhances, appropriate tests, and improve exist bad implementation details in the framework.  🥇

webwart-bln:
Thank you!

merks:
Wow, I meant to merge this as soon as the build finished!   Sorry, sorry!!

PR #2413, opened 2026-05-20
Handle missing units in PPTXUtil.convertCssToEnum more robustly

### LINKED ISSUE #2412: BIRT 4.23 reproducable IllegalArgumentException in PPTX output with HTML content
### Description
When generating a PPTX Output of the attached report the output is invalid for PPTX and contatins the attached html-error.
The problem occurs, when a report contains HTML content with lists (order / unordered doesn't matter). I simplified the testcase to a nearly empty report that contains only 2 elements with static HTML code to reproduce. It will also occur, when the HTML content is fetched from a data set.

### Environment
- OS: Windows and Linux (doesn't matter)
- Browser: Firefox and Chrome (doesn't matter)
- BIRT Version: 4.23
### Screenshots
Add screenshots if applicable.

### Additional Context
I tried my best and tracked it down to the pull request https://github.com/eclipse-birt/birt/pull/2235
In this change in HTML2Content.java lines 645 & 646 and 656 & 657 a padding right was added. In BIRT Versions prior this change, the given report worked for PPTX as well.

In my further analysis I understood, that the PPTXUtil should probably behave better for CSS values without unit. Perhaps use a default unit or something like that.
Unfortunately I understand to little of the emitters to "copy" the behavior of the PDF emitter or get the right idea for a solution. 
Perhaps someone can give a hint on what to do here or probably create a good solution.

[PPTX HTML List-Problem.rptdesign.txt](https://github.com/user-attachments/files/28014240/PPTX.HTML.List-Problem.rptdesign.txt)
[stacktrace in pptx.html](https://github.com/user-attachments/files/28014255/stacktrace.in.pptx.html)

merks:
Fixes https://github.com/eclipse-birt/birt/issues/2412

### DIFF
--- a/engine/org.eclipse.birt.report.engine.emitter.pptx/src/org/eclipse/birt/report/engine/emitter/pptx/util/PPTXUtil.java
+++ b/engine/org.eclipse.birt.report.engine.emitter.pptx/src/org/eclipse/birt/report/engine/emitter/pptx/util/PPTXUtil.java
@@ -71,8 +71,14 @@ public class PPTXUtil {
 		if (cssDimension == null || "0".equals(cssDimension)) {
 			return 0;
 		}
-
 		DimensionType dimensionType = DimensionType.parserUnit(cssDimension);
-		return dimensionType != null ? convertToEnums(dimensionType.convertTo(DimensionType.UNITS_PT) * 1000) : 0;
+		if (dimensionType == null) {
+			return 0;
+		}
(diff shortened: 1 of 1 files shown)


merks:
@speckyspooky  @hvbtup 

Any concerns?  It fixes the problem in the issue and the tests pass, so it seems like a good thing...

speckyspooky:
Hi Ed, of course it looks good and make sense.

Issue #2410, opened 2026-05-11
PDF/A1a Conformance breaks Charts


arsenelapostolet-enercoop:
### Description

When applying PDF/A-1a conformance to a document which contains a Chart, the chart becomes blank.

### Environment
- OS: Linux
- Browser: Firefox
- BIRT Version: 4.23.0

### Additional Context

| PDF without PDF/A-1 Conformance | PDF with PDF/A-1 Conformance |
| - | - |
| <img width="652" height="841" alt="Image" src="https://github.com/user-attachments/assets/655146af-fe26-490d-b7c8-61a2b914c751" /> | <img width="652" height="841" alt="Image" src="https://github.com/user-attachments/assets/5ccd7b3a-9685-4f5e-82dd-ec23072d4eb7" /> |
| [BarChartWithinTableGroup.pdf](https://github.com/user-attachments/files/27598615/BarChartWithinTableGroup.pdf) | [BarChartWithinTableGroup_PDFA1.pdf](https://github.com/user-attachments/files/27598624/BarChartWithinTableGroup_PDFA1.pdf) |

See attached report from PDF chart libary samples (before and after enabling PDF/A1a conformance :

[ExempleChartPDFA1A.zip](https://github.com/user-attachments/files/27598809/ExempleChartPDFA1A.zip)

hvbtup:
Accessibility is not implemented for charts.
If it's just about PDF/A conformance and not about accessibility, you should try PDF/A-1b.

arsenelapostolet-enercoop:
Hi @hvbtup ,
thanks for your answer ! Indeed, we only need some kind of PDF/A conformance. 
Sadly I just tried this, but the result is the same : 

<img width="618" height="289" alt="Image" src="https://github.com/user-attachments/assets/39550eed-8fcf-4c54-83c3-16a1da68f66c" />

<img width="649" height="846" alt="Image" src="https://github.com/user-attachments/assets/1bef0247-4f22-4223-be92-6a352d1b9c3a" />

hvbtup:
This is caused by transparency, which is used somewhere in the chart (IDK where), but transpareny is not allowed in PDF/A-1.

ERROR: null
Enclosed Exception:
Transparency is not allowed: /ca = 0.0
org.openpdf.text.pdf.PdfXConformanceException: Transparency is not allowed: /ca = 0.0
[...]

PR #2422, opened 2026-06-08
[Charts/SVG] Only append the hotspot layer when it has child nodes

sebastienrougetenercoop:
This is a quick fix for #2410 which we tested by debugging our report. If we delete the hotspot layer before the rendering step, transparency is avoided and the chart appears on the output document.

## Goals

* Allow to render a chart with Chart Interactivity enabled and no event configured.
* Allow to render a chart with Chart Interactivity disabled in a PDF set to PDF/A-1 conformance.

## Next Steps ?

1. Implement test in `org.eclipse.birt.report.engine.emitter.pdf.PDFSvgEmbeddingTest` to prove the bug
2. Study some other implementation options (`org.eclipse.birt.chart.model.attribute.Interactivity.isEnable()`).

### DIFF
--- a/chart/org.eclipse.birt.chart.device.svg/src/org/eclipse/birt/chart/device/svg/SVGInteractiveRenderer.java
+++ b/chart/org.eclipse.birt.chart.device.svg/src/org/eclipse/birt/chart/device/svg/SVGInteractiveRenderer.java
@@ -1115,10 +1115,20 @@ public class SVGInteractiveRenderer {
 		mapOnloadAdded.clear();
 	}
 
+	/**
+	 * Returns the layer mapping elements to events.
+	 *
+	 * @return a node with events mapped to elements.
+	 */
 	public Node getHotspotLayer() {
 		return hotspotLayer;
 	}
 
+	/**
+	 * Creates a transparent general node as a layer to map events to SVG elements.
+	 *
+	 * @param dom is the document used to create the element.
+	 */
 	public void createHotspotLayer(Document dom) {
 		hotspotLayer = dom.createElement("g"); //$NON-NLS-1$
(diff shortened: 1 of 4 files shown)


hvbtup (approved):
I only looked at the code, and the changes look reasonable.
I did not perform any actual test.

merks:
@sebastienrougetenercoop 

Could you please reduce this to a single commit that is rebased on the current master?

Unless you are based on this commit:

https://github.com/eclipse-birt/birt/commit/4664248cfed67aecb086adabd9be1aaaa05244e0

I.e., on this:

image

The build will be using the wrong dependencies and I see from the failure that indeed it's using the last Orbit milestone and that's cause the failure.

merks:
FYI, because time is of the essence, I will create PR reusing this content...

merks:
@sebastienrougetenercoop 

I created this PR from your changes.  I added an `@since 4.24` for the new method and copied of your credentials as the author of the commit:

https://github.com/eclipse-birt/birt/pull/2424

merks:
@sebastienrougetenercoop 

Thanks for the contribution!  ❣️  I'll close this because these changes are are now committed 🚀

sebastienrougetenercoop:
Thanks for fixing the workflows @merks ! I was unable to see the error on Jenkins. And woopsie for the rebase… I appreciate your reactivity!

merks:
@sebastienrougetenercoop 

I forgot that you need to log into Jenkins now.  😭 Also, I didn't approve the GitHub action which would have been helpful.  Sorry about that!

All too often folks come here and demand things (for free) while showing no initiative whatsoever.  So when someone comes here and creates a PR request, then I'm happy to help such people help not only themselves but also all the users of BIRT!  That's how open source is supposed to work...

PR #2426, opened 2026-06-10
Prepare for the 4.25 release

merks:
- Increment all versions to 4.25.0
- Update the setup to resolve the current target platform
- Update the *.target to current dependencies

https://github.com/eclipse-birt/birt/issues/2425


merks:
It's a bit tricky because of the brotli4j native dependencies (a new non-optional dependency for openpdf) and the way we build the runtimes (with ant) from a single product (windows x86) that misses the natives for all the other platforms.  I will try to avoid the whole mess by making the requirement on the native fragments optional in Orbit; I know that openpdf guards against missing natives.

merks:
I created another PR with the same content.  It's using an updated version or Orbit where the dependency that caused this build to fail is optional.

PR #2433, opened 2026-06-15
#2432 Add missing case in the switch for PDFA3U. Validation is ok cor…

### LINKED ISSUE #2432: Validation failure for PDF.A3U
### Description
When generating a pdf with PDF.A3U and then validating it this message comes up 
```
The "conformance" property of the PDF/A Identification Schema is null instead of "U" for PDF/A-3u conforming file
```

[veraPDF-tempXMLReport161283859607825166.xml](https://github.com/user-attachments/files/28968024/veraPDF-tempXMLReport161283859607825166.xml)

It seems the enumeration PDFA3U is missing in the switch in 
```
engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
```

giangianoulas:
Added the missing case. Validation does not complain now and it should fix #2432 
```xml
      <validationReport jobEndStatus="normal" profileName="PDF/A-3u validation profile" statement="PDF file is compliant with Validation Profile requirements." isCompliant="true">
        <details passedRules="148" failedRules="0" passedChecks="26513" failedChecks="0"></details>
      </validationReport>
```

### DIFF
--- a/engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
+++ b/engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
@@ -1192,6 +1192,9 @@ public class PDFPageDevice implements IPageDevice {
 				case PDFA3B:
 					a1.addConformance("B");
 					break;
+				case PDFA3U:
+					a1.addConformance("U");
+					break;
 				case PDFA4F:
 					a1.addConformance("F");
 				}


hvbtup (approved):
Might be a silly question, but did you also check other places in the code where PDFA* is used for decisions and which should be extended? Maybe there are some, I don't know.

That said, this change certainly cannot do harm and obviously improves PDF/A-3u support.

giangianoulas:
It is not really that silly because I did not check prior to this PR. I did check now and it seems to the only place it was missing

wimjongman:
Do not forget to update the milestone for every issue and PR.

Issue #2434, opened 2026-06-16
PDF.A1A/PDF.A3A java.lang.IllegalArgumentException: The structure has kids


giangianoulas:
### Description
When setting `PdfEmitter.Conformance` to `PDF.A1A` or `PDF.A3A` with a relative simple design i get the attached exception. That happens even after applying this PR #2430. They seem to be related.

[stacktrace](https://github.com/user-attachments/files/28983405/new.28.txt)

merks:
@giangianoulas, thanks for the PRs!  And @hvbtup, thanks for the quick reviews!

PR #2430, opened 2026-06-11
Fix PDF/UA accessibility: IllegalArgumentException in openTag() and m…

Dress13:
…issing Lbl tag for lists

Fix 1 - PDFPageDevice.java:
OpenPDF throws IllegalArgumentException ("The parent has already another function") when a second child is added to a structure element whose K field contains a single non-array entry. Pre-convert K to PdfArray before calling new PdfStructureElement() to avoid the exception. Also fix incorrect error handling that created an extra nested element, add null guards in closeTag(), and handle single-child K field when restoring multi-page containers.

Fix 2 - HTML2Content.java:
For <ol> list items, setTagType("Lbl") was never called on the marker text, causing numbered list markers to appear without the required Lbl tag in the PDF structure tree. PAC 2024 reported this as an accessibility error. Also move setTagType("Lbl") outside the disc-fallback block for <ul> so all bullet types get the correct tag.

Both fixes verified with PAC 2024 on BIRT 4.24.0.

Fixes #2429

Issue #2429, opened 2026-06-11
PDF/UA: Missing Lbl tag for ordered list markers + IllegalArgumentException in openTag()


Dress13:
## Bug 1: IllegalArgumentException in PDFPageDevice.openTag()

**Affected versions:** 4.21, 4.23, 4.24
**Condition:** Tagged PDF / Enable Accessibility = true

When rendering tables or lists, OpenPDF throws:
  IllegalArgumentException: The parent has already another function

Root cause: OpenPDF stores the first child of a structure element directly
(not as array). Adding a second child throws instead of converting to array.

Fix: Pre-convert the K field to PdfArray before calling new PdfStructureElement().

---

## Bug 2: Missing Lbl tag for <ol> list markers (PAC 2024 failure)

In HTML2Content.java, setTagType("Lbl") is only set for the disc fallback
bullet (•), but not for <ol> numbered markers or other <ul> bullet types.

Fix: Always call text.setTagType("Lbl") for both <ol> and <ul> markers.

Both bugs verified fixed and tested with PAC 2024 on BIRT 4.24.0-202606100854.

merks:
These descriptions don’t appear actionable. Perhaps you plan a pull request. Without at least a test case, I wouldn’t expect much to happen.

Dress13:
Here is the full stack trace that reproduces Bug 1:


Caused by: java.lang.IllegalArgumentException: The parent has already another function.
	at org.openpdf.text.pdf.PdfStructureElement.init(PdfStructureElement.java:104)
	at org.openpdf.text.pdf.PdfStructureElement.<init>(PdfStructureElement.java:80)
	at org.eclipse.birt.report.engine.emitter.pdf.PDFPageDevice.openTag(PDFPageDevice.java:1459)
	at org.eclipse.birt.report.engine.emitter.pdf.PDFRender.visitChildren(PDFRender.java:348)
	at org.eclipse.birt.report.engine.layout.emitter.PageDeviceRender.visitContainer(PageDeviceRender.java:200)
	at org.eclipse.birt.report.engine.nLayout.area.impl.ContainerArea.accept(ContainerArea.java:533)
	...
Condition: Tagged PDF with Enable Accessibility = true. Happens with tables and nested containers.

Root cause: PdfStructureElement.init() throws when a parent element already has one child stored as a direct reference (not array). OpenPDF throws instead of converting to array.

Fix applied and verified with PAC 2024:

[...]

PR #2435, opened 2026-06-16
Fix IllegalArgumentException by calling PdfContentByte.beginMarkedCon…

giangianoulas:
Fix IllegalArgumentException by calling PdfContentByte.beginMarkedContentSequence only if required (#2434), related to #2430

### DIFF
--- a/engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPage.java
+++ b/engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPage.java
@@ -48,13 +48,16 @@ import org.openpdf.text.pdf.BaseFont;
 import org.openpdf.text.pdf.LayoutProcessor;
 import org.openpdf.text.pdf.PdfAction;
 import org.openpdf.text.pdf.PdfAnnotation;
+import org.openpdf.text.pdf.PdfArray;
 import org.openpdf.text.pdf.PdfBorderDictionary;
 import org.openpdf.text.pdf.PdfContentByte;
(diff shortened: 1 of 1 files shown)


engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPage.java
```diff
@@ -581,9 +600,12 @@ private void drawText(String text, float textX, float textY, FontInfo fontInfo,
 
 		// start drawing the text content
 		contentByte.beginText();
-
+		boolean shouldBeginMarkedContentSequence = false;
```
hvbtup: I'm not happy with the variable naming. I'd prefer something like `isMarkedContentSequence`.

hvbtup (changes requested):

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPage.java
giangianoulas: Updated

hvbtup (approved):
Well, I actually meant to rename the variable only, but I don't mind.

Is this PR independent from #2435 or does it depend on it?

If independent: OK

merks:
I tried to rebase, but it didn't want to do that.  The changes are in the same file as the previous merged changed.  I restarted the GHA to se that it passed.

giangianoulas:
@hvbtup you probably meant #2432 or #2429. It is a standalone change. 
I didn't like the method name that much too

PR #2441, opened 2026-07-08
Exclude data sets with output parameters from the implicit result cache

### LINKED ISSUE #2440: Stored-procedure OUT parameters return null when the dataset result is cached
### Description
A stored-procedure dataset (SPSelectDataSet) that exposes OUT parameters returns correct OUT-parameter values only on the first execution for a given set of parameter values. When the same dataset is referenced by two or more report elements, BIRT implicitly caches its result; on the cache hit the OUT parameters silently become null (no error, no warning). Result rows are returned correctly — only the OUT parameters are lost — so a field bound to outputParams["..."] shows the value on the first occurrence and blank/null on every repeat, producing silently wrong output.

### Environment
- OS: any (server-side data-engine logic, OS-independent)
- Browser: N/A (not a viewer/UI issue)
- BIRT Version: reproduced on 4.20.0; the responsible code is unchanged on 4.24.0 and current master (HEAD df906ce2), so all current versions are affected.

### Screenshots
N/A. Behavior with the attached design (one SP dataset referenced by two tables, input p_part=5): the first table shows p_weight=50, the second shows blank/null.

#### Root cause
On a cache hit the engine substitutes org.eclipse.birt.data.engine.executor.dscache.DataSourceQuery, whose OUT-parameter accessors unconditionally return null:
- getOutputParameterValue(int) — body is just return null;
- getOutputParameterValue(String) — body is just return null;

https://github.com/eclipse-birt/birt/blob/master/data/org.eclipse.birt.data/src/org/eclipse/birt/data/engine/executor/dscache/DataSourceQuery.java#L119-L131

org.eclipse.birt.data.engine.executor.dscache.DataSetToCache stores only result rows (no OUT-parameter path). The implicit cache is enabled by EngineExecutionHints.populateCachedDataSets (marks the dataset when needCache() is true) and built by DataSetCacheUtil.getDteDataSetCacheConfig — neither considers OUT parameters. Because DataSourceQuery in the dscache package is instantiated directly (no factory/extension point), consumers cannot override it.

#### Steps to reproduce (attached sp-outparam-cache-bug.rptdesign + setup.sql):
1. Create the stored procedure from setup.sql (an IN parameter, a one-row result set, an OUT parameter derived from the input).
2. Open the attached design and point its JDBC data source at your database.
3. The design uses one SPSelectDataSet ({call sp_part_weight(?)}, p_part=5) referenced by two tables, each displaying outputParams["p_weight"].
4. Run the report to any output format.

#### Expected vs actual
Expected: both tables show 50. Actual: first table 50, second table blank/null; result rows are correct in both.

#### Workaround
Set the model property needsCache=false on the stored-procedure dataset — this disables the implicit cache, so OUT parameters are read fresh and stay correct. It works but forfeits the result cache for that dataset.

[sp-outparam-cache-bug.xml](https://github.com/user-attachments/files/29746067/sp-outparam-cache-bug.xml)

[setup.sql](https://github.com/user-attachments/files/29746071/setup.sql)

raoulduke98:
Fixes #2440

### What

A data set that declares output parameters is now excluded from the implicit dataset result cache. After `EngineExecutionHints.populateCachedDataSets` has finalized the cached-candidate set, any data set whose design contains an `IParameterDefinition` with `isOutputMode()` is removed from it - mirroring the existing exclusion pass for extension-only filters directly below.

### Why

The implicit result cache stores only rows. On a cache hit the engine substitutes `dscache.DataSourceQuery`, whose `getOutputParameterValue(int/String)` unconditionally returns `null`, so a stored-procedure data set (`SPSelectDataSet`) re-executed with the same parameter values within one render silently loses its OUT parameter values (rows stay correct). See #2440 for the full analysis and a minimal reproducing report design.

Excluding such a data set from the row cache is behavior-preserving: re-execution returns identical rows, only the caching optimization is lost - and it is required for correctness whenever a consumer reads `outputParams["..."]`.

This targets the implicit cache path only; user-configured caching (`getJVMDataSetCacheConfig`) is not gated by `needCacheDataSet` and is out of scope for this issue.

### Test

`OutputParameterDataSetCacheTest` registers two ODA data sets (one with an output parameter, one input-only control), each referenced by two queries so both become implicit-cache candidates, and asserts via `IEngineExecutionHints.needCacheDataSet` that only the input-only data set stays cached. Verified red/green: the test fails on the unpatched engine (the output-parameter data set is cached) and passes with this change, with the control assertion unaffected.

### DIFF
--- a/data/org.eclipse.birt.data.tests/test/org/eclipse/birt/data/engine/api/OutputParameterDataSetCacheTest.java
+++ b/data/org.eclipse.birt.data.tests/test/org/eclipse/birt/data/engine/api/OutputParameterDataSetCacheTest.java
@@ -1,0 +1,83 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Eclipse contributors and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * https://www.eclipse.org/legal/epl-2.0/.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.birt.data.engine.api;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import org.eclipse.birt.core.data.DataType;
+import org.eclipse.birt.core.exception.BirtException;
+import org.eclipse.birt.core.framework.PlatformConfig;
+import org.eclipse.birt.data.engine.api.querydefn.Binding;
+import org.eclipse.birt.data.engine.api.querydefn.OdaDataSetDesign;
+import org.eclipse.birt.data.engine.api.querydefn.OdaDataSourceDesign;
+import org.eclipse.birt.data.engine.api.querydefn.ParameterDefinition;
+import org.eclipse.birt.data.engine.api.querydefn.QueryDefinition;
+import org.eclipse.birt.data.engine.api.querydefn.ScriptExpression;
+import org.eclipse.birt.data.engine.impl.DataEngineImpl;
+import org.eclipse.birt.data.engine.impl.IEngineExecutionHints;
+import org.junit.Test;
+
+import testutil.BaseTestCase;
+
+/**
+ * Data sets with output parameters must not be implicitly cached (issue #2440).
+ */
+public class OutputParameterDataSetCacheTest extends BaseTestCase {
+
+	@Test
+	public void testOutputParameterDataSetIsNotImplicitlyCached() throws BirtException {
+		DataEngineContext context = DataEngineContext.newInstance(DataEngineContext.DIRECT_PRESENTATION,
+				this.scriptContext, null, null, null);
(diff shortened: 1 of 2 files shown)


merks:
@raoulduke98 

We need you to complete the ECA to be able to merge your PR.

https://github.com/eclipse-birt/birt/blob/master/CONTRIBUTING.md#eclipse-contributor-agreement

merks:
I verified that your new test runs:

https://ci.eclipse.org/birt/job/build/job/PR-2441/1/testReport/org.eclipse.birt.data.engine.api/OutputParameterDataSetCacheTest/

And @hvbtup has approved by virtue of his comment in the issue ; better would be to mark the review approval in the PR itself.

https://github.com/eclipse-birt/birt/issues/2440#issuecomment-4913605152

So if we can get the ECA in place, I plan to merge this.

Thanks!!

raoulduke98:
thank you, eca is in place!

merks:
The changes in in the latest nightly build now

PR #2450, opened 2026-08-04
Fix issue 2429

hvbtup:
I took the changes from Dress13 and just created a PDF test runner (for manually checking the output PDFs)


hvbtup:
Hmm, the test works here. Probably because I'm using Windows, seems like a missing TTF font.
We would need to include a free TTF font (eg Liberation Sans) into the PDF test packages, or run the test on Windows only.

merks:
FYI, I would generally prefer that fixing a single problem doesn't involved a whole chain of commits, with what are effectively broken intermediate states that have less than no value in the long term history.   Nicer is to rebase on master and to amend force push until you get it right...

hvbtup:
> Nicer is to rebase on master and to amend force push until you get it right...

I thought the "Squash and merge" button will do this anyway?

merks:
Sort of.  I always try to use rebase and merge because squash and merge makes a hash of the commit message(s)

image

And it ends up making some github bot look like the author. 

But I don't want to hold things up so let's see what it produces...

merks:
Will you merge it now?

I want to create a PR related to centralizing the creation of DocumentBuilderFactory...

merks:
The committer is recorded as GitHub rather than you.

image

image

Not so ideal, but no disaster either...

PR #2455, opened 2026-08-06
PDF/UA Add support for open-only and close-only tags and for no tag

### LINKED ISSUE #2454: PDF/UA enhancements
**Is your feature request related to a problem? Please describe.**

Sometimes, the structure of the BIRT layout elements does not quite fit the logical structure which we need for PDF tags.

One example:
You have a grid with 6 columns and eg. 2 rows. Each row contains 3 pairs of abbreviation and explanation, in two adjacent colors.

The PDF tag structure should be a list (L) of items (LI), each item consisting of a (Lbl) with the acronym and a (LBody) with the explanation.

To support this, we would need opening-only and closing-only PDF tags.

Furthermore, sometimes we don't want to generate a PDF tag at all, but the "auto" default is in our way.

**Describe the solution you'd like**

Opening tag marked with a "<" e.g. "<LI" and closing tag marked with a ">" e.g. "LI>".

No PDF tag at all with "none".

**Can you contribute time or money**
I'll provide a PR.

hvbtup:
Examples:
Open-only tag: <LI
Close-only tag: LI>
No tag: none

Fixes https://github.com/eclipse-birt/birt/issues/2454


merks (approved):
Looks reasonable, but I'm really not a good judge on what's correct and best in this regard.  Your best judgement is trusted.

Issue #2385, opened 2026-03-06
BIRT 4.22 Export PDF fails in multi threaded environment when report uses Chinese characters


bineetsingh30:
### Description
In our software, we are using BIRT 4.22 to generate reports. 
While creating PDF report (which uses Chinese characters) parallelly, report fails with java.lang.UnsupportedOperationException: LayoutProcessor is already enabled (stack trace attached)

[stackTrace.txt](https://github.com/user-attachments/files/25792298/stackTrace.txt)

### Environment
- OS: Linux (Ubuntu 22.04.5 LTS)
- BIRT Version: 4.22.0

### Additional Context
Upon analysis, the issue occurs due to race condition in enabling and disabling `org.openpdf.text.pdf.LayoutProcessor`. (It cannot be enabled more than once, otherwise it raises the exception mentioned in the issue. If enabled or not, is stored in a static variable in `LayoutProcessor`.)

During PDF report generation of each report (containing Chinese character), it goes to
1. Enable Kerning And Ligatures  (`org.eclipse.birt.report.engine.layout.pdf.font.FontHandler#enableKerningAndLigatures`): This calls enable in `org.openpdf.text.pdf.LayoutProcessor`.
2. Disable `org.openpdf.text.pdf.LayoutProcessor`: In case of CJK Fonts (Chinese characters in this issue), the code disables LayoutProcessor, in the method `org.eclipse.birt.report.engine.emitter.pdf.PDFPage#validateSymbolicFont`. It is invoked in `org.eclipse.birt.report.engine.emitter.pdf.PDFPage` in drawText method.

This results in race condition when running in a multi threaded environment.

Birt template:
[JDBC_New.rptdesign.txt](https://github.com/user-attachments/files/25793469/JDBC_New.rptdesign.txt)  (remove the .txt extension from file name)

In the data, you can use value "Id-東京" in all the rows. 

Output report:
[output report.pdf](https://github.com/user-attachments/files/25793591/output.report.pdf)

merks:
It looks related to this:

https://github.com/eclipse-birt/birt/pull/2144#issuecomment-2906746149

hvbtup:
There's a new PR for OpenPDF: Once https://github.com/LibrePDF/OpenPDF/pull/1519 is merged and a new OpenPDF release is available, we should be able to adapt our code.

PR #2463, opened 2026-08-14
Migrate PDF font layout to per-document GlyphLayoutManager (#2444)

### LINKED ISSUE #2394: BIRT 4.22 Export PDF fails in multi threaded environment when running many reports first time
### Description

Given no BIRT report has run before after the process started, and now if many (17 in my case) run parallelly, one of the report fails sometimes with NullPointerException.  (2 stack traces attached, stackTrace2 is being caused due to stackTrace1)
[stackTrace1.txt](https://github.com/user-attachments/files/25970622/stackTrace1.txt)
[stackTrace2.txt](https://github.com/user-attachments/files/25970623/stackTrace2.txt)

### Environment
- OS: Linux
- BIRT Version:  4.22.0

### LINKED ISSUE #2444: Thread safety with PDF emitter
### Description
BIRT uses OpenPDF for PDF generation. There were a few changes in the last 2 years for improving kerning, ligatures etc. In particular, for fixing issues with non-latin fonts.

I said that I think these changes will cause issues in a multi-threaded environment due to the use of global variables (personally, I stopped using BIRT multi-threaded years before).

Recently, OpenPDF 3.0.4 was released, which deprecated the API BIRT uses.
The new API is explicitly described as suitable for multi-threading:

> Multithreading
> 
> GlyphLayoutManager is enabled per Document, and has no static state, so it is designed to work in a multithreading  environment processing multiple documents in separate threads.
> 

https://github.com/LibrePDF/OpenPDF/wiki/Accents,-DIN-91379,-non-Latin-scripts

It would be good to make use of this improvement in BIRT.

hyrahul:
## What this does

Migrates the PDF emitter's complex-font layout from OpenPDF's deprecated,
process-global `LayoutProcessor` to the per-`Document` `GlyphLayoutManager`
(OpenPDF's designated replacement, from LibrePDF/OpenPDF#1519). The manager holds
no static state and is designed for concurrent multi-document rendering, which is
the goal of #2444.

The global `LayoutProcessor` state is the root of the thread-safety problem in
#2444 — the same class of race that surfaced in production as #2385 (multi-threaded
PDF export failing with Chinese characters). By moving to per-document state, that
class of race is removed. (Note: races are non-deterministic, so this is addressed
by the design change rather than proven by a test; see Testing below.)

## Approach

- **`FontHandler`** and **`PDFPage.validateSymbolicFont`** — removed all
  `LayoutProcessor` usage.
- **`PDFPageDevice`** — creates a per-`Document` `GlyphLayoutManager`, gated behind
  an opt-in property (`birt.pdf.complex.font.layout.enabled` system property or
  `PdfEmitter.ComplexFontLayoutEnabled` user property), **off by default** so
  existing output is unchanged.
- **Lazy font loading** — the font configuration is not processed when the device
  is constructed (the registered font paths are empty at that point), so fonts are
  loaded into the manager lazily on the first draw. TrueType/OpenType fonts are
  loaded with kerning and ligatures enabled via `FontOptions`.
- **`PDFPage.drawText`** — attaches the manager only for fonts it has loaded, and
  detaches (`setGlyphLayoutManager(null)`) for base-14/Type1/symbolic fonts. This
  is required because `GlyphLayoutManager.supportsFont(...)` throws for any font not
  loaded through it, so un-loadable fonts must not reach it — they draw on the
  normal path instead.

## Why off by default / no regression

With the property off, no manager is created or attached, and rendering is
byte-for-byte unchanged. The removed `LayoutProcessor` path was only ever reachable
when `LayoutProcessor.supportsFont(...)` was true, i.e. for fonts loaded into it;
default rendering already flowed through `showTextBasic`.

## Testing

- Added **`PdfConcurrentRenderTest`** (in `emitter.pdf.tests`) — renders the same
  design from multiple threads on a shared engine and asserts none fail. It is a
  **regression guard / demonstration** of concurrent rendering, not a proof that
  the race is gone (by nature, races can't be deterministically triggered). The
  real safety argument is the per-document design.
- Verified manually that rendering succeeds with the feature **on** and is
  unchanged with it **off**.
- I have **not** been able to automatically verify glyph-level correctness
  (kerning/ligatures/non-Latin shaping) — automated glyph-position checks are
  impractical, consistent with the existing tests in this module. I'd very much
  value review against the non-Latin cases this feature originally targeted
  (Arabic #2133, Chinese #2385), since you have the reference for what correct
  output looks like.

## Removed test

Removed `FontHandlingPdfTest` — a manual demo (`main()` with hardcoded `C:/temp/`
paths, no assertions; its own Javadoc says results must be validated manually) built
entirely on the `LayoutProcessor` API this PR removes. Its purpose is superseded by
the `GlyphLayoutManager` path.

## OpenPDF version

No version change needed — both bundles already declare
`com.github.librepdf.openpdf;bundle-version="[3.0.0,4.0.0)"` in their manifests.

## Open questions / for review

1. **FopGlyphProcessor interaction** — `GlyphLayoutManager` and the existing
   `setGlyphSubstitutionEnabled(...)` (FopGlyphProcessor, #2141) are mutually
   exclusive. This PR does not yet coordinate them; guidance welcome (@speckyspooky).
2. **Symbolic fonts** — `validateSymbolicFont`'s handling is now implicit (symbolic
   fonts take the detached/normal path). Please confirm this preserves the intended
   behaviour for Webdings-type fonts.
3. Opening as **draft** to confirm the direction before polishing.


merks:
It would be better if you maintained a single commit by rebasing on master (as necessary), amending your commit and force pushing the changes.

engine/org.eclipse.birt.report.engine.emitter.pdf.tests/test/org/eclipse/birt/report/engine/emitter/pdf/AllTests.java
```diff
@@ -1,6 +1,6 @@
 
 /*******************************************************************************
- * Copyright (c) 2004, 2005 Actuate Corporation.
+ * Copyright (c) 2004, 2005, 2026 Actuate Corporation.
```
merks: Should just change the upper bound.
```suggestion
 * Copyright (c) 2004, 2026 Actuate Corporation.
```

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPage.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2004, 2007, 2024, 2025 Actuate Corporation and others
+ * Copyright (c) 2004, 2007, 2024, 2025, 2026 Actuate Corporation and others
```
merks: So others have been making lists too.  😢 

You can imagine the list can get arbitrarily long over time....  It should just be a range, but who can tell that given the current state.

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPage.java
```diff
@@ -642,6 +641,19 @@ private void drawText(String text, float textX, float textY, FontInfo fontInfo,
...
+		// (TrueType/OpenType); detach for base-14/Type1/symbolic fonts so they draw
+		// on the normal path.
+		GlyphLayoutManager glm = pageDevice.getGlyphLayoutManager();
+		if (glm != null) {
+			if (pageDevice.useManagerForFont(font)) {
+				pageDevice.getDocument().setGlyphLayoutManager(glm);
+			} else {
+				pageDevice.getDocument().setGlyphLayoutManager(null);
+			}
+		}
```
merks: This could use a conditional expression as the argument for setGlyphLayoutManager to reduce the 5 lines to a single line...

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
```diff
@@ -145,6 +147,26 @@ public class PDFPageDevice implements IPageDevice {
 	 */
 	protected Document doc = null;
 
+	/**
+	 * The per-document glyph layout manager (issue #2444). It is null when the
+	 * complex font layout feature is disabled.
+	 */
+	private GlyphLayoutManager glyphLayoutManager = null;
```
merks: Don't explicitly initialize to null because that's the implicit default.

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
```diff
@@ -145,6 +147,26 @@ public class PDFPageDevice implements IPageDevice {
...
+
+	/**
+	 * Whether the manager's fonts have been loaded yet (done lazily on first draw).
+	 */
+	private boolean glyphFontsLoaded = false;
```
merks: Don't explicitly initialize to false because that's the implicit default.

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
```diff
@@ -145,6 +147,26 @@ public class PDFPageDevice implements IPageDevice {
...
+
+	/**
+	 * Fonts successfully loaded into the manager (matched by identity at draw
+	 * time).
+	 */
+	private final java.util.Set<BaseFont> managerLoadedFonts = new java.util.HashSet<>();
```
merks: Please import names in general rather than using fully qualified names.

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
```diff
@@ -145,6 +147,26 @@ public class PDFPageDevice implements IPageDevice {
...
+
+	/** Fonts known NOT loadable into the manager - decided once, never retried. */
+	private final java.util.Set<BaseFont> managerUnloadableFonts = new java.util.HashSet<>();
```
merks: Please import names in general rather than using fully qualified names.

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
```diff
@@ -221,6 +243,15 @@ public class PDFPageDevice implements IPageDevice {
 	/** System property of the JavaScript version */
 	private static final String PDF_GLYPH_SUBSTITUTION_PROPERTY_KEY = "birt.pdf.glyph.substitution.enabled"; //$NON-NLS-1$
 
+	/** User property to enable per-document complex font layout */
+	private final static String PDF_COMPLEX_FONT_LAYOUT = "PdfEmitter.ComplexFontLayoutEnabled";
```
merks: I assume this should have an non-nls marker as well.

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
```diff
@@ -1781,4 +1813,177 @@ private boolean isPdfFontGlypSubstitutionEnabled() {
...
+	 * {@code GlyphLayoutManager}) is enabled. Disabled by default; can be enabled
+	 * via the system property {@code birt.pdf.complex.font.layout.enabled} or the
+	 * report user property {@code PdfEmitter.ComplexFontLayoutEnabled}.
+	 *
+	 * @return {@code true} if complex font layout should be used
+	 */
+	private boolean isPdfComplexFontLayoutEnabled() {
+		boolean enabled = Boolean.getBoolean(PDF_COMPLEX_FONT_LAYOUT_PROPERTY_KEY);
+
+		if (this.userProperties != null && !enabled
```
merks: I don't think the project is generally using `this.` qualification except when necessary.

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
```diff
@@ -1781,4 +1813,177 @@ private boolean isPdfFontGlypSubstitutionEnabled() {
...
+	 * Any failure is logged and leaves the document without a manager, so rendering
+	 * is never broken by font-layout setup.
+	 */
+	private void setupGlyphLayoutManager() {
+		if (!isPdfComplexFontLayoutEnabled()) {
+			return;
+		}
+		try {
+			glyphLayoutManager = new GlyphLayoutManager();
+		} catch (Throwable t) {
```
merks: Why might this throw an exception?

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
```diff
@@ -1781,4 +1813,177 @@ private boolean isPdfFontGlypSubstitutionEnabled() {
...
+	 * {@code glyphFontsLoaded} flag is set up front so a failure does not loop).
+	 * Only TrueType/OpenType fonts can be loaded into the manager; base-14 and Type1
+	 * fonts are handled on the normal draw path.
+	 */
+	private void ensureGlyphFontsLoaded() {
+		if (glyphLayoutManager == null || glyphFontsLoaded) {
+			return;
+		}
+		glyphFontsLoaded = true; // set first so a failure doesn't loop
+		java.util.Set<String> fontPaths = org.eclipse.birt.report.engine.layout.pdf.font.FontMappingManagerFactory
```
merks: Use imports.

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
```diff
@@ -1781,4 +1813,177 @@ private boolean isPdfFontGlypSubstitutionEnabled() {
...
+				.getInstance().getFontPathes();
+		if (fontPaths == null) {
+			return;
+		}
+		for (String pathEntry : fontPaths) {
+			if (pathEntry == null) {
+				continue;
+			}
+			java.io.File f = new java.io.File(pathEntry);
```
merks: Imports.

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
```diff
@@ -1781,4 +1813,177 @@ private boolean isPdfFontGlypSubstitutionEnabled() {
...
+	 * Ensures fonts are loaded (lazily) on first use. Returns {@code true} only for
+	 * fonts that were successfully loaded into the manager (TrueType/OpenType);
+	 * base-14, Type1 and symbolic fonts return {@code false} and are drawn on the
+	 * normal path, because {@code supportsFont} throws for any font not loaded
+	 * through the manager.
+	 *
+	 * @param font the base font about to be drawn
+	 * @return {@code true} if the manager should be attached for this font
+	 */
+	boolean useManagerForFont(BaseFont font) {
```
merks, line 1862: Does this need to be package protected?

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
```diff
@@ -1781,4 +1813,177 @@ private boolean isPdfFontGlypSubstitutionEnabled() {
...
+	}
+
+	/**
+	 * Get the per-document glyph layout manager, used by the page to attach or
+	 * detach it per draw. Returns {@code null} when complex font layout is
+	 * disabled.
+	 *
+	 * @return the glyph layout manager, or {@code null}
+	 */
+	GlyphLayoutManager getGlyphLayoutManager() {
```
merks, line 1902: Does this need to be package protected?

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
```diff
@@ -1781,4 +1813,177 @@ private boolean isPdfFontGlypSubstitutionEnabled() {
...
+		return glyphLayoutManager;
+	}
+
+	/**
+	 * Get the underlying PDF document, used by the page to attach or detach the
+	 * glyph layout manager per draw.
+	 *
+	 * @return the PDF document
+	 */
+	Document getDocument() {
```
merks: Does this need to be package protected?

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
```diff
@@ -1781,4 +1813,177 @@ private boolean isPdfFontGlypSubstitutionEnabled() {
...
+	}
+
+	/**
+	 * Recursively load the TrueType/OpenType font files from a directory into the
+	 * glyph layout manager.
+	 *
+	 * @param dir the directory to scan
+	 */
+	private void loadFontsFromDirectory(java.io.File dir) {
+		java.io.File[] files = dir.listFiles();
```
merks: Import

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
```diff
@@ -1781,4 +1813,177 @@ private boolean isPdfFontGlypSubstitutionEnabled() {
...
+		if (files == null) {
+			return;
+		}
+		for (java.io.File file : files) {
```
merks: Import.

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
```diff
@@ -1781,4 +1813,177 @@ private boolean isPdfFontGlypSubstitutionEnabled() {
...
+	 * {@code .otf} extension).
+	 *
+	 * @param file the file to test
+	 * @return {@code true} if it is a loadable TrueType/OpenType font file
+	 */
+	private boolean isOpenTypeFontFile(java.io.File file) {
+		if (file == null || !file.isFile()) {
+			return false;
+		}
+		String n = file.getName().toLowerCase();
```
merks: Only windows file systems are case sensitive, so this is a bit questionable.

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/README.md
```diff
@@ -201,3 +201,32 @@ A detailed description is given at the top level configuration file fontsConfig.
...
+					base-14 and Type1 fonts continue to use the standard rendering path.
+					This replaces the deprecated global LayoutProcessor.
+	Parameter  	-Dbirt.pdf.complex.font.layout.enabled
+	Location   	JVM
+	Data type  	boolean
+	Values     	true, complex font layout with kerning and ligatures is enabled
+					false, complex font layout is disabled
+	Default    	false
+	Reference  	see PdfEmitter.ComplexFontLayoutEnabled
+	Since      	4.21
```
merks: Is this really when it was added?  Anything you add now is  `@since 4.24`.

engine/org.eclipse.birt.report.engine.tests/test/org/eclipse/birt/report/engine/layout/pdf/font/FontHandlingPdfTest.java
```diff
@@ -1,177 +0,0 @@
...
- * of font handling and pdf configuration based on the layout processor which is
- * used by openPDF. The primary target is to test the different options of
- * kerning and ligatures. The class is not established to instantiate JUnits
- * tests. This class support the creation of further test examples and the
- * results must be validated manually.
- *
- * @since 3.3
- *
- */
-public class FontHandlingPdfTest {
```
merks, line 42: I guess this wasn't really a test but something bogus folks ran on Windows themselves.

merks (commented):
Many small comments.  In general copyright dates should be a range and type names should be imported not fully qualified in the body.

hvbtup:
> It would be better if you maintained a single commit by rebasing on master (as necessary), amending your commit and force pushing the changes.

For Git noobs like me:

Once you `--amend` to an already-pushed commit, you need to use the `--force` option with `git push`. I was always scared of using it, but this seems to be the usual way.

engine/org.eclipse.birt.report.engine.tests/test/org/eclipse/birt/report/engine/layout/pdf/font/FontHandlingPdfTest.java
hyrahul: @merks This test appears to be a legacy/manual demo test that was added quite some time ago. Since it doesn't seem to be serving a testing purpose anymore, I removed FontHandlingPdfTest as part of this PR and it built entirely on the LayoutProcessor API. Its purpose is superseded by the GlyphLayoutManager path as part of new enhancement.

merks: Yes it looked like garbage to me too.

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
hyrahul: it's used by other classes e.g. PDFPage.java in the same package, but not outside the package so package-private visibility should be sufficient here.

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
hyrahul: it's used by other classes e.g. PDFPage.java in the same package, but not outside the package so package-private visibility should be sufficient here.

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
hyrahul: The intention here is to make the file-extension check case-insensitive. I’ll use Locale.ROOT for the lowercase conversion so the behaviour is deterministic and independent of the system’s default locale.

merks: I would just leave that out. Most of such things will run on a Linux server and it’s just not sensible there.

hvbtup: No, please keep it. There are still enough files with a *.TTF extension instead of *.ttf.
Or, just strip the extension first and then check both *.TTF and *.ttf.

merks: I defer to the judgment of those with domain knowledge. 🤓

engine/org.eclipse.birt.report.engine.emitter.pdf.tests/test/org/eclipse/birt/report/engine/emitter/pdf/PdfConcurrentRenderTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+	@Override
+	protected void setUp() throws Exception {
+		super.setUp(); // creates the shared engine (and configures fonts)
+		outputDir = Files.createTempDirectory("birt-pdf-concurrent").toFile();
+	}
+
+	@Override
+	protected void tearDown() throws Exception {
+		deleteRecursively(outputDir);
```
hvbtup, line 69: Hmm, I don't know if it is a good idea to cleanup the output directory after the test.
Wouldn't it be better to do this in the setUp() method?
I mean, what if you want to manually check the generated PDFs?

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPage.java
```diff
@@ -642,6 +641,16 @@ private void drawText(String text, float textX, float textY, FontInfo fontInfo,
...
 			contentByte.setFontAndSize(font, MIN_FONT_SIZE * 2);
 		}
+		// issue #2444: OpenPDF's showText routes through the glyph layout manager
+		// when one is attached. supportsFont(...) throws for any font not loaded
+		// into the manager, so attach it only for fonts it has loaded
+		// (TrueType/OpenType); detach for base-14/Type1/symbolic fonts so they draw
+		// on the normal path.
+		GlyphLayoutManager glm = pageDevice.getGlyphLayoutManager();
+		if (glm != null) {
+			pageDevice.getDocument().setGlyphLayoutManager(pageDevice.useManagerForFont(font) ? glm : null);
```
hvbtup: Setting a document property on and off while drawing text, depending on the font for this particular piece of text, seems like a code smell to me, but probably this is necessary here.
OTOH it's still much better than changing a global. At least this won't introduce race conditions, because the same document is only writen/modified by the same thread.

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
```diff
@@ -1781,4 +1817,171 @@ private boolean isPdfFontGlypSubstitutionEnabled() {
...
+		}
+		for (String pathEntry : fontPaths) {
+			if (pathEntry == null) {
+				continue;
+			}
+			File f = new File(pathEntry);
+			if (f.isDirectory()) {
+				loadFontsFromDirectory(f);
+			} else if (isOpenTypeFontFile(f)) {
+				tryLoadFont(f);
```
hvbtup: Isn't this causing a massive overhead? On my machine, the directory C:\windows\fonts is 400 MB.

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
```diff
@@ -1781,4 +1817,171 @@ private boolean isPdfFontGlypSubstitutionEnabled() {
...
+	 * skipped.
+	 *
+	 * @param file the font file to load
+	 */
+	private void tryLoadFont(File file) {
+		try {
+			FontOptions options = new FontOptions()
+					.setKerningOn()
+					.setLigaturesOn();
+			Font loaded = glyphLayoutManager.loadFont(file.getAbsolutePath(), 12f, options);
```
hvbtup: What is this magic constant 12f?
A font size in point? If so, why do we need a font size here? And what happens when the report wants to render the font in e.g. 8pt?

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
```diff
@@ -1781,4 +1817,171 @@ private boolean isPdfFontGlypSubstitutionEnabled() {
...
+	 * Attempt to load a single font file into the glyph layout manager with kerning
+	 * and ligatures enabled, recording the resulting base font so it can be matched
+	 * at draw time. Files that cannot be loaded (e.g. unsupported legacy fonts) are
+	 * skipped.
+	 *
+	 * @param file the font file to load
+	 */
+	private void tryLoadFont(File file) {
+		try {
+			FontOptions options = new FontOptions()
```
hvbtup: Maybe later we will need a way to make Kerning and Ligature usage a bit more configurable, but OK for now.

engine/org.eclipse.birt.report.engine.tests/test/org/eclipse/birt/report/engine/layout/pdf/font/FontHandlingPdfTest.java
hvbtup: The idea of this test was certainly to create PDFs with different options for kerning and ligatures and to thoroughly examine the results with a sharp eye.
I think we need something similar for the new PdfEmitter.ComplexFontLayoutEnabled option.

hvbtup (changes requested):
Some questions here and there.
I think we need a test case that allows manually checking the PDF results.
In particular, some of the examples from the old issues with Arabic, Chinese etc, and a demo report with a large font size to make the kerning actually visible.
Like this text: "AV cables VA Maya Vase WM WA AW with Kerning".

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPage.java
hyrahul: Agreed it's not pretty. It's forced by the API: supportsFont(...) throws for any font not loaded into the manager rather than returning false, so an unloaded font reaching showText fails the render (hit this with base-14 Helvetica). Since base-14 can't be loaded at all and reports mix them with TrueType on the same page, the decision has to be per-text-run. Cost is a Set lookup plus a field assignment, and as you note it stays thread-local.

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
hyrahul: Wired the FontOptions to the existing kerning-and-ligatures tag in fontsConfig.xml (via FontMappingManager.useFontKerningAndLigatures()) instead of always enabling both. Note this means the new property alone won't turn kerning on unless that tag is also enabled — say the word if you'd rather the property imply it.

hvbtup: It's a difficult decision, what the default should be.
There are pros and cons.
If kerning/ligatures need to be enabled explicitly, then this is the old behavior, which is good for users of latin fonts like me, because we don't need it, and enabling it now will in some cases result in different line-breaking behavior, so the same report with the same data can look different after updating BIRT. OTOH the old behavior (no kerning/ligatures) was bad for those languages where kerning/ligatures are actually necessary.
We don't know our users, but given that BIRT was not good at writing Arabic, Chinese etc, I guess that most of the existing users used BIRT with latin-based alphabets. Anyway, if the whole feature can be enabled or disabled with a single system property (as it seems), then I think it's better to enable it by default, because the result should look better/correct, and if not, disabling it is just a SystemProperty away.

What I actually meant with "more configurable", is the ability to enable/disable it for parts of the layout of a report, e.g. as as an report item property like `fontSize`. I was vague about it, because I don't know if such fine granularity is actually needed or not.

hvbtup:
I startet a CI run. Using the build artifacts, it will be easier for other users to test.

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
hyrahul: Good catch — it's a leftover magic number. Looking at GlyphLayoutFontManager.loadFont, the size is only used to build the returned Font wrapper (new Font(baseFont, fontSize)), which we discard; it isn't stored in the manager's AWT font map, and the actual size is applied per draw via setFontAndSize. So it has no effect on layout. I've replaced it with a named constant and documented

engine/org.eclipse.birt.report.engine.emitter.pdf.tests/test/org/eclipse/birt/report/engine/emitter/pdf/PdfConcurrentRenderTest.java
hyrahul: I moved it to setUp() method.

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
hyrahul: On the default: worth flagging it's currently two switches, not one — the new property attaches the manager, and the existing kerning-and-ligatures tag in fontsConfig.xml controls whether kerning/ligatures are applied. So enabling it today means setting both.

I'd lean toward keeping it off for this change and revisiting once it's had some exposure, mainly because of the line-breaking point you raise — different metrics mean different pagination, so the same report and data can paginate differently after an upgrade. That feels like something to opt into rather than inherit. Happy to flip it if you prefer; it's a one-line change.

On per-item configurability: that'd mean a new report item property threaded through the model and layout down to FontInfo — well outside this PR, but worth its own issue if it turns out to be needed.

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
hyrahul: Verified with numbers rather than eyeballing:

Why my first check showed nothing: NotoSans (the bundled test font) reports hasKernPairs=false, with AV/VA/WA/AW all 0 — it has no kerning data, so identical output was correct rather than a fault. For comparison, Arial: AV/VA = -74, WA/AW = -37; Times New Roman: -128/-128/-110/-80.

With Arial loaded through the manager, same report:

kerning on vs off → 737 bytes differ in the content stream (28467 bytes each)
kerning off vs off (control) → byte-identical

So renders are deterministic, and the difference is attributable to the kerning options.

Font routing in a single render: Arial (TrueType-Unicode) loaded into the manager; Times-Roman (base-14) and STSong-Light (CJK) rejected and drawn on the normal path, no exception.

Load volume : on my Mac lazy loading is now 6 across the 8 concurrent documents — only the fonts actually drawn. Most of the configured font directories don't exist on macOS, I haven't measured on Windows.

Arabic rendered once an Arabic-capable font was present (blank with NotoSans alone). I can't judge whether the shaping is correct — that's where your review of the non-Latin cases would help.
image

engine/org.eclipse.birt.report.engine.tests/test/org/eclipse/birt/report/engine/layout/pdf/font/FontHandlingPdfTest.java
hyrahul: Agreed. I've added glm-manual-check.rptdesign for this — kerning pairs (AV/VA/WA/AW) at 40pt in regular and bold, a ligature line, plus Arabic and Chinese samples, with instructions in the report itself to render with the option off and on and compare.

One catch: the bundled NotoSans reports hasKernPairs=false with 0 for all those pairs, so the design can't show a kerning difference as-is. I verified the mechanism with Arial (AV = -74), which isn't redistributable. Bundling a kern-capable font would make the design work standalone — let me know if there's one already acceptable in the tree.

hyrahul:
Pushed an update (amended, single commit):

- Lazy per-font loading via `FontFactory.getFontImp().getFontPath(...)` instead of scanning font directories
- Kerning/ligatures now driven by the existing `kerning-and-ligatures` config rather than always on
- `FONT_LOAD_SIZE` constant replacing the `12f`
- Added `glm-manual-check.rptdesign` for the manual check
- Test cleanup moved to `setUp` so the PDFs survive for inspection; failures now reported as a summary
- Style points: imports, no explicit default initialisers, NLS marker, conditional expression, copyright ranges, `Since 4.24`

Verification details in the comment above.

engine/org.eclipse.birt.report.engine.emitter.pdf.tests/test/org/eclipse/birt/report/engine/emitter/pdf/AllTests.java
```diff
@@ -1,6 +1,6 @@
 
 /*******************************************************************************
- * Copyright (c) 2004, 2005 Actuate Corporation.
+ * Copyright (c) (2004-2026) Actuate Corporation.
```
merks: It should be like this:

```suggestion
 * Copyright (c) 2004, 2026 Actuate Corporation.
```

engine/org.eclipse.birt.report.engine.emitter.pdf.tests/test/org/eclipse/birt/report/engine/emitter/pdf/PdfConcurrentRenderTest.java
```diff
@@ -0,0 +1,159 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Rahul Pal
```
merks:
```suggestion
 * Copyright (c) 2026 Rahul Pal and others.
```

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPage.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2004, 2007, 2024, 2025 Actuate Corporation and others
+ * Copyright (c) (2004-2026) Actuate Corporation and others
```
merks:
```suggestion
 * Copyright (c) 2004, 2026 Actuate Corporation and others
```

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
```diff
@@ -1781,4 +1826,160 @@ private boolean isPdfFontGlypSubstitutionEnabled() {
...
+	private boolean isOpenTypeFontFile(File file) {
+		if (file == null || !file.isFile()) {
+			return false;
+		}
+		String n = file.getName().toLowerCase(Locale.ROOT);
+		return n.endsWith(".ttf") || n.endsWith(".otf");
+	}
+
+
+
```
merks: One blank line is enough.

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/README.md
```diff
@@ -201,3 +201,34 @@ A detailed description is given at the top level configuration file fontsConfig.
...
+					"kerning-and-ligatures" configuration tag.
+					This replaces the deprecated global LayoutProcessor.
+	Parameter  	-Dbirt.pdf.complex.font.layout.enabled
+	Location   	JVM
+	Data type  	boolean
+	Values     	true, complex font layout is enabled
+					false, complex font layout is disabled
+	Default    	false
+	Reference  	see PdfEmitter.ComplexFontLayoutEnabled, Kerning and Ligatures
+	Since      	4.24
```
merks: If you just added this for this release cycle, it should be 4.25...

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/README.md
```diff
@@ -201,3 +201,34 @@ A detailed description is given at the top level configuration file fontsConfig.
...
+
+	Content    	configuration to enable per-document complex font layout for PDF output
+					The user property works only if the global configuration of complex font layout is disabled.
+	Location   	report
+	Data type  	boolean
+	Values     	true, complex font layout is enabled
+					false, complex font layout is disabled
+	Default    	false
+	Reference  	see -Dbirt.pdf.complex.font.layout.enabled, Kerning and Ligatures
+	Since      	4.24
```
merks: If you just added this for this release cycle, it should be 4.25...

engine/org.eclipse.birt.report.engine/src/org/eclipse/birt/report/engine/layout/pdf/font/FontHandler.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2004, 2008, 2025 Actuate Corporation and others
+ * Copyright (c) (2004-2026) Actuate Corporation and others
```
merks:
```suggestion
 * Copyright (c) 2004, 2026 Actuate Corporation and others
```

merks (commented):
It generally looks good to me.  Just some very nitty suggestion.

If @hvbtup considers this good to go then I'm fine with that.  I can't really judge the low level implementation details personally.

engine/org.eclipse.birt.report.engine.emitter.pdf.tests/test/org/eclipse/birt/report/engine/emitter/pdf/AllTests.java
hyrahul: Resolved!

engine/org.eclipse.birt.report.engine.emitter.pdf.tests/test/org/eclipse/birt/report/engine/emitter/pdf/PdfConcurrentRenderTest.java
hyrahul: Resolved!

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPage.java
hyrahul: Resolved!

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/PDFPageDevice.java
hyrahul: Resolved!

engine/org.eclipse.birt.report.engine.emitter.pdf/src/org/eclipse/birt/report/engine/emitter/pdf/README.md
hyrahul: updated with 4.25.

merks (approved):
Looks fine to me.

hvbtup (changes requested):
I couldn't spot an error in the code, but I tested it and it simply does not work.

What I did for testing:

1. Downloaded and unpacked the build artifact (birt-report-designer-win32-x86_64.zip)
2. Downloaded the report glm-manual-check.rptdesign
3. Opened the report in the designer
4. No Kerning or ligatures visibile (so far, this is expected)
5. Closed the designer
6. Modified the file birt.ini and added -Dbirt.pdf.complex.font.layout.enabled=true
7. Repeated the steps 3-5, still no kerning, still, this is expected
8. Modified the file D:\eclipse\birt-pr-2463\eclipse\plugins\org.eclipse.birt.report.engine.fonts_4.25.0.v202606101527\fontsConfig.xml and changed the line 68 to `<kerning-and-ligatures enabled="true" />`
9. Repeated steps 3-5, still no kerning.

Since the kerning is enabled in fontConfig.xml and in the system property now, this should work AFAIK, but it does not.
The visible result is no different with kerning enabled, the gap between the letters V and A is still the same (too far).

I also created a report of my own, which shows the system property and user property; with the same result. Please find it attached here.

[kerning1.zip](https://github.com/user-attachments/files/31134179/kerning1.zip)

hyrahul:
# Test Case Verification Update

@hvbtup Thanks for sharing the `rptdesign` template. It was really helpful in reproducing and verifying one of the test-case scenarios.

I’ve made some minor fixes based on the observations and verified the following scenarios **before and after the fix**:

| System Property | User Property | Result |
|---|---|---|
| Off | No user property | `enabled=false` |
| Off | `user prop=true` | `enabled=true`(was broken) |
| On | No user property | `enabled=true` |
| On | `user prop=true` | `enabled=true` |

With Arial (which has kern pairs, unlike the bundled Noto Sans), the content stream changes from a plain show-text to a positioned array, and the adjustments match the font's kern table:

kerning off:  <0024 0039 ...>Tj
kerning on:   [<0024>74.22<0039>91.8<...>]TJ

0024=A, 0039=V — 74.22 matches Arial's AV kern of -74; W/A shows 37.11, matching -37. Your own report renders the same way via the user property alone: [<0024>74.22<0039>...]TJ for "AV module VA".

The attached results show the behavior **before and after the fix**.
image

Could you please give your test case another shot with the latest changes? 
It would be great to confirm the fix from your scenario as well.

Thanks!

hvbtup:
OK, I can now see that it seems to works with my test report.

What did you change in the code to make it work now?

I can also see that kerning works for the English text in the `glm-manual-check.rptdesign`.

I can not verify if the rendering of the Arabic and Chinese texts is correct in that report's PDF output.
I don't know if its wrong or right, I only see that the Arabic text is rendered differently.

To me it seems that there might be an issue with the reading direction LTR / RTL?

I think we need someone who can read Arabic to verify if it's correct.

hyrahul:
> What did you change in the code to make it work now? -> Three fixes:
1. `setupGlyphLayoutManager()` ran in the constructor, before `userProperties` was assigned, so the user property was always null there. The manager is now created lazily on first draw.

2. `Document.setGlyphLayoutManager(...)` sets a field on that instance, but `PdfContentByte.showText` reads it from `writer.getPdfDocument()`, a different object. Now set via `PdfWriter.setGlyphLayoutManager(...)`.

3. `loadFont(...)` creates its own `BaseFont` and only recognises that instance, so `supportsFont` returned false for BIRT's font. The device now keeps the manager's base font and draws with it.
 
> 
> I can also see that kerning works for the English text in the `glm-manual-check.rptdesign`.
> 
> I can not verify if the rendering of the Arabic and Chinese texts is correct in that report's PDF output. I don't know if its wrong or right, I only see that the Arabic text is rendered differently.
> 
> To me it seems that there might be an issue with the reading direction LTR / RTL? -> On the Arabic: you were right, the reading direction was wrong. BIRT applies bidi reordering and shaping before the emitter, so the text arrives in visual order using Arabic presentation forms (U+FExx). The manager was running its own bidi over that and reversing it a second time. Fixed by setting `FontOptions.setRunDirectionLtr()`, so the manager draws the text as delivered.

> I think we need someone who can read Arabic to verify if it's correct. -> The output now matches the source string and starts from the right margin, but I can't judge the finer points, so a second pair of eyes on that would still be welcome.

hyrahul:
Hi @merks and @hvbtup 
All review comments are addressed and verified. CI is green. Ready to merge from my side whenever you've had a chance to look.

merks (approved):
If @hvbtup approves then I'm fine with the changes.

hvbtup (approved):

hvbtup:
@wimjongman @merks I think for the next release this change should be emphasized; people who use BIRT for non-latin writing system should definitely test. Also, it should be pointed out that, for kerning and ligatures to work, two settings are necessary: 
First, either a PDF emitter specific setting in the rptdesign file or a System property.
Second, `<kerning-and-ligatures enabled="true" />` in `fontsConfig.xml`.

merks:
@wimjongman 

I don't know where such documentation should exist that people would find it.  We have the root READ.md buch not much else.  

I have no clue how the website is maintained.

https://github.com/eclipse-birt/birt-website/commits/main/

There is a **massive** ignored backlog of security issues

https://github.com/eclipse-birt/birt-website/security/dependabot

PR #2471, opened 2026-08-23
Set encodings for all the *.msg files

merks:
- Remove feature*.msg files the few features that have them; these are unused.

https://github.com/eclipse-birt/birt/issues/2467


merks:
@hyrahul 


I believe this globally sets the encodings for all the *.msg in the workspace.

I specified it like this and then called toString on the Charset instance name:
```
			"_de_DE.msg", Charset.forName("Cp1252"), //
			"_fr_FR.msg", Charset.forName("Cp1252"), //
			"_es_ES.msg", Charset.forName("Cp1252"), //
			"_zh_CN.msg", Charset.forName("GBK"), //
			"_ja_JP.msg", Charset.forName("SJIS"), //
			"_ko_KR.msg", Charset.forName("MS949"), //
```

Maybe I should have just used the string directly?  

It's generated so I could change it with little effort...

merks:
This morning I started to wonder if it wouldn't be better to convert all the *.msg file to UTF-8?   We'd simply need to change this pom.xml to specify UTF-8 for the source file:

image

I think modern Java versions also directly support reading *.properties that are encoded as UTF-8

https://openjdk.org/jeps/226

It makes one wonder if we even need *.msg representation?

hyrahul:
> This morning I started to wonder if it wouldn't be better to convert all the *.msg file to UTF-8? We'd simply need to change this pom.xml to specify UTF-8 for the source file:
> 
> image
> I think modern Java versions also directly support reading *.properties that are encoded as UTF-8
> 
> https://openjdk.org/jeps/226
> 
> It makes one wonder if we even need *.msg representation?

@merks 
One thing worth verifying before this goes further: BIRT's engine loads
messages through ICU4J's `UResourceBundle.getBundleInstance(...)`
(see `core/org.eclipse.birt.core/.../i18n/ResourceHandle.java`), not
directly through `java.util.PropertyResourceBundle`. JEP 226 only changes
the *latter's* default encoding — whether it actually reaches all the way
through ICU4J's bundle loading for `.properties`-format resources is worth
confirming with a real test before committing to the UTF-8 conversion,
since if it doesn't, we'd get a build that succeeds but a runtime that
still mis-decodes.

Also, to be precise about scope: just changing the `native2ascii` encoding
attribute in `nl/pom.xml` isn't sufficient on its own — the actual bytes
of every `.msg` file would need re-encoding too, otherwise native2ascii
would misread the still-native-encoded bytes as UTF-8.

merks:
Yes, best not to assume that UTF-8 encoding works reading the *.properties so easier to leave that part alone. 
Indeed my plan was to do actual encoding conversion.  I.e., read with the old/expected encoding and write with UTF-8.

So I started with the reading part:

image

But these throw exception while reading:

- /org.eclipse.birt.chart.feature/feature_ja_JP.msg `Input length = 1`
- /org.eclipse.birt.chart.feature/feature_zh_CN.msg `Input length = 1`
- /org.eclipse.birt.data.oda.pojo.nl/plugin_ja_JP.msg `Input length = 1`
- /org.eclipse.birt.data.oda.pojo.nl/plugin_ko_KR.msg `Input length = 1`
- /org.eclipse.birt.data.oda.pojo.nl/plugin_zh_CN.msg `Input length = 1`
- /org.eclipse.birt.doc.feature/feature_ja_JP.msg `Input length = 1`
- /org.eclipse.birt.doc.feature/feature_ko_KR.msg `Input length = 1`
- /org.eclipse.birt.example.feature/feature_ja_JP.msg `Input length = 1`
- /org.eclipse.birt.example.feature/feature_ko_KR.msg `Input length = 1`
- /org.eclipse.birt.report.data.oda.excel.nl/plugin_es_ES.msg `Input length = 1`
- /org.eclipse.birt.report.data.oda.excel.nl/plugin_ja_JP.msg `Input length = 1`
- /org.eclipse.birt.report.data.oda.excel.nl/plugin_ko_KR.msg `Input length = 1`
- /org.eclipse.birt.report.data.oda.excel.nl/plugin_zh_CN.msg `Input length = 1`
- /org.eclipse.birt.report.data.oda.jdbc.dbprofile.sampledb.nl/plugin_ja_JP.msg `Input length = 1`
- /org.eclipse.birt.report.data.oda.jdbc.dbprofile.sampledb.nl/plugin_ko_KR.msg `Input length = 1`
- /org.eclipse.birt.report.data.oda.jdbc.dbprofile.sampledb.nl/plugin_zh_CN.msg `Input length = 1`
- /org.eclipse.birt.sdk.feature/feature_ja_JP.msg `Input length = 1`
- /org.eclipse.birt.sdk.feature/feature_ko_KR.msg `Input length = 1`

I was going to delete the feature `*.msg` but that still leaves a few others that apparently are incorrect in some way...  Given the languages involved, I'm not really sure how to determine the problem for those `plugin*.msg` files...

hyrahul:
> Yes, best not to assume that UTF-8 encoding works reading the *.properties so easier to leave that part alone. Indeed my plan was to do actual encoding conversion. I.e., read with the old/expected encoding and write with UTF-8.
> 
> So I started with the reading part:
> 
> image
> But these throw exception while reading:
> 
> * /org.eclipse.birt.chart.feature/feature_ja_JP.msg `Input length = 1`
> * /org.eclipse.birt.chart.feature/feature_zh_CN.msg `Input length = 1`
> * /org.eclipse.birt.data.oda.pojo.nl/plugin_ja_JP.msg `Input length = 1`
> * /org.eclipse.birt.data.oda.pojo.nl/plugin_ko_KR.msg `Input length = 1`
> * /org.eclipse.birt.data.oda.pojo.nl/plugin_zh_CN.msg `Input length = 1`
> * /org.eclipse.birt.doc.feature/feature_ja_JP.msg `Input length = 1`
> * /org.eclipse.birt.doc.feature/feature_ko_KR.msg `Input length = 1`
> * /org.eclipse.birt.example.feature/feature_ja_JP.msg `Input length = 1`
> * /org.eclipse.birt.example.feature/feature_ko_KR.msg `Input length = 1`
> * /org.eclipse.birt.report.data.oda.excel.nl/plugin_es_ES.msg `Input length = 1`
> * /org.eclipse.birt.report.data.oda.excel.nl/plugin_ja_JP.msg `Input length = 1`
> * /org.eclipse.birt.report.data.oda.excel.nl/plugin_ko_KR.msg `Input length = 1`
> * /org.eclipse.birt.report.data.oda.excel.nl/plugin_zh_CN.msg `Input length = 1`
> * /org.eclipse.birt.report.data.oda.jdbc.dbprofile.sampledb.nl/plugin_ja_JP.msg `Input length = 1`
> * /org.eclipse.birt.report.data.oda.jdbc.dbprofile.sampledb.nl/plugin_ko_KR.msg `Input length = 1`
> * /org.eclipse.birt.report.data.oda.jdbc.dbprofile.sampledb.nl/plugin_zh_CN.msg `Input length = 1`
> * /org.eclipse.birt.sdk.feature/feature_ja_JP.msg `Input length = 1`
> * /org.eclipse.birt.sdk.feature/feature_ko_KR.msg `Input length = 1`
> 
> I was going to delete the feature `*.msg` but that still leaves a few others that apparently are incorrect in some way... Given the languages involved, I'm not really sure how to determine the problem for those `plugin*.msg` files...


@merks 
Ran the diagnostic on all 18 files. Two different issues:

10 files (plugin_*.msg under data.oda.pojo, data.oda.excel, data.oda.jdbc.dbprofile.sampledb) are already valid UTF-8, just mislabeled — no conversion needed, just relabel.
8 files (feature_*.msg under chart/doc/example/sdk) already have replacement-character bytes baked in — that text is already lost, unrelated to any conversion attempt. Needs git-history check or native-speaker fix, not a re-encode.

merks:
Attached is the utility I hacked together:

[birt-convert.zip](https://github.com/user-attachments/files/31364425/birt-convert.zip)

I launch it specify the real workspace where birt is located:

One must close that IDE so the workspace isn't locked.  I also make the workspace not auto-build because something the utility had problems with the builder kicking in.

I only used a workspace so I could call setCharset, but for converting the encodings of the files, one doesn't really need the workspace API.

So you want to take over the process of converting all `*.msg`files to UTF-8?   And change the pom to expect that encoding?  That would be the best long term I think!

hvbtup:
> So you want to take over the process of converting all *.msgfiles to UTF-8? And change the pom to expect that encoding? That would be the best long term I think!

+1

hyrahul:
> Attached is the utility I hacked together:
> 
> [birt-convert.zip](https://github.com/user-attachments/files/31364425/birt-convert.zip)
> 
> I launch it specify the real workspace where birt is located:
> 
> One must close that IDE so the workspace isn't locked. I also make the workspace not auto-build because something the utility had problems with the builder kicking in.
> 
> I only used a workspace so I could call setCharset, but for converting the encodings of the files, one doesn't really need the workspace API.
> 
> So you want to take over the process of converting all `*.msg`files to UTF-8? And change the pom to expect that encoding? That would be the best long term I think!

@merks and @hvbtup 
Yes — I'll take this over: convert all *.msg files to UTF-8 and update the pom to expect that encoding, as you suggested.

merks:
@hyrahul 

Your thorough approach is very much appreciated!

merks:
FYI, I expect we can close this PR when you create a better alternative that re-encodes the files correctly.  Thanks again.

hyrahul:
> FYI, I expect we can close this PR when you create a better alternative that re-encodes the files correctly. Thanks again.

sounds good to me.
Merci!

PR #2473, opened 2026-08-25
Add reportContext.openDataSet() to read data sets from script (#1351)

### LINKED ISSUE #1351: Add possibility to process data from a datasource before generation starts.
There are quite some use-cases where you want to be able to determine certain properties of the output based on information from a data source. A number of things are processed before a connection with data-source is established.
Some examples:
- output file name
- output type
- master page background image

The only possibility to get data from a data source earlier in the process, is using a dynamic parameter.  However when the runtime engine is used this is not a viable way. 

Would it be possible to have some init process step where you can connect a data-source to determine these kind of things?

hyrahul:
### Summary

- Implemented the `openDataSet` idea from **#1351**.
- This allows `beforeFactory` / `onPrepare` scripts to read the report's own data sets **before the layout is built**.

### What I verified

- Confirmed the report lifecycle:
  - Parameters are resolved before `onPrepare`.
  - `usingParameterValues()` runs before `prepareDesign()`.
  - Data sets are available at this stage.
- Tested with the sample database:
  - Reading data-set rows directly from a script.
  - Using a data-set value to change the report design.
  - Verified the design change in the generated PDF.

### `openDataSet` API

- Added `reportContext.openDataSet(name, params)`.
- `params` contains the input parameter values for the data set, or `null` when no parameters are needed.
- Returns a small **read-only cursor** with:
  - `next()`
  - `getValue()`
  - `getString()`
  - `close()`
- Uses the same query execution approach as `DatasetPreviewTask.extractQuery`.

### Parameter Support

- Added parameter support based on the feedback from @hvbtup.
- Parameters are bound directly to the query using `QueryDefinition.addInputParamBinding`.
- This avoids building query strings manually and does not introduce SQL injection risk.
- Tested with both **string and numeric parameters** and confirmed that filtering works correctly.

### Related Fix

- Moved `needOutputResultSet` from `createContentEmitter` to `setupRenderOption`.
- Previously, it was set **after `startFactory`**, so opening the data engine from a script could miss this setting.

### Tests

Added and verified tests for:

- Reading a data set and using its value to change the report design.
- Opening a non-existent data set.
  - The error is checked through `task.getErrors()` because script exceptions are not propagated from `run()`.
- A script that does not call `close()`.
- Filtering a data set using an input parameter.

- Also verified that each test **fails when the related functionality is broken**, so the tests are checking the expected behaviour rather than simply passing by default.

### DIFF
--- a/engine/org.eclipse.birt.report.engine.emitter.pdf.tests/test/org/eclipse/birt/report/engine/emitter/pdf/AllTests.java
+++ b/engine/org.eclipse.birt.report.engine.emitter.pdf.tests/test/org/eclipse/birt/report/engine/emitter/pdf/AllTests.java
@@ -30,6 +30,7 @@ public class AllTests {
 		/* in package: org.eclipse.birt.report.engine.emitter.pdf */
 		suite.addTestSuite(org.eclipse.birt.report.engine.emitter.pdf.PdfRenderTest.class);
 		suite.addTestSuite(org.eclipse.birt.report.engine.emitter.pdf.PdfConcurrentRenderTest.class);
+		suite.addTestSuite(org.eclipse.birt.report.engine.emitter.pdf.OpenDataSetTest.class);
 
 		// $JUnit-END$
 		return suite;
--- a/engine/org.eclipse.birt.report.engine.emitter.pdf.tests/test/org/eclipse/birt/report/engine/emitter/pdf/OpenDataSetTest.java
+++ b/engine/org.eclipse.birt.report.engine.emitter.pdf.tests/test/org/eclipse/birt/report/engine/emitter/pdf/OpenDataSetTest.java
@@ -1,0 +1,143 @@
+
+/*******************************************************************************
+ * Copyright (c) 2026 Actuate Corporation.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * https://www.eclipse.org/legal/epl-2.0/.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ *
+ * Contributors:
+ *  Actuate Corporation  - initial API and implementation (issue #1351)
+ *******************************************************************************/
+package org.eclipse.birt.report.engine.emitter.pdf;
+
+import java.io.File;
+import java.util.List;
+
+import org.eclipse.birt.report.engine.api.IRunAndRenderTask;
+import org.eclipse.birt.report.engine.api.PDFRenderOption;
+
+/**
+ * Covers {@code reportContext.openDataSet(...)} (issue #1351): reading a data
+ * set from a beforeFactory/onPrepare script, before the report is laid out.
+ * <p>
+ * These tests assert on render completion and on engine errors recorded via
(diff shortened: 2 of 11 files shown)


hvbtup:
I played around with it and it works ad it definitely extends what can be done with BIRT.

However, what if the dataset needs input parameters?

For toy databases, it might work to just filter the data in BIRT.
For production databases, this is not sufficient.
Of course, probably one could modify the SQL query dynamically to add a WHERE clause. But that's error-prone and opens the door for SQL injection attacks unless special care is taken.

Can you extend the feature to supply dataset parameters values? Or is that too complicated?

hvbtup:
Another point worth considering is the following:

In its current form, this is not a solution for what @SteveSchafer-Innovent needs (the current usual work around for this by using hidden tables or lists).

To describe this in more detail, sometimes you have an outer table based on DS "master", and in a table row of this outer table you want to show eg. an inner table, based on an inner DS "detail" and depending on the current outer row data.
(Sometimes this is called a sub-report or master-detail report).
You can easily hide the inner table based on `row.__rownum`, but assume that, for some reason  (e.g. PDF/UA conformance!), you need to hide the whole table detail row of the outer table if the inner table is empty. The difficulty is that you don't have access to the inner data in the visibility expression of the outer table detail row. So the solution is often to perform the same query as for the inner table, but *before/above* this TR is rendered, and save the result (is there inner data or not?) in a report variable which then can be used in the TR visibility expression.

If we could use `openDataSet` - including DS parameter support - in visibility expressions, this could be solved without additional hidden layout objects, by using the function in the visiblity expression for the TR.

hyrahul:
> I played around with it and it works ad it definitely extends what can be done with BIRT.
> 
> However, what if the dataset needs input parameters?
> 
> For toy databases, it might work to just filter the data in BIRT. For production databases, this is not sufficient. Of course, probably one could modify the SQL query dynamically to add a WHERE clause. But that's error-prone and opens the door for SQL injection attacks unless special care is taken. 

-> BIRT's data engine already has mechanism for this **QueryDefination.addInputParamBinding(..)** which binds value to query the same way as ODA drivers own place holders work, so no string concatenation and no injection risk.
 
> Can you extend the feature to supply dataset parameters values? Or is that too complicated? 

-> I have tried adding openDataSet(name, Map<String,Object> params) and it works for single call I have verified with numeric and string parameters both bind and filters correctly. but one bug i got hit hard calling it second time in the same script for any data set throws can not read property undefined.
in summary first call works but second parameterised call doesn't still digging deeper to find the root.

hyrahul:
@hvbtup  
> Can you extend the feature to supply dataset parameters values? -> the feature extended to supply dataset params.

hvbtup:
This looks good to me.
I played around with it a bit. Now you can use `openDataSet` in visibility expressions, and supplying parameters works (I only tested with strings).
The `beforeOpen` events etc. are executed, too, this is fine.
I even checked that it is possible to use `openDataSet` for the same data set while iterating through the results of this dataset (which is nonsense, but should not result in endless loops or garbage output), and it works, too.

So, from my POV: +1 for changing this to a normal PR and merging.

hvbtup (approved):
Looks good to me.

PR #2475, opened 2026-08-26
adding exportDataWithColumnName checkbox added for enabling header export along with data/Value

Ankit3105-dev:
exportDataWithColumnName checkbox absent when exporting the data in CSV. Resulting in missing of header column in exported CSV report.

Environment

OS: [e.g., Windows, macOS, Linux]
Browser: [e.g., Chrome, Firefox]
BIRT Version: [e.g., 4.21.0]
Screenshots

Image - before fix
Screenshot 2026-08-26 at 2 19 20 PM

Additional Context
result before fix
Screenshot 2026-08-26 at 2 20 15 PM

Image - after Fix

Screenshot 2026-08-26 at 2 26 19 PM

result After fix
Screenshot 2026-08-26 at 2 27 24 PM

### DIFF
--- a/viewer/org.eclipse.birt.report.viewer/birt/webcontent/birt/pages/dialog/SimpleExportDataDialogFragment.jsp
+++ b/viewer/org.eclipse.birt.report.viewer/birt/webcontent/birt/pages/dialog/SimpleExportDataDialogFragment.jsp
@@ -275,6 +275,12 @@
 							<label for="exportDataWithCR"><%= BirtResources.getMessage( "birt.viewer.dialog.exportdata.carriage_return" )%></label>
 						</TD>
 					</TR>
+					<TR>
+						<TD><INPUT TYPE="checkbox" ID="exportDataWithColumnName"></TD>
+						<TD nowrap="nowrap">
+							<label for="exportDataWithColumnName"><%= BirtResources.getMessage( "birt.viewer.dialog.exportdata.column.name" )%></label>
+						</TD>
+					</TR>
 				</TABLE>
 			</DIV>
 		</TD>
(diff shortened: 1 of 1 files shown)


viewer/org.eclipse.birt.report.viewer/birt/webcontent/birt/pages/dialog/SimpleExportDataDialogFragment.jsp
hvbtup: Isn't this the same as what you inserted, with the same ID?
If it is not visible and not working, maybe this should be investigated and corrected. 
Adding a new TR seems more like an unclean workaround than a fix.

hvbtup (changes requested):
Besides the solution seeming technically unclean (see my comment),
I think it would make sense to place this new checkbox in the second column instead of adding a new TR.

merks:
@Ankit3105-dev 

FYI, the release cycle ends next week so if you want this completed, the runway is short.

Also, you will need to complete the ECA before we can merge any contribution from you:

https://github.com/eclipse-birt/birt/blob/master/CONTRIBUTING.md#eclipse-contributor-agreement

viewer/org.eclipse.birt.report.viewer/birt/webcontent/birt/pages/dialog/SimpleExportDataDialogFragment.jsp
Ankit3105-dev: changed the fix, to use TD instead of TR.
Screenshot 2026-08-28 at 2 02 39 PM

merks:
I would personally prefer that you rebase and force push so that there is a single commit, but that's a nice-to-have.  The ECA we really needed completed...

Ankit3105-dev:
this issue was appear to be in 4.22 birt, now it is fixed, I can see in 4.24.

merks:
It's always best to test the latest nightly build...

PR #2470, opened 2026-08-23
Replace iText references with OpenPDF in locale message resources

hyrahul:
Fixes #2467.

## What Changed
Replaced "iText" (the PDF  library BIRT no longer uses) with "OpenPDF" in the message resources that still referenced it - 12 files, 2 message keys:
- `designer.preview.previewaction.ErrorMessage` : base English `message.properties` (UI module) + 6 locale `.msg` sources in `org.eclipse.birt.report.designer.core.nl` (de_DE, es_ES, fr_FR, ja_JP, ko_KR, zh_CN). Plain word substitution.
- `PDFCreator`: 5 locale `.msg` sources in `org.eclipse.birt.report.engine.nl` (de_DE, fr_FR, ja_JP, ko_KR, zh_CN). `es_ES` and base English were already clean.

## Why `.msg`, not  `.properties`
- The `_xx_XX.properties` files under each `.nl` module are generated build output, not source.
- `.nl/pom.xml` deletes them on `clean` and regenerates  them from the `.msg` files via Ant's `native2ascii` on `compile`.
- Each module's `.gitignore` confirms they're untracked.
- So this PR edits the `.msg` sources only; `.properties` regenerate on build as usval.

## Why `PDFCreator` isn't a plain word swap in 5 locales

- The `PDFCreator` pattern has a `{1}` placeholder, but `PDFPageDevice.java`
  only supplies one value now (`versionInfo = { BundleVersionUtil
  .getBundleVersion(...) }` — just the engine version, for `{0}`).
- Any locale still referencing `{1}` was already rendering an unresolved
  literal `{1}` in the PDF's Creator metadata — a pre-existing bug,
  independent of the iText wording.
- Base English and `es_ES` had already dropped the "using iText {1}" clause
  entirely, rather than swap it for an OpenPDF equivalent.
- This PR applies that same fix to the remaining 5 locales, instead of a
  literal find-and-replace that would leave the placeholder bug in place.

## Out of scope (left untouched, deliberately)

- Code comments in `PDFPageDevice.java`, `FontMappingManager.java`, and
  `FontMappingManagerFactory.java` also mention "iText" — per the issue's own
  scope ("all non-java resources"), these weren't touched here.
- `designer.preview.previewaction.ErrorMessage` doesn't appear to be
  referenced anywhere in Java source under that key (checked across `UI/`
  and `engine/`) - may be dead code from when iText/OpenPDF was an optional
  detected jar rather than a bundled dependency. 
 
[...]

Issue #2467, opened 2026-08-20
Messages still reference iText, though we're using OpenPDF since a few releases ago.


hvbtup:
### Description

Some messages still contain the text "iText". This should be replace with "OpenPDF".

For example, this also results in setting the PDF "Creator" metadata.

### Environment
Master branch

### Additional Context

All non-java resources of BIRT itself should be scanned.

The text occurs in several messages*.properties and messages*.msg files.

The *.msg files seem to be binary files generated from the *.properties files, I don't know how and when exactly this happens.
So probably only the *.properties files should be modified and the *.msg files hopefully are built automatically.

An example finding:

<img width="587" height="211" alt="Image" src="https://github.com/user-attachments/assets/f63cba4a-6ed6-4fec-9585-5dc457117765" />

merks:
I'm pretty sure it's the other way around because the *.msg has the repository decorator indicating that it's controlled by the repository but the *.properties does not have that decorator and has no history in the History view.

<img width="291" height="330" alt="Image" src="https://github.com/user-attachments/assets/59be2d9f-bf31-4a7f-a8b3-73f8fdf90360" />

merks:
What makes this quite a bit less easy is that I'm not sure the encoding of the *.msg file. Opening in the IDE currently treats it as UTF-8 but it is not UTF-8.

<img width="742" height="910" alt="Image" src="https://github.com/user-attachments/assets/2a9e02d6-2a65-46b3-a491-cd9206f3e7a1" />

The conversion from *.msg  to *.properties encodes the non-ASCII characters use `\u`.  This is all really from the olden days!!

hyrahul:
Hi @merks and @hvbtup ,
I'd like to work on this could some one assign it to me ?

hvbtup:
Google AI says: Edit the *.msg files - the *.properties files are auto-generated during build.
Regarding the encoding: For example, the message_de_DE.msg file is encoded in windows-1252 or iso-8859-1 (non distinguishible for the given text). But the messages_zh_CN.msg is using GB 2312 (I don't know, at least it looks like Chinese to me when I open it in VS Code and reopen it with this encoding).
Definitely not a good idea to edit the files without finding out the actual encoding.
And when editing, one should use a good editor with the correct encoding.
It seems like the encoding is defined in the .xml files of the .nl packages. Search for native2ascii there:

[...]

PR #2476, opened 2026-08-26
Convert *.msg locale resources to UTF-8 encoding (fixes #2470)

hyrahul:
## Summary

- Converted **590 `.msg` locale files** from legacy encodings (`Cp1252`, `GBK`, `SJIS`, `MS949`) to **UTF-8**.
- Updated `nl/pom.xml` so `native2ascii` now uses UTF-8 consistently instead of having four separate encoding rules.
- This is a follow-up to the discussion in **#2471** and **#2467**.

## What Changed

- **590 `.msg` files converted to UTF-8**
  - The actual text was preserved; only the file encoding changed.
  - This covers both locale file formats used in the repository:
    - `<name>_xx_XX.msg`
    - `nl/<locale>/templates/<name>.msg`

- **Updated `nl/pom.xml`**
  - Replaced the four legacy encoding rules with a single UTF-8 rule.
  - The generated `.properties` format remains unchanged, so the runtime loading behavior is not affected.

- **19 files were already UTF-8**
  - These files didn't need any conversion.
  - However, they were previously processed using the wrong legacy encoding during the build.
  - This caused incorrect characters to appear in the generated `.properties` files.
  - Using UTF-8 consistently fixes this issue.

- **8 files were left unchanged**
  - These are `feature_xx_XX.msg` files under `features/*.feature/`.
  - They use a different build mechanism and are outside the scope of `nl/pom.xml`.
  - These are covered separately below.

## Verification

- **Round-trip check**
  - Every converted file was written as UTF-8, read back, and compared with the original content.
  - This confirmed that the actual text was preserved.

- **Full build**
  - Ran the full reactor build before and after the changes:
    - `mvn clean package -DskipTests`
  - Both builds completed successfully.

- **Generated `.properties` comparison**
  - Compared all generated `nl/**/*_xx_XX.properties` files.
  - **649 files** generated from the converted `.msg` files were byte-for-byte identical before and after.
    - This confirms that the encoding conversion itself was lossless.
  - **19 already-UTF-8 files changed in the generated output.**
    - This is an actual bug fix, not just a change to the file encoding.
    - Previously, these files were incorrectly decoded using legacy encodings, resulting in mojibake and unexpected `U+FFFD` characters.
    - For example, `plugin_ja_JP.properties` in `org.eclipse.birt.data.oda.pojo.nl` was affected by this issue.

- **Independent content check**
  - Ran a separate locale-encoding sanity check against the same 649 generated `.properties` files.
  - It checks for unexpected Unicode characters, including `U+FFFD`, and characters outside the expected script for each locale.
  - Results:
    - **Before:** 19 files flagged
    - **After:** 0 files flagged
  - This independently confirms the results from the file-by-file comparison.

## Excluded: 8 `features/*.feature/` Files

- These files are outside the scope of `nl/pom.xml` and use a different build mechanism.
- Git history shows that they contain pre-existing data loss from an older, unrelated commit.
- The original text is recoverable, so I'll handle these in a **separate PR**.

## Also Flagged — Not Part of This PR

- **11 chart-template files** under `org.eclipse.birt.chart.integration.wtp.ui.nl` are not currently processed by `native2ascii`.
- This appears to be a separate build-wiring issue and is unrelated to the encoding changes in this PR.


merks:
This looks super promising and correct. But I only looked on my iPhone so I’ll need to look on a bigger screen.  Thanks so much!

merks:
Sorry, I didn't get to this today.  Today was RC1 day for SimRel.  It's on my TODO list for tomorrow.

hyrahul:
> Sorry, I didn't get to this today. Today was RC1 day for SimRel. It's on my TODO list for tomorrow.

no worries!

merks (approved):
Other than features that were already corrupted and aren't actually used, this looks good.

PR #2479, opened 2026-08-28
Remove feature*.msg

merks:
- These are not converted to *.properties by the build and even if they were, those files are not reference by the build.properties of the features


wimjongman:
Do not forget to update the milestone for every issue and PR.

merks:
Probably I will always forget because I work on too many different projects and if they have different rules I tend to forget them.  But I can try...

PR #2481, opened 2026-09-01
Fix invalid PDF/UA tag structure for Drop: Detail cells

### LINKED ISSUE #2458: PDF/UA Drop: Detail creates invalid tag structure
### Description

Using the Drop: Detail feature (at cell level) in a table group header row creates a corrupt PDF structure.

It might be possible to work around this with the new PDF tag features introduced today.
I don't know and I didn't try it.

If anyone comes up with a fix, that would be nice, but I don't think this will happen, and so the result is "Won't fix" and this issue is just to publish the fact.

Note that the Drop: Detail feature is a dirty hack anyway, and in most of the emitters it doesn't work.

A similar effect (but without automatic reprinting after a page break) can be achieved in two ways:

* For data items in table cells, there is the option to "suppress duplicates".
* Generally, using the SQL LAG analytic function or Javascript, one can compare the current value with the previous value and suppress the output if they are equal.

hyrahul:
## What

Fixes an issue with the PDF/UA tagging for tables that contain a `Drop: Detail` or `Drop: All` cell.

## Root cause

While looking into the issue, I found that `AbstractContentWrapper.getTagType()` and `setTagType()` were not forwarding the tag type to the wrapped content.

This is different from how the other properties in the class work. For tables with a drop cell, the cell content goes through this wrapper, so the tag type was getting lost along the way. As a result, the cells were not getting the expected `TD`/`TH` tags.

I also found a `ClassCastException` in `PDFPageDevice`. The code was casting the cell content directly to `CellContent` instead of using the `ICellContent` interface. Once the tag type was being passed through correctly, this code path could receive a wrapped content instance, which caused the cast to fail.

## Verification

I reproduced the issue locally and rendered the PDF/UA output before and after the fix. I then checked the `/StructTreeRoot` structure.

* **Before:** The rows were ending up as untagged cells, with no `TD`/`TH` tags.
* **After:** The cells are correctly tagged as `TD`/`TH`, and the drop cell's `RowSpan` correctly covers its full extent.
* I also tested two tables without drop cells to make sure the changes didn't introduce any regression.

Fixes #2458

### DIFF
--- a/engine/org.eclipse.birt.report.engine.emitter.pdf.tests/test/org/eclipse/birt/report/engine/emitter/pdf/PdfRenderTest.java
+++ b/engine/org.eclipse.birt.report.engine.emitter.pdf.tests/test/org/eclipse/birt/report/engine/emitter/pdf/PdfRenderTest.java
@@ -23,7 +23,7 @@ import org.eclipse.birt.report.engine.api.PDFRenderOption;
 public class PdfRenderTest extends EngineCase {
 	public void testRenderReport() throws Exception {
 		String thePackage = "test/org/eclipse/birt/report/engine/emitter/pdf/";
-		String[] designs = { "issue-2429", "issue-2445", "issue-2446", "issue-2454", "issue-2466" };
+		String[] designs = { "issue-2429", "issue-2445", "issue-2446", "issue-2454", "issue-2458", "issue-2466" };
 		String suffix = ".rptdesign";
 		PDFRenderOption options = new PDFRenderOption();
 		options.setOutputFormat("pdf");
--- a/engine/org.eclipse.birt.report.engine.emitter.pdf.tests/test/org/eclipse/birt/report/engine/emitter/pdf/issue-2458.rptdesign
+++ b/engine/org.eclipse.birt.report.engine.emitter.pdf.tests/test/org/eclipse/birt/report/engine/emitter/pdf/issue-2458.rptdesign
@@ -1,0 +1,782 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<report xmlns="http://www.eclipse.org/birt/2005/design" version="3.2.27" id="1">
+    <property name="author">Henning von Bargen</property>
+    <property name="createdBy">Eclipse BIRT Designer Version 4.25.0.v202608160751</property>
+    <property name="language">English</property>
+    <text-property name="title">BIRT Issue 2429</text-property>
+    <property name="units">mm</property>
+    <property name="iconFile">/templates/blank_report.gif</property>
+    <property name="bidiLayoutOrientation">ltr</property>
+    <property name="imageDPI">96</property>
+    <property name="locale">en_US</property>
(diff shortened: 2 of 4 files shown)


merks (approved):
It looks logical to me.  But I'm no expert.

hvbtup:
[issue-2458.zip](https://github.com/user-attachments/files/31696347/issue-2458.zip)

The code is OK.

I've added a report for testing.
Please add this to the other PDF/UA test reports and extend the test cases.

It should be noted that it is a bit complicated to achieve the desired result:
The "scope: col" property must be set for the ordinary table header cells **and** for the group header cells with "drop: detail".
But this follows the same strange logic as the "drop: detail" functionality itself, so not too hard to guess.

Can you find a solution for the irregular table rows? - That would be great.
PAC 2026 complains "Irregular Table row" for every "dropped column" instance:
grafik

I am not really an accessibility expert and I don't how severe this warning is for actual accessibility of the generated PDF.
Is the output comprehensible with a screen reader like NVDA?

If you can't find a solution, I'm neverheless fine with merging this. It's much better than an exception.

hvbtup (changes requested):
Please add the test report I provided. 
Look for the directory with the other issue-2xxx.rptdesign reports and the Java class to generate them.
Without the test report as an example on how to use this, the functionality will be too obscure.

Otherwise, great work!

hyrahul:
> Please add the test report I provided. Look for the directory with the other issue-2xxx.rptdesign reports and the Java class to generate them. Without the test report as an example on how to use this, the functionality will be too obscure.
> 
> Otherwise, great work!

Added test along with actual issue-2458.rptdesign report.

hyrahul:
> [issue-2458.zip](https://github.com/user-attachments/files/31696347/issue-2458.zip)
> 
> The code is OK.
> 
> I've added a report for testing. Please add this to the other PDF/UA test reports and extend the test cases.
> 
> It should be noted that it is a bit complicated to achieve the desired result: The "scope: col" property must be set for the ordinary table header cells **and** for the group header cells with "drop: detail". But this follows the same strange logic as the "drop: detail" functionality itself, so not too hard to guess.
> 
> Can you find a solution for the irregular table rows? - That would be great. PAC 2026 complains "Irregular Table row" for every "dropped column" instance: grafik
> 
> I am not really an accessibility expert and I don't how severe this warning is for actual accessibility of the generated PDF. Is the output comprehensible with a screen reader like NVDA?
> 
> If you can't find a solution, I'm neverheless fine with merging this. It's much better than an exception.

While testing, I found in some cases, a page-break check on a shared row-span continuation cell could treat it as a split continuation even though the first part had not been tagged yet. This could later lead to a NullPointerException. The code now safely creates a new structure element in this case instead of crashing.

hvbtup (approved):
Thank you.

merks:
Let's get this in for then coming release.  I will be building a milestone tomorrow after SimRel publishes the final RC2 milestone...

PR #2490, opened 2026-09-09
Repair, reload, and regenerate all models

merks:
- All models should do cleanup, import organizing, and code formatting.
- No model should update the classpath.
- Model projects should disable javadoc comment formatting.
- Move the chart.engine model to a model folder outside of the src folder.
- Disable generating edit, editor, and test projects.
- Move the ChartObject.javajetinc to Class/insert.javajetinc to avoid needing a copying of the out-dated Class.javajet for dynamic templates.
- Repair the insert.javajetinc to remove bogus whitespace, to avoid generating copy operations for java.lang.Object, and to avoid using classes that aren't imported, i.e, java.util.Map.
- Delete duplicate models.
- Fix the oda model.xsd to use the correct namespace  so it can be reconciled by the model.genmodel to produce a correct model.ecore; support keeping a temporary copy of DesignXML.xsd from DTP during reload process.


merks:
With this massive set of changes, there is a non-zero chance that something was overwritten by the generator.  But we're early in the cycle, and this is the only way to migrate our way out of the swamp.

